PR #1, opened 2022-09-09
Add nix flake

goodlyrottenapple:
@tothtamas28 Do we want to set up CI as well?

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -41,8 +41,8 @@ jobs:
         if: always()
         run: |
           docker stop pyk-ci
-  run-tests-nix:
-    name: 'Run Tests Nix'
+  build-on-nix:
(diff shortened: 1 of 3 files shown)


tothtamas28:
> Do we want to set up CI as well?

Yes, the Nix package should be tested in a workflow.

.github/workflows/run-tests.yml
```diff
@@ -41,3 +41,27 @@ jobs:
         if: always()
         run: |
           docker stop pyk-ci
+  run-tests-nix:
+    name: 'Run Tests Nix'
```
tothtamas28, line 45: What do you think, would this be a better name for the job?

```suggestion
  build-on-nix:
    name: 'Build on Nix'
```

tothtamas28 (approved):

PR #16, opened 2022-09-14
Refactor `prelude`

### LINKED ISSUE #10: Reword pyk prelude to be a module
File `pyk/prelude.py` turns into directory `pyk/prelude`, with subfiles for each K module: `pyk/prelude/bool.py`, `pyk/prelude/ml.py`, etc...

See https://github.com/runtimeverification/k/pull/2756#discussion_r933136016

tothtamas28:
Fixes #10


src/pyk/prelude/kbool.py
```diff
@@ -0,0 +1,31 @@
...
+from ..kast import TRUE as KAST_TRUE
+from ..kast import KApply, KInner, KLabel, KToken, build_assoc
+from ..utils import unique
+
+BOOL: Final = KAST_BOOL
+TRUE: Final = KAST_TRUE
+FALSE: Final = KAST_FALSE
+
+
+def boolToken(b: bool) -> KToken:  # noqa: N802
```
tothtamas28, line 14: This can be taken one step further.

```python3
def token(b: bool) -> KToken:
    ...
```

Then on can either

```python3
from pyk.prelude import kbool

kbool.token(b)
```

or

```python3
from pyk.prelude.kbool import token as boolToken

boolToken(b)
```

The two names can also coexist, i.e.

```python3
def token(b: bool) -> KToken:
    ...

boolToken = token
```

then `as` is not needed in the `import` (which ensures consistency).


These same tricks can be done with the basic sorts (i.e. `kbool.SORT` vs `BOOL`).

src/pyk/prelude/k.py
```diff
@@ -0,0 +1,12 @@
...
+from ..kast import KLabel, KSort, KToken
+
+K: Final = KSort('K')
+GENERATED_TOP_CELL: Final = KSort('GeneratedTopCell')
+
+DOTS: Final = KToken('...', K)
+
+# Shouldn't these be KApply instead?
+K_CELLS: Final = KLabel('#KCells')
+EMPTY_K: Final = KLabel('#EmptyK')
```
tothtamas28, line 12: It seems that `EMPTY_K` is always used as `KApply(EMPTY_K)`, so we might as well change the type. (`K_CELLS` does not seem to have uses.)

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "pyk"
-version = "0.1.1"
+version = "0.1.2"
```
radumereuta, line 7: How is this getting updated? Is it manual?

src/pyk/kast.py
```diff
@@ -353,9 +353,9 @@ def match(self, term: KInner) -> Optional[Subst]:
         return None
 
 
-TRUE: Final = KToken('true', 'Bool')
-FALSE: Final = KToken('false', 'Bool')
-DOTS: Final = KToken('...', 'K')
+BOOL: Final = KSort('Bool')
+TRUE: Final = KToken('true', BOOL)
+FALSE: Final = KToken('false', BOOL)
```
radumereuta, line 358: What happened to `DOTS`? Was it not used?

radumereuta (approved):
Looks reasonable to me. Mostly refactoring.
Only two questions^
I'm not that confident with my python level, so if this is not urgent, you could wait for another review. Up to you.

src/pyk/kast.py
tothtamas28: Moved to `pyk.prelude.k`: https://github.com/runtimeverification/pyk/blob/4e717f432bd2a0a0c56efdabcf536ddcc271cba4/src/pyk/prelude/k.py#L8

radumereuta: Then why aren't these tokens defined there as well? Just trying to understand why the split.

tothtamas28: `TRUE` is used as a default value in `pyk.kast`, so it has to be defined there. `FALSE` is there for symmetry. Both tokens are also present in `pyk.prelude.kbool`:

https://github.com/runtimeverification/pyk/blob/4e717f432bd2a0a0c56efdabcf536ddcc271cba4/src/pyk/prelude/kbool.py#L10

src/pyk/prelude/kint.py
```diff
@@ -0,0 +1,17 @@
+from typing import Final
```
ehildenb, line 1: Is it not possible to call it `int.py` instead of `kint.py`? Maybe it conflicts with builtin name `int`.

tothtamas28: Yes, this is because of the naming conflict.

src/pyk/prelude/kbool.py
tothtamas28: @ehildenb, what do you think, should we do this. :point_up:

src/pyk/prelude/k.py
ehildenb: We can remove `#KCells` then, and can make the change to `EMPTY_K`.

src/pyk/prelude/kbool.py
ehildenb: I usually prefer things to be more explicit, because implicit things lead to confusion. But this does seem nice.

ehildenb: I would leave it as is though, and see how people use it. Our changes should be informed by use.

ehildenb (approved):

PR #21, opened 2022-09-19
Add `pipe_stdout` and `pipe_stderr` to `run_process` arguments

### LINKED ISSUE #14: Function `run_process` should be propagating stderr
Currently, when we make subprocess calls, the stderr is silently dropped, which leads to warnings printed by K being ignored.

We should be propagating it instead: https://github.com/runtimeverification/k/pull/2763#discussion_r937034967

tothtamas28:
Fixes #14

(This issue has already been fixed it seems.)

### DIFF
--- a/src/pyk/cli_utils.py
+++ b/src/pyk/cli_utils.py
@@ -41,11 +41,12 @@ def run_process(
     args: Union[str, Iterable[str]],
     *,
     check: bool = True,
-    profile: bool = False,
-    suppress_stderr: bool = False,
     input: Optional[str] = None,
+    pipe_stdout: bool = True,
+    pipe_stderr: bool = False,
     env: Optional[Mapping[str, str]] = None,
     logger: Optional[Logger] = None,
+    profile: bool = False,
 ) -> CompletedProcess:
 
     if type(args) is str:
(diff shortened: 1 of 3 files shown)


src/pyk/cli_utils.py
```diff
@@ -57,12 +58,14 @@ def run_process(
     if not logger:
         logger = _LOGGER
 
+    stdout = subprocess.PIPE if pipe_stdout else None
+    stderr = subprocess.PIPE if pipe_stderr else None
+
```
ehildenb, line 63: By default, will we be passing stderr through to the terminal? That is a must I think, we cannot be ignoring warningsn printed by the underlying tools (K fronend and Haskell backend).

src/pyk/cli_utils.py
```diff
@@ -41,11 +41,12 @@ def run_process(
     args: Union[str, Iterable[str]],
     *,
     check: bool = True,
-    profile: bool = False,
-    suppress_stderr: bool = False,
     input: Optional[str] = None,
+    pipe_stdout: bool = True,
+    pipe_stderr: bool = False,
```
tothtamas28, line 46: This argument has been renamed, but behavior stays the same. Default value is `False`.

src/tests/test_cli_utils.py
```diff
@@ -0,0 +1,49 @@
...
+                else:
+                    self.assertIsNone(res.stdout)
+                    self.assertEqual(os.read(stdout, 1024), b'out\n')
+
+                if pipe_stderr:
+                    self.assertEqual(res.stderr, 'err\n')
+                    self.assertEqual(os.stat(stderr)[ST_SIZE], 0)
+                else:
+                    self.assertIsNone(res.stderr)
+                    self.assertEqual(os.read(stderr, 1024), b'err\n')
```
tothtamas28, line 38: This assertion demonstrates that if `pipe_stderr` is `False`, then the subprocess writes to `stderr`,

src/pyk/cli_utils.py
tothtamas28: Yes, the behavior is unchanged in that regard.

ehildenb (approved):

PR #22, opened 2022-09-19
Eliminate direct call to `run_process` in `KProve`

### LINKED ISSUE #8: Method `KProve.prove` should reuse method `_kprove` instead of calling `run_process` directly
In this PR: https://github.com/runtimeverification/k/pull/2789

I tried to make it so that `_kprove` could be used by `KProve.prove` instead of it calling `run_process` itself directly. Similarly, the `KRun.run` does this. Unfortunately, I was not able to find a working solution.

- Where `_kprove` is defined (called "`kprove`" in this link): https://github.com/runtimeverification/k/blob/master/pyk/src/pyk/ktool/kprove.py#L34
- Where `KProve.prove` calls `run_process` directly instead of using the factored out `_kprove` helper: https://github.com/runtimeverification/k/blob/master/pyk/src/pyk/ktool/kprove.py#L143

The solution here needs to preserve that `stderr` is propagated up.

This isn't blocking anything, but it's annoying.

tothtamas28:
Fixes #8

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "pyk"
-version = "0.1.5"
+version = "0.1.6"
 description = ""
 authors = [
     "Runtime Verification, Inc. <contact@runtimeverification.com>",
--- a/src/pyk/ktool/kprove.py
+++ b/src/pyk/ktool/kprove.py
@@ -170,7 +170,7 @@ class KProve(KPrint):
             raise RuntimeError('kprove failed!')
 
         if dry_run:
-            return mlTop()
+            return mlBottom()
 
         debug_log = _get_rule_log(log_file)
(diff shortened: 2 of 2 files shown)


src/tests/test_cli_utils.py
```diff
@@ -0,0 +1,49 @@
...
+        command = ('sh', '-c', 'echo out ; >&2 echo err')
+        bools = (False, True)
+        stdout, stdout_writer = os.pipe()
+        stderr, stderr_writer = os.pipe()
+
+        for pipe_stdout, pipe_stderr in product(bools, bools):
+            with self.subTest(pipe_stdout=pipe_stdout, pipe_stderr=pipe_stderr):
+                # When
+                with redirect(stdout_writer, 1), redirect(stderr_writer, 2):
+                    res = run_process(command, pipe_stdout=pipe_stdout, pipe_stderr=pipe_stderr)
```
nishantjr, line 23: Neat!

nishantjr (approved):
Looks good!

Issue #31, opened 2022-09-22
KVariables should be aware of their sorts.


nishantjr:
During parsing, the sorts of `KVariable`s are inferred. This is currently stored in a (KAST) attribute. This information should be used when during unparsing. We should also make this class member, or python attribute, and allow passing it in via the contructor. Unparsing, matching, antiunification etc should all take this into account.

PR #33, opened 2022-09-23
Add sort field to KVariable

nishantjr:
- Add sort field to KVariable,
- Make parsing and unparsing aware of sorts.

Beginning of fix for #31 (Note that matching etc also need to be sort-aware.)


src/pyk/kast.py
```diff
@@ -262,10 +262,20 @@ class KVariable(KInner, WithKAtt):
...
+            att = att.update({KVariable.SORT_ATTRIBUTE: sort.to_dict()})
         object.__setattr__(self, 'att', att)
 
+    @property
+    def sort(self) -> Optional['KSort']:
+        if KVariable.SORT_ATTRIBUTE in self.att:
+            return KSort.from_dict(self.att[KVariable.SORT_ATTRIBUTE])
+        return None
+
     @classmethod
```
ehildenb, line 279: Would it be preferable to store the sort directly as an optional field, instead of as an attribute? We could serialize it as an attribute in `from_dict` and `to_dict`.

src/tests/test_kast.py
```diff
@@ -1,15 +1,45 @@
...
+        (
+            'sort-overrides-att',
+            KVariable('Foo', sort=INT, att=KAtt({'org.kframework.kore.Sort': FrozenDict(BOOL.to_dict())})),
+            KVariable('Foo', sort=INT),
+        ),
+    )
+
+    def test_construct(self) -> None:
+        for i, params in enumerate(self.TEST_DATA):
+            name, input, expected = params
```
ehildenb, line 32:
```suggestion
        for i, (name, input, expected) in enumerate(self.TEST_DATA):
```

Does this work instead?

src/pyk/kast.py
nishantjr: The JSON that the current frontend tools produce includes the attribute. This means that we don't lose sort information in the parse -> pyk -> unparse -> kprove cycle.

ehildenb: I don't think it would involve changing the frontend, we just would be using a different representation than the frontend. When we read/write it, we do it as an attribute like the frontend (by adjusting `from_dict` and `to_dict`), but otherwise store it as a field.

The reason would be if accessing a variables sort is going to be a common operation, then we should store it directly instead of having to read it from an attribute every time. But I guess having the same wrapper property `sort` achieves the same interface anyway.

src/tests/test_kast.py
```diff
@@ -1,15 +1,44 @@
...
+
+    def test_construct(self) -> None:
+        for i, (name, input, expected) in enumerate(self.TEST_DATA):
+            with self.subTest(i=i, name=name):
+                self.assertEqual(input, expected)
+
+    def test_to_dict(self) -> None:
+        for i, params in enumerate(self.TEST_DATA):
+            name, input, expected = params
+            with self.subTest(i=i, name=name):
```
ehildenb, line 38:
```suggestion
    def test_to_dict(self) -> None:
        for i, (name, input, expected) in enumerate(self.TEST_DATA):
            with self.subTest(i=i, name=name):
```

Same change here, and anywhere else where this pattern is used in this PR.

src/pyk/kast.py
```diff
@@ -262,10 +262,20 @@ class KVariable(KInner, WithKAtt):
     name: str
     att: KAtt
 
-    def __init__(self, name: str, att: KAtt = EMPTY_ATT):
+    SORT_ATTRIBUTE: Final[str] = 'org.kframework.kore.Sort'
```
ehildenb, line 265: How should we store these special attribute keys @tothtamas28 ? Maybe they belong as static members of the `KAtt` class, so we say `KAtt.SORT`? There are many other special named attributes, and we'd eventually like to have a whitelist, but for now we need to figure out where to store this.

src/pyk/kast.py
nishantjr: That was pretty much my first implementation of this, but there were a lot of corner cases that needed to be handled, for the constructor and the serialization/deserialization function. We could also use `@cachedproperty` if efficiency is an issue.

src/pyk/kast.py
tothtamas28: `KAtt.SORT` sounds good.

src/pyk/kast.py
```diff
@@ -256,16 +256,27 @@ def ml_pred(self) -> KInner:
...
+
+
 @final
 @dataclass(frozen=True)
 class KVariable(KInner, WithKAtt):
     name: str
     att: KAtt
 
-    def __init__(self, name: str, att: KAtt = EMPTY_ATT):
+    def __init__(self, name: str, *, sort: Optional['KSort'] = None, att: KAtt = EMPTY_ATT):
```
tothtamas28, line 268: Consider moving `KVariable` after `KSort`, then

```suggestion
    def __init__(self, name: str, *, sort: Optional[KSort] = None, att: KAtt = EMPTY_ATT):
```

src/pyk/kast.py
```diff
@@ -256,16 +256,27 @@ def ml_pred(self) -> KInner:
...
         object.__setattr__(self, 'name', name)
+        if sort:
+            att = att.update({SORT_ATTRIBUTE: sort.to_dict()})
```
tothtamas28, line 271: Is this the right way to store the attribute? Shouldn't conversion to `dict` occur only when `att.to_dict` is called?

src/tests/test_kast.py
```diff
@@ -1,15 +1,43 @@
...
+        ('no-sort', KVariable('Foo'), KVariable('Foo')),
+        ('sort', KVariable('Foo', sort=INT), KVariable('Foo', sort=INT)),
+        (
+            'sort-via-att',
+            KVariable('Foo', att=KAtt({'org.kframework.kore.Sort': FrozenDict(INT.to_dict())})),
+            KVariable('Foo', sort=INT),
+        ),
+        (
+            'sort-overrides-att',
+            KVariable('Foo', sort=INT, att=KAtt({'org.kframework.kore.Sort': FrozenDict(BOOL.to_dict())})),
```
tothtamas28, line 25: In such a case, consider raising `ValueError` instead.

nishantjr: @tothtamas28 I feel that this works better with the `let` interface? Or do you feel its better to throw anyway, and introduce extra logic in let to deal with this case?

tothtamas28: Yes, let's raise an exception to prevent accidental redefinition of the sort. (Similarly in `let`).

src/pyk/kast.py
tothtamas28: Good catch. It should work properly with `SORT: ClassVar`.

src/tests/test_kast.py
nishantjr: Done.

src/pyk/kast.py
nishantjr: Done.

src/pyk/kast.py
nishantjr: Interesting, looks like you can't have  a `Final[ClassVar[]]`, and that although the type checker infers that final with a default is a ClassVar correctly, the default `str` implementation doesn't.

src/pyk/kast.py
nishantjr: https://github.com/runtimeverification/pyk/issues/35

ehildenb (approved):

PR #34, opened 2022-09-23
Implement kompilation with `kbuild`

tothtamas28:


src/pyk/kbuild/project.py
```diff
@@ -0,0 +1,199 @@
...
+            dependencies=tuple(
+                Dependency(name=name, source=Source.from_dict(project_path, source))
+                for name, source in dct.get('dependencies', {}).items()
+            ),
+            targets=tuple(Target.from_dict(name, target) for name, target in dct.get('targets', {}).items()),
+        )
+
+    @staticmethod
+    def load_from_dir(start_dir: Path) -> 'Project':
+        config_file = find_file_upwards(PROJECT_FILE_NAME, start_dir)
```
ehildenb, line 180: What is this needed for in practice? Seems like we shouldn't be walking upwards all the time like this, might have unexpected behaviors when a user would want it to just error out. If it's not needed for v1, can we remove this feature, and just say to people that `kbuild` must be called in the root of the repo? People will let us know if it's a problem.

src/pyk/kbuild/package.py
```diff
@@ -0,0 +1,155 @@
...
+        return tuple(DepsPackage(dependency) for dependency in self.project.dependencies)
+
+    @cached_property
+    def packages(self) -> Tuple['Package', ...]:
+        res: Tuple[Package, ...] = (self,)
+        for package in self.deps_packages:
+            res += package.packages
+        return res
+
+    def kompile(self, target_name: str) -> Path:
```
ehildenb, line 65: So doing `kbuild kompile` will trigger a `kbuild install`, but not the other way around? It seems counter-intuitive to me.

I would expect `install` to do everything taht `kompile` does, plus to move files into place.

Maybe we need another name for `install`, because it's really more like `sync-sources`.

tothtamas28: Good point. Maybe `sync_include`?

src/pyk/kbuild/project.py
tothtamas28: Good catch, fixed.

(Note that the CLI still works from anywhere within the project, as do `git`, `poetry`, `cargo`, `npm`, etc...)

src/pyk/kbuild/package.py
ehildenb: I guess I would just go for `sync` for now.

src/pyk/kbuild/__main__.py
```diff
@@ -0,0 +1,69 @@
...
+
+    return parser
+
+
+def do_clean(**kwargs: Any) -> None:
+    shutil.rmtree(KBUILD_DIR, ignore_errors=True)
+
+
+def do_install(start_dir: Path, **kwargs: Any) -> None:
+    project_file = find_file_upwards(PROJECT_FILE_NAME, start_dir)
```
ehildenb, line 52: Still doing `find_file_upwards` here.

src/pyk/kbuild/__main__.py
```diff
@@ -0,0 +1,69 @@
...
+    project = Project.load(project_file)
+    package = RootPackage(project)
+    installed_files = package.install()
+    for installed_file in installed_files:
+        print(installed_file.relative_to(KBUILD_DIR))
+
+
+def do_kompile(start_dir: Path, target_name: str, **kwargs: Any) -> None:
+    project_file = find_file_upwards(PROJECT_FILE_NAME, start_dir)
```
ehildenb, line 61: And here. Should we remove these as well?

ehildenb: I would lean towards removing these, people should be working from the root of the repo/kbuild project. Would rather have people tell us they want the feature than assume so and have it cause unexpected behavior.

ehildenb (approved):

src/pyk/kbuild/package.py
tothtamas28: Done.

src/pyk/kbuild/__main__.py
tothtamas28: I don't think this can cause unexpected behavior for CLI use. It is not uncommon for similar tools to work like that, e.g. `cargo`, `npm`, and `poetry` behave this way (also `git`). The advantage is if you're working on some part of a project, you do not need to navigate / open a new terminal to the project root to invoke the command. Besides, since we want to expose a `kbuild` as a Python library as well, we can't make any assumption on the current working directory, so implementation is not any more complex with the feature than without.

tothtamas28:
I tried the tool on a `foundry` target:

```toml
[targets.foundry]
backend = "haskell"
main-file = "foundry.md"
main-module = "FOUNDRY"
syntax-module = "FOUNDRY"
md-selector = "k & ! nobytes & ! node"
hook-namespaces = [ "JSON", "KRYPTO", "BLOCKCHAIN" ]
emit-json = true
```

Kompilation works if  `requires` in `lemmas/lemmas.k` are modified either to an "absolute" form:

```k
requires "evm-semantics/evm.md"
requires "evm-semantics/buf.md"
requires "int-simplification.k"
```

or to a "relative" one:

```k
requires "../evm.md"
requires "../buf.md"
requires "int-simplification.k"
```

I think this is a reasonable restriction.

PR #42, opened 2022-10-05
Add --keep-cells to pyk print

fiedlr:
This PR adds the `--keep-cells` functionality to `pyk print`.

# Meaning
The `--keep-cells` option will make `pyk print` to keep only indicated cells *and "skeletons" of their parents* in the input configuration, the rest is substituted with `...` and minimized accordingly.

# Usage

Usage is the same as `--omit-labels`. Let's say we have the following configuration:

```
<kevm>
<gas> 0 </gas>
<pc> 0 </pc>
<nonce> 0 </nonce>
</kevm>
```

Then running `pyk print {PATH_TO_DEF} {PATH_TO_FILE} --keep-cells "<pc>,<nonce>"` will roughly return 

```
<kevm>
<pc> PC </pc>
<nonce> NONCE </nonce>
...
</kevm>
#And {POSSIBLY_PATH_CONDITION}
```
(with `PC` and `NONCE` depending on the current contents of the cells).

*Note*: `--keep-cells` cannot be used together with `--omit-labels`. Setting both will make the CLI throw an exception.

# Documentation

A description was added to `pyk print -h`.

# Quirks

Currently, only cells with primitive values are supported! Selecting cells containing other cells will return an empty term. Consider that this is actually correct: in this case we want to keep cells, which contain cells that we do not want to keep => the contents of the selected cell get rewritten to `...` => even the selected parent cell is extracted.
*Workaround*: choose the cells with primitive values you actually want to see, then their parent is included as well. There's no option to select multiple cells simply with their parent cell. If there's a large cell with several children cells that you want to track, consider using `--omit-labels` instead.


src/pyk/kastManip.py
```diff
@@ -437,6 +437,32 @@ def _labels_to_dots(k: KInner) -> KInner:
...
 
+def extract_cells(kast: KInner, cells: Collection[str]) -> KInner:
+    """Keep specific cells for printing.
+
+    -   Input: kast term, and list of cells to keep.
+    -   Output: kast term with given cells kept, rest abstracted.
+    """
+
+    contains_selected_cell: Dict[str, bool] = {}
+
```
ehildenb, line 448: Do away with the module comments for now, we aren't using them.

src/pyk/__main__.py
```diff
@@ -47,10 +47,14 @@ def main() -> None:
             _LOGGER.info(f'Wrote file: {args["output_file"].name}')
         else:
             if args['minimize']:
-                abstract_labels = [] if args['omit_labels'] is None else args['omit_labels'].split(',')
+                if args['omit_labels'] != "" and args['keep_cells'] != "":
```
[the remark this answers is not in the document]
ehildenb: There are a few more examples later in the file too, make sure to adjust them all.

fiedlr:
@ehildenb @tothtamas28 Please, make sure you read through *Quirks* in the PR description before review. Thanks!

src/pyk/kastManip.py
```diff
@@ -437,6 +437,23 @@ def _labels_to_dots(k: KInner) -> KInner:
...
+    def _extract_cells(k: KInner) -> KInner:
+        if type(k) is KApply and k.is_cell:
+            contains_selected_cell[k.label.name] = (k.label.name in cells) or any(
+                type(arg) is KApply and arg.is_cell and contains_selected_cell[arg.label.name] for arg in k.args
+            )
+
+            if not contains_selected_cell[k.label.name]:
+                return DOTS
+
+        return k
```
tothtamas28, line 452: Is this a valid simplification?

```suggestion
    def _extract_cells(k: KInner) -> KInner:
        if (
            type(k) is KApply and 
            k.is_cell and
            not k.label.name in cells and
            all(arg == DOTS for arg in k.args)
        ):
            return DOTS
        return k
```

src/pyk/kastManip.py
```diff
@@ -437,6 +437,23 @@ def _labels_to_dots(k: KInner) -> KInner:
     return bottom_up(_labels_to_dots, kast)
 
 
+def extract_cells(kast: KInner, cells: Collection[str]) -> KInner:
```
tothtamas28, line 440:
```suggestion
def extract_cells(kast: KInner, keep_cells: Collection[str]) -> KInner:
```

src/pyk/kastManip.py
```diff
@@ -465,8 +484,14 @@ def minimize_term(term: KInner, keep_vars: Iterable[str] = (), abstract_labels:
...
     term = remove_semantic_casts(term)
     term = useless_vars_to_dots(term, keep_vars=keep_vars)
-    term = labels_to_dots(term, abstract_labels)
+
+    if keep_cells:
+        term = extract_cells(term, keep_cells)
+    else:
+        term = labels_to_dots(term, abstract_labels)
+
     term = collapse_dots(term)
```
tothtamas28, line 493: > If an empty term is returned (for example with --keep-cells "nonexistent_cell", then this is not handled and an exception is thrown in other part of the code. Question: should I catch this or no?

My guess is that this is the part that raises the exception - if so, `collapse_dots` should be fixed so that `collapse_dots(DOTS) == DOTS `. After all, the term `...` is minimal, so it can be returned as value.

fiedlr: Actually, that is not the case. `collapse_dots(DOTS) == DOTS` holds. The thing is that `minimize_term` is used in `__main__.py` where we actually assume that the result contains a config cell because we apply the result to the `split_config_and_constraints` function. This does not have to occur, as ML terms do not care specifically about cells. 

To keep the universality of `minimize_term` for any ML terms, I think the correct solution is to add a `try .. catch` block in this place. If it can be applied, then the printing algorithm works correctly. If it can't, then we can return an exception (that would also handle the case where we are left with something like `... #And PATH_COND) or simply return identity in these cases.

tothtamas28: I'd prefer to handle this case explicitly with `if` rather than exception handling. Can this be easily done?

fiedlr: I understand. However, I can only think of doing the same thing as the check that is done in `split_config_and_constraints`, which would lead to duplicate code.

Or we can edit `split_config_and_constraints` to do something else than throw an exception. For example, return a tuple with `[EMPTY_K, constraints]` in a case like this and check for that.

tothtamas28: Exception handling is fine then.

tothtamas28 (approved):

PR #49, opened 2022-10-07
Implement K-REPL server skeleton

tothtamas28:


dwightguth (approved):
Looks good for an initial skeleton. There's room for improvement with the test harness and of course we still need to add the main rpc request/response loop, but that can be separate prs.

PR #53, opened 2022-10-11
Add module `pyk.krepl.rpc`

tothtamas28:


src/integration_tests/krepl/utils.py
```diff
@@ -25,12 +27,33 @@ def pid(self) -> int:
 
 
 class KReplTest(TestCase):
-    KREPL_ARGS: Iterable[str] = ()
+    KREPL_PORT: int = 8080
```
dwightguth, line 30: we probably should pick a different port than this; this is a pretty commonly used port elsewhere.

src/pyk/krepl/server/__main__.py
```diff
@@ -14,31 +14,26 @@ def main() -> None:
...
+        server.serve_forever()
+    except KeyboardInterrupt:
+        ...
     _LOGGER.info('Server stopped')
 
 
 def argument_parser() -> ArgumentParser:
     parser = ArgumentParser(description='K-REPL Server')
     parser.add_argument('-l', '--loglevel', type=loglevel, default=logging.INFO, metavar='LEVEL', help='log level')
+    parser.add_argument('-p', '--port', type=int, default=8080, help='server port')
```
dwightguth, line 31: same here; pick a different default port.

dwightguth (changes requested):
I have a couple comments, but also it looks like there aren't any commands yet? I would really recommend adding a basic "hello" command to the server so we can test the full process of request/response before merging this. We can delete it later once we have real commands.

src/pyk/krepl/server/__main__.py
```diff
@@ -14,31 +14,26 @@ def main() -> None:
     args = argument_parser().parse_args()
     logging.basicConfig(level=args.loglevel, format=_LOG_FORMAT)
 
-    _LOGGER.info('Server started')
-
-    listener = _Listener()
-    while not listener.terminated:
-        time.sleep(1)
+    server = JsonRpcServer(args.port)
+    server.register(echo)
```
tothtamas28, line 18: A simple endpoint `echo` is registered with the server here.

src/integration_tests/krepl/test_krepl.py
```diff
@@ -14,3 +14,18 @@ def assertPid(self, pid: int) -> None:  # noqa: N802
             os.kill(pid, 0)
         except OSError:
             raise AssertionError(f'Process with PID {pid} does not exist') from None
+
+
+class KReplRpcTest(KReplTest):
+    KREPL_PORT = 8888
+
+    def test_request(self) -> None:
```
tothtamas28, line 22: This test checks whether the server instance handles the `echo` request as expected.

dwightguth (commented):
Sorry, I missed it because it was so concise lol. Go ahead and ignore that comment of mine.

src/integration_tests/krepl/utils.py
tothtamas28: Done.

src/pyk/krepl/server/__main__.py
tothtamas28: Done.

dwightguth (approved):

PR #54, opened 2022-10-11
Quick dockerfile fix 

F-WRunTime:
Updating the default entry user to conatiners built by pyk


tothtamas28:
There's a [build failure](https://github.com/runtimeverification/pyk/actions/runs/3240392152/jobs/5310983706) I can't reproduce locally, or figure out what might cause.

When `make test-pyk` is run in the container as `root`, everything works fine. When however the user is set to `user`, it fails:

```
poetry run make -C pyk-tests
make[1]: Entering directory '/home/user/pyk-tests'
kompile ../k-files/d.k             \
    --output-definition definitions/d \
    --backend haskell                  \
    --emit-json
Error:  Critical: Could not write to ./definitions/d/scanner
(NoSuchFileException: ./definitions/d/scanner)
make[1]: Leaving directory '/home/user/pyk-tests'
make[1]: *** [Makefile:50: d.kompiled] Error 113
make: *** [Makefile:47: test-pyk] Error 2
Error: Process completed with exit code 2.
```

It looks like at that point, `./definitions/d/scanner` should exist for rewriting, but it doesn't. The exception is probably raised [here](https://github.com/runtimeverification/k/blob/f1326214038cac6fc5c955c847938ac445f251e0/kernel/src/main/java/org/kframework/parser/inner/kernel/Scanner.java#L133).

@ehildenb @radumereuta, any ideas what might cause this?

radumereuta:
I haven't seen this before. Did you try running the tests again?

tothtamas28:
> I haven't seen this before. Did you try running the tests again?

Yes, and the failure persists.

radumereuta:
Running this locally I can see that kompile is trying to copy a generated scanner
from: `/tmp/tmp-kompile-18140886876713790802`
to: `./test-kompiled/scanner`
Maybe running this in user mode doesn't allow you to copy from the `/tmp` directory?

I find it surprising that it's using the default `/tmp` dir instead of the `.kompile-***` directory we usually use.
If this isn't something you can solve, we may be able to bypass this issue by changing how the scanners are generated in the front-end.

tothtamas28:
Thanks, @radumereuta, for looking into this, this might well be the root cause. @F-WRunTime, is it possible that the Docker container run on the worker restricts access to `/tmp`?

> changing how the scanners are generated in the front-end

I think doing this might be a good idea in any case, especially if during kompilation other temporary files are created under `.kompile-*`.

radumereuta:
> I think doing this might be a good idea in any case, especially if during kompilation other temporary files are created under .kompile-*.
If you can't find a workaround with the container, please open an issue in the front end and we'll see how we can schedule such a fix.

F-WRunTime:
That's a good question. The only thing I can find on this topic is the base container we use could have set the /tmp folder permissions such that we can't access it. 
I tested this and didn't appear to have any issues with the container build process and creating a file within /tmp 

image

radumereuta:
The problem is not with the generation, but with the copy.
```java
            Files.copy(scanner.toPath(), output.toPath(), StandardCopyOption.REPLACE_EXISTING);
```
If it was created as root, and the copy is done at the user level, then it might be an issue.
Or it's trying to override something already in the output directory and it fails.
But I'm just guessing.

F-WRunTime:
> The problem is not with the generation, but with the copy.
> 
> ```java
>             Files.copy(scanner.toPath(), output.toPath(), StandardCopyOption.REPLACE_EXISTING);
> ```
> 
> If it was created as root, and the copy is done at the user level, then it might be an issue. Or it's trying to override something already in the output directory and it fails. But I'm just guessing.

Ah okay I think I found the root of the issue. We really should have done this where we mount the runner user into the container. And run out of shared folders, I forget why we didn't. But if we want to continue to work it the way we are with this 'user'. I'll post a quick and dirty solution in my next commit but ultimately I would strongly recommend we were running the 'docker run ` command something like this:
```
docker run                             \
          --name ${TAG_NAME}                 \
          --rm -it                           \
          --detach                           \
          --workdir ${{ github.workspace }} \
          --user $(id -u):$(id -g)                  \
          -v ${HOME}:${HOME}                 \
          -v "$(pwd):/opt/workspace"         \
          -v "/etc/passwd:/etc/passwd:ro"    \
          -v "/etc/group:/etc/group:ro"      \
          runtimeverification/${TAG_NAME}
```

F-WRunTime:
This makes sense why none of us could reproduce. Likely case is we are all the FIRST user  on our accounts UID which I think almost always is given the 1001 uid which would match as we're building this such that 'user' (first user in the container) coincidentally has the same uid as we do.

tothtamas28:
> This makes sense why none of us could reproduce. Likely case is we are all the FIRST user on our accounts UID which I think almost always is given the 1001 uid which would match as we're building this such that 'user' (first user in the container) coincidentally has the same uid as we do.

This was exactly the case. I changed the `UID` and `GID` of `user`, and was able to reproduce the issue locally.

The direct cause of the failure was that `kompile` couldn't create folder `/home/user/pyk/pyk-tests/.kompile-*` due to a permission issue. This was because the owner of the copied files were the host user (and this is only a problem with a host user other than `1000`).

This is somewhat surprising, as the `docker cp` [docs](https://docs.docker.com/engine/reference/commandline/cp/) state the following:

> Ownership is set to the user and primary group at the destination.

This seems to be true if the container is run as `root`, but not in general.

Anyways, I created a PR (https://github.com/runtimeverification/pyk/pull/63) that makes tests work with this generic non-root user. I also cherry-picked some commits from this PR.

> We really should have done this where we mount the runner user into the container. And run out of shared folders, I forget why we didn't.

As our runners are not ephemeral, I wanted to copy rather than mount to avoid side-effects on the runner. Also, I find this approach more reproducible as it provides better isolation of guest from host.

F-WRunTime:
> > This makes sense why none of us could reproduce. Likely case is we are all the FIRST user on our accounts UID which I think almost always is given the 1001 uid which would match as we're building this such that 'user' (first user in the container) coincidentally has the same uid as we do.
> 
> This was exactly the case. I changed the `UID` and `GID` of `user`, and was able to reproduce the issue locally.
> 
> The direct cause of the failure was that `kompile` couldn't create folder `/home/user/pyk/pyk-tests/.kompile-*` due to a permission issue. This was because the owner of the copied files were the host user (and this is only a problem with a host user other than `1000`).
> 
> This is somewhat surprising, as the `docker cp` [docs](https://docs.docker.com/engine/reference/commandline/cp/) state the following:
> 
> > Ownership is set to the user and primary group at the destination.
> 
> This seems to be true if the container is run as `root`, but not in general.
> 
> Anyways, I created a PR (#63) that makes tests work with this generic non-root user. I also cherry-picked some commits from this PR.
> 
> > We really should have done this where we mount the runner user into the container. And run out of shared folders, I forget why we didn't.
> 
> As our runners are not ephemeral, I wanted to copy rather than mount to avoid side-effects on the runner. Also, I find this approach more reproducible as it provides better isolation of guest from host.

As long as we are running our jobs in containers they are ephemeral. The piece that won't be is if we try to install something to the host. But this method is fine. the _work folders where jobs run are ephemeral and won't liter or effect other jobs or jobs from the same repository. Regardless this approach is okay to do. 

Should I delete this PR and the branch?

PR #56, opened 2022-10-12
Repl Command reference

dwightguth:
Not yet complete; still need to add case-split and check-implication, as well as the section on authentication. Feel free to review the commands that are here though.


src/pyk/krepl/server/README.md
```diff
@@ -0,0 +1,208 @@
...
+#### Params
+
+* depth: Int. Default 1. The number of rewrite steps to go back.
+
+#### Behavior
+
+Generates a new configuration if it does not yet exist taken by rewinding the
+current configuration `depth` steps.
+
+### step-to-branch
```
virgil-serbanuta, line 75: How about a "step-to-real-branch" command?

The problem is that, in many of my proofs (and I assume that this would be the same for generic symbolic execution), when branching, one of the branches is immediately (without any extra step) marked as "done", and having something like "step-to-branch" stop for those is usually annoying.

virgil-serbanuta (commented):
Something about debugging equation rules would be nice (e.g. listing all equations that were attempted, together with the result, perhaps turning on and off equation debugging).

It's not obvious if this API is also meant for proofs. If that's the case, then something about claim management would be nice (switching between claims, displaying the number of claims and their status)

dwightguth:
We are definitely aware of the need for additional commands beyond what you see here; however, with the exception of the two I mention above, they are not necessarily part of the initial version. We had to put a scope on what we're providing initially, but that doesn't mean additional possibilities can't be added later. For example, one thing we definitely know we want is functionality for breaking, but this is more complex and I'm focusing right now on commands that are fairly primitive.

src/pyk/krepl/server/README.md
ehildenb: I think we can merge that as follow-up, but the execute endpoint of haskell bakcend (which will be used for tihs, instead of the kprove interface), should not have these spurious branches.

src/pyk/krepl/server/README.md
```diff
@@ -0,0 +1,245 @@
...
+name to (program, module, sort) tuples corresponding to the configuration
+parameters of the semantics. Providing the empty string for any module or sort
+name corresponds to the default parser for that parameter if one exists.
+
+#### Behavior
+
+Parses the provided program and initializes the REPL with the resulting initial
+configuration.
+
+### load-raw
```
ehildenb, line 40: I think this can be made as just an option on `load` can't it? No need for a separate command for basically the same thing, just a different format IMO.

tothtamas28 (approved):

PR #57, opened 2022-10-13
Implement K-REPL client skeleton

tothtamas28:


src/pyk/krepl/client/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class KReplClient:
+    _client: JsonRpcClient
+
+    def __init__(self, port: int):
+        self._client = JsonRpcClient(port)
+
+    def load_raw(self, path: Path) -> str:
+        check_file_path(path)
```
dwightguth, line 21: obviously reading the file has to happen on the client side since the server may not have the file in question, but do we want to do the kore parsing on the client or the server? It looks like you chose the client, but I'm curious your reasoning.

src/pyk/krepl/client/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+        return response
+
+    def step_to_branch(self) -> str:
+        response = self._client.step_to_branch()
+        return response
+
+
+class Repl(Cmd):
+    intro = 'K-REPL Shell\nType "help" or "?" for more information.'
+    prompt = '> '
```
dwightguth, line 40: this is definitely something we can improve later, but we may want the name of the semantics in question to appear here.

src/pyk/krepl/client/__main__.py
```diff
@@ -0,0 +1,21 @@
...
+
+from . import DEFAULT_PORT, Repl
+
+
+def main() -> None:
+    args = argument_parser().parse_args()
+    try:
+        Repl(args.port).cmdloop()
+    except KeyboardInterrupt:
+        ...
```
dwightguth, line 11: we should probably do something sensible to close the connection here.

src/pyk/krepl/server/__init__.py
```diff
@@ -1,3 +1,37 @@
...
+from typing import Any, Final, List, Mapping
 
 DEFAULT_PORT: Final = 42412
+
+from pyk.kore.syntax import Pattern
+
+from ..rpc.server import JsonRpcServer
+
+
+class KReplState:
```
dwightguth, line 10: this class needs a docstring explaining its intent. It looks partially implemented currently, but it's not entirely clear what it's supposed to be doing.

src/pyk/krepl/client/__init__.py
tothtamas28: Good point, it's better to perform parsing on the server side. This will simplify client implementation.

src/pyk/krepl/client/__init__.py
tothtamas28: Good idea.

src/pyk/krepl/client/__main__.py
tothtamas28: There's no need, the underlying `requests.request` function which is called under the hood takes care of that.

src/pyk/krepl/client/__init__.py
tothtamas28: Done.

src/pyk/krepl/server/__init__.py
tothtamas28: Done.

src/pyk/krepl/server/__init__.py
```diff
@@ -1,3 +1,40 @@
...
+from pyk.kore.parser import KoreParser
+from pyk.kore.syntax import Pattern
+
+from ..rpc.server import JsonRpcServer
+
+
+class KReplState:
+    """Store proof tree and expose domain-specific methods over it. Will eventually be encapsualted in a user session."""
+
+    def load_raw(self, pattern: Pattern) -> None:
```
dwightguth, line 14: Do we want to have load functions here or do we want to put this as part of the constructor of the state class? It seems to me that that depends on whether we might want to maintain multiple program executions in a single graph or not. If we do, we would need it like this, but if we don't, and we want this class to represent a single initial state only, then it makes more sense from an encapsulation perspective to make this be a constructor. What do you think?

tothtamas28: I like the idea. But this class will probably evolve a lot, this is just a placeholder for server-side session state.

dwightguth (approved):
This is okay for an initial version. There's going to have to be some refactoring done as we actually implement the commands, but I'm approving because I don't want to get too stuck on details when things are still essentially placeholders.

PR #58, opened 2022-10-13
Implement K-REPL server skeleton with Flask

tothtamas28:


dwightguth:
Fix the conflicts and I'll review this.

src/pyk/krepl/client/__init__.py
```diff
@@ -1,31 +1,39 @@
...
 
-    def load_raw(self, path: Path) -> str:
+    def load_raw(self, path: Path) -> Dict[str, Mapping]:
         check_file_path(path)
         with open(path, 'r') as f:
             text = f.read()
-
-        response = self._client.load_raw(text)
-        return response
+        response = self._request('POST', 'load-raw', json={'term': text})
```
dwightguth, line 31: it seems like we ought to be able to implement these changes in a better way somehow, so that it's more transparent, and we don't have to put a bunch of details about how the requests are sent into the methods here.

dwightguth (changes requested):
it doesn't look like this approach has any built in support for the JSON RPC protocol, which means if we're going to use it, we need to explicitly add that layer somewhere. It's not as simple as just receiving json requests and sending a json response; there's a whole protocol with certain structure and certain error reporting that we have to make sure to obey.

src/pyk/krepl/client/__init__.py
tothtamas28: > it doesn't look like this approach has any built in support for the JSON RPC protocol, which means if we're going to use it, we need to explicitly add that layer somewhere

Implementing or pulling in an existing JSON-RPC layer would be relatively straightforward. But I don't think we should use JSON-RPC for the API, as the scope has shifted from RPC towards a web application. In this setting, an [OpenAPI](https://www.openapis.org/) or [GraphQL](https://graphql.org/) interface would make more sense I think.

tothtamas28:
> it doesn't look like this approach has any built in support for the JSON RPC protocol, which means if we're going to use it, we need to explicitly add that layer somewhere

Implementing or pulling in an existing JSON-RPC layer would be relatively straightforward. But I don't think we should use JSON-RPC for the API, as the scope has shifted from RPC towards a web application. In this setting, an [OpenAPI](https://www.openapis.org/) or [GraphQL](https://graphql.org/) interface would make more sense I think.

src/pyk/krepl/client/__init__.py
tothtamas28: Here, only method, the path section of the URL and JSON payload is provided. But we can come up with conventions to simplify this further, or use some library to abstract over this layer based on a schema description.

dwightguth: I think we ought to be able to make this considerably simpler with a judicious getattr function

tothtamas28: I do not see that solution yet, can you elaborate?

Also, note that this layer will be [replaced](https://github.com/runtimeverification/pyk/blob/930c39d2d669a59e319cf6e341686992853649d8/src/pyk/krepl/client/client.py) by a GraphQL client in the upcoming PR.

dwightguth: If this code is getting replaced anyway, it's fine.

dwightguth (approved):

PR #62, opened 2022-10-18
Add schema for proposed GraphQL API

tothtamas28:


src/pyk/krepl/server/docs/schema.graphql
```diff
@@ -0,0 +1,128 @@
...
+  tags: [String!]!
+  dot: String!
+  tree: String!
+}
+
+type Node {
+  id: ID!
+  aliases: [String!]!
+  term: Term!
+  inEdge: Node
```
tothtamas28, line 51: I assumed at most one incoming edge (exactly one for all except the root), as implied by REPL command `predecessor`. This suggests a tree data structure, which I think is the right representation for symbolic execution, as it is
* easier to handle algorithmically than a generic graph;
* is explicit about the order of exploration on a given path.

Circularity and other forms of subsumption between configurations can then be represented by `cover` edges, so clients can nevertheless display results as a graph.

@dwightguth, what do you think?

src/pyk/krepl/server/docs/schema.graphql
```diff
@@ -0,0 +1,128 @@
...
+}
+
+type Mutation {
+  # Initialization
+  load(input: LoadInput!): Graph!
+  loadRaw(kore: String!): Graph!
+
+  # Behavior
+  step(depth: Int, branch: Bool = True, wait: Bool = False): StepProcess!
+  prune(id: ID!): PruneResult!
```
tothtamas28, line 26: This is for REPL command `rewind`. I made it synchronous, but similarly to `step`, the mutation could return a `PruneProcess`.

src/pyk/krepl/server/docs/schema.graphql
dwightguth: The problem here is that if you have a nonterminating execution that reaches the exact same state it was at earlier in the execution, and try to represent this as a tree, not only will you keep executing forever without realizing there is an infinite loop, but you also won't be able to implement state ids as cryptographic hashes of the configuration because you will have two identical states (and thus the same id) that you will have to somehow track distinctly.

src/pyk/krepl/server/docs/schema.graphql
dwightguth: Instead of having an actual command for this, we should just keep track of how many steps have been taken at each state, and then step forward the appropriate number of steps from the state's predecessor. This works for all states that have exactly one predecessor, whereas if there is more than one, we probably just report an error and tell the user to do the same thing by hand.

src/pyk/krepl/server/docs/schema.graphql
tothtamas28: These issues can be addressed even if the backing data structure is a tree:
* Visited the same state -> mark the relationship with a cover edge and stop exploration on the branch.
* State ids as hashes -> have ids of the form `hash(parent):hash(current)` or `hash(hash(parent), current)`

src/pyk/krepl/server/docs/schema.graphql
tothtamas28: Oh, I think I misunderstood the command then. Will add `steps: Int` to `Edge` and replace `prune` by `rewind: RewindProcess`.

src/pyk/krepl/server/docs/schema.graphql
dwightguth: I'm not entirely sure I understand the concept of cover edges fully so I'd kinda like to get Everett's opinion heru, but this does sound reasonable to me. Since we would stop execution after reaching a state we've seen before, it ought to always be the case that the conjunction of the state and its predecessor is unique.

src/pyk/krepl/server/docs/graphql-api.md
```diff
@@ -0,0 +1,23 @@
...
+The following table lists how each REPL command, as described in the [README](../README.md), is covered by a query or mutation of the GraphQL API.
+
+|    REPL command    |      GraphQL API       |
+|--------------------|------------------------|
+| `load`             | `mutation load`        |
+| `load-raw`         | `mutation load-raw`    |
+| `step`             | `mutation step`        |  
+| `step-to-branch`   | `mutation step`        |
+| `rewind`           | `mutation prune`       |
+| `alias`            | `mutation addAlias`    |
```
dwightguth, line 12: Do we want aliases to live on the server or should that just be something that the client keeps track of, like `select`?

src/pyk/krepl/server/docs/schema.graphql
```diff
@@ -0,0 +1,128 @@
...
+type Query {
+  # Model
+  graph: Graph!
+
+  # Process management
+  processes(id: ID, status: ProcessStatus): [Process!]!
+  wait(id: ID!): Process!
+
+  # Presentation
+  diff(from: ID, to: ID): String!
```
dwightguth, line 16: I'm not convinced using the same id type for both process ids and config ids makes sense here.

src/pyk/krepl/server/docs/schema.graphql
```diff
@@ -0,0 +1,128 @@
...
+}
+
+type Mutation {
+  # Initialization
+  load(input: LoadInput!): Graph!
+  loadRaw(kore: String!): Graph!
+
+  # Behavior
+  step(depth: Int, branch: Bool = True, wait: Bool = False): StepProcess!
```
dwightguth, line 25: If we're doing defaults this should be a default of depth 1.

src/pyk/krepl/server/docs/graphql-api.md
tothtamas28: Good point, I think we can safely move this to the client.

src/pyk/krepl/server/docs/schema.graphql
tothtamas28: Added custom scalar types `NodeID`, `EdgeID`, `ProcessID`.

src/pyk/krepl/server/docs/schema.graphql
tothtamas28: Done.

src/pyk/krepl/server/docs/schema.graphql
dwightguth: Ah, that makes sense.

tothtamas28:
Superseded by #72

tothtamas28:
Superseded by https://github.com/runtimeverification/pyk/pull/72

Issue #27, opened 2022-09-19
Possible failure on CI going undetected


ehildenb:
On a successful CI run, I see this output:

![image](https://user-images.githubusercontent.com/12721972/191092218-dcee6e05-e30a-45d8-a3fc-90928efe8097.png)

This looks suspicious to me, because it looks like a failure of the tool, but is not failing the overall CI job. Is there some way this is marked? Or for jobs like this where we _expect_ failure, should we be silencing the output or capturing it somehow else?

PR #65, opened 2022-10-20
Run integration tests in parallel

### LINKED ISSUE #26: Run integration tests in parallel
The integration tests are doing a bunch of `kompile` and `kprove` calls, which currently are done fully serial.

All the `kompile` calls can go in parallel, and then all the `kprove` calls can go in parallel too. This should greatly reduce CI time.

tothtamas28:
Fixes #26


ehildenb:
This is a known failure? 
image

ehildenb:
Did you test that this is correctly propogating failures up through the Makefile? Honestly, I would prefer if the parallelism happened in the Python, not in the Makefile. This is somewhat convoluted, and I'd like a parallel way to run Python tests set up as a default anyway.

tothtamas28:
> This is a known failure?

Yes: https://github.com/runtimeverification/pyk/issues/27

tothtamas28:
> I would prefer if the parallelism happened in the Python, not in the Makefile. This is somewhat convoluted, and I'd like a parallel way to run Python tests set up as a default anyway.

I agree, will look into it. I'll cherry-pick the refactoring commits into another PR so that they ca be reviewed and merged until I figure out the rest. This moves back to draft.

PR #73, opened 2022-10-25
Initial implementation of kast_to_kore and kore_to_kast

ehildenb:
Part of: https://github.com/runtimeverification/pyk/issues/76

This is focused on implementing enough of `kast_to_kore` and `kore_to_kast` so that the new rewriter experiment with KEVM can make efficient queries into the LLVM backend without ever having to invoke `kast` binary.

- Adds some tests to the `test_kore_to_kast.py` file, and updates some of them.
- Add some helper facilities to `KDefinition`, including:
  - Memoize the computation of `production_for_klabel` (expensive).
  - Add a `subsorts` computation, and memoize it.
  - Add a `return_sort: KLabel -> KSort` and `argument_sorts: KLabel -> List[KSort]` methods, which make it easy to know the return and argument types of a given `KLabel`.
- Adds initial simple conversions between KAst and Kore in `KPrint`, enough to handle the test inputs. Sets it up so that if conversion will not work with what's implemented so far, it falls back to `kast` binary.
- Passes the `--profile` option through on `KRun.run`.
- Adds a new `KRun.run_kore` which uses `kast_to_kore` and `kore_to_kast` to convert the input program rather than leaving it to `krun` script.


src/pyk/ktool/kprint.py
```diff
@@ -91,6 +95,41 @@ def __init__(self, definition_dir: Path, use_directory: Optional[Path] = None, p
         self._definition = None
         self._symbol_table = None
         self._profile = profile
+        self._unmunge_codes = {
```
tothtamas28, line 98: `munge` and `unmunge` do not depend on the `self`, consider making them top-level functions.

tothtamas28 (approved):

dwightguth (approved):
What you have seems okay, although of course it won't be able to handle everything yet. But what it can handle it seems like it's handling correctly to me.

src/pyk/ktool/kprint.py

PR #80, opened 2022-10-31
Fix version bumper

ehildenb:
Currently, if you open a PR and another PR gets opened and merged in the meantime, it will not correctly detect that it needs to bump the version again.

This change makes it so that it always tries to bump the version number directly from current master, and then tries to commit and push the results, instead of trying to detect if the version needs to be bumped by comparing against current master.


ehildenb:
It worked: 
image

tothtamas28 (approved):

PR #81, opened 2022-11-03
Implement `execute` method via RPC server in `KProve` class

ehildenb:
This PR takes some steps towards: https://github.com/runtimeverification/hs-backend-booster/issues/3

I made an initial attempt to have the new KEVM rewriter use kore-rpc to query the old backend (as a replacement for textual `kprove` interface), and ran into several limitations in our Kast <-> Kore conversions and had to figure out the logic for wrapping the RPC server to make the calls.

- Adds the `depth` option directly to the various methods in `KProve` class.
- Handles Kast <-> Kore conversion for variables, with several tests added of examples that were run into while trying to handle KEVM.
  - The munger/unmunger had to be modified to deal with KVariable as well as KLabel.
  - The `_add_sort_injections` pass is made to handle the case where the input/output sorts are the same, so that we don't need to check that they are different at the call sites.
  - The built-in data-structures conversion routine will print a better message about which specific subterm can't be converted between the two.
- Adds a method `KProve.execute`, which wraps the RPC server in the following way (tests are added):
  - The first time you use `KProve.execute`, a server and client are started up, and the `execute` endpoint of the server is called.
  - The server is kept around for subsequent calls.
  - It has roughly the same interface as `get_claim_basic_block` so it can be plugged into downstream libraries. We may need to change this, but if not then we may be able to simplify the `execute` endpoint.

So the server will only be started if it's needed, but won't need to be restarted for subsequent queries. I'm hoping this gives us better overhead with the backend communication.


src/pyk/ktool/kprove.py
```diff
@@ -96,6 +102,9 @@ def _build_arg_list(
     if dry_run:
         args.append('--dry-run')
 
+    if depth is not None and depth >= 0:
```
tothtamas28, line 105: `_kprove` should raise `ValueError` on `depth < 0`. This can then be simplified.

src/pyk/ktool/kprove.py
```diff
@@ -113,17 +124,28 @@ def __init__(
         use_directory: Optional[Path] = None,
         profile: bool = False,
         command: str = 'kprove',
+        port: int = 3000,
```
tothtamas28, line 127: Maybe `port` could be `Optional` as well, what do you think?

src/pyk/ktool/kprove.py
```diff
@@ -284,6 +313,25 @@ def _is_fatal_error_log_entry(line: str) -> bool:
...
+        depth: Optional[int] = None,
+        cut_point_rules: Optional[Iterable[str]] = None,
+        terminal_rules: Optional[Iterable[str]] = None,
+    ) -> Tuple[int, bool, KInner]:
+        kore = self.kast_to_kore(cterm.kast, GENERATED_TOP_CELL)
+        assert isinstance(kore, Pattern)
+        _, kore_client = self.kore_rpc
+        er = kore_client.execute(kore, max_depth=depth, cut_point_rules=cut_point_rules, terminal_rules=terminal_rules)
+        depth = er.depth
+        branching = er.next_states is not None and len(er.next_states) > 0
```
tothtamas28, line 328:
```suggestion
        branching = bool(er.next_states)
```

tothtamas28: Is this correct though? Can it not be the case that `len(er.next_states) == 1`, in particular for a `CutPointResult`?

ehildenb: Hmmmm, yes, I guess it should be `> 1`, not `> 0` then. But I don't think we should do `bool(...)` instead?

tothtamas28: Yes, in that case, `bool(...)` won't work, so

```python3
branching = er.next_states is not None and len(er.next_states) > 1
```
is fine. Alternatively you can do

```python3
branching = len(er.next_states or []) > 1
```

src/pyk/ktool/kprove.py
ehildenb: I've made this code raise `ValueError` for `depth < 0`. This code is invoked by `_kprove` to build the argument list.

src/pyk/ktool/kprove.py
ehildenb: I'm not sure that actually gets us anything, because we need to have a port set if the RPC server is called, but I set it as an optional argument to the constructor.

src/pyk/ktool/kprove.py
ehildenb: I've done the former.

ehildenb:
Ready for re-review. I've also:

- Added another testing harness for execute endpoint but against `SIMPLE-PROOFS` module, with a simple transition system defined there.
- Added better cleanup of the RPC server when testing, so that it's possible to have multple tests of the execute endpoint run in close time to each other.

tothtamas28 (approved):

src/pyk/ktool/kprove.py
```diff
@@ -113,17 +126,33 @@ def __init__(
...
+
+    def kore_rpc(self) -> Tuple[KoreServer, KoreClient]:
+        if not self._kore_rpc:
+            _kore_server = KoreServer(self.definition_dir, self.main_module, self.port)
+            _kore_client = KoreClient('localhost', self.port)
+            self._kore_rpc = (_kore_server, _kore_client)
+        return self._kore_rpc
+
+    def close_kore_rpc(self) -> None:
+        if self._kore_rpc is not None:
```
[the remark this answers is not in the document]
ehildenb: I've gotten into the habit of being more explicit with these checks, and think I will keep it that way. There were a few times where I relied on truthiness and the results surprised me, so I'd rather just play it safe with the code.

src/pyk/ktool/kprove.py
```diff
@@ -284,6 +320,25 @@ def _is_fatal_error_log_entry(line: str) -> bool:
...
+        depth: Optional[int] = None,
+        cut_point_rules: Optional[Iterable[str]] = None,
+        terminal_rules: Optional[Iterable[str]] = None,
+    ) -> Tuple[int, bool, KInner]:
+        kore = self.kast_to_kore(cterm.kast, GENERATED_TOP_CELL)
+        assert isinstance(kore, Pattern)
+        _, kore_client = self.kore_rpc()
+        er = kore_client.execute(kore, max_depth=depth, cut_point_rules=cut_point_rules, terminal_rules=terminal_rules)
+        depth = er.depth
+        branching = er.next_states is not None and len(er.next_states) > 1
```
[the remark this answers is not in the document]
ehildenb: Once again, I really prefer to be explicit here about the fact that I'm checking for `None` specifically. I think I will continue to do this in my code, because I've had unexpected behavior which slowed me down.

src/pyk/ktool/kprove.py
```diff
@@ -96,6 +102,11 @@ def _build_arg_list(
     if dry_run:
         args.append('--dry-run')
 
+    if depth is not None:
+        if depth < 0:
+            raise ValueError(f'Option --depth must be non-negative, got: {depth}')
```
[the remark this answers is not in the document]
ehildenb: My thinking was that someone may call `_build_arg_list` without going through `_kprove`, so we want to handle those cases too.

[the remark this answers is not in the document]
ehildenb: I've adjusted the message, and moved the check.

PR #83, opened 2022-11-04
More kast <-> kore conversions

ehildenb:
This PR makes a few changes:

- The default sort for symbols in `prelude.ml` is changed from `K` to `GENERATED_TOP_CELL`, which is more frequently what we want.
- Conversions for ML symbols `#And` and `#Equals` are added, which enables us to begin dealing with constrained terms passed from `CTerm` class.
- When retrieving productions from the definition by `KLabel`, we only compare the klabel name and not the sorts. This should still be safe, because we're still asserting that only one production is returned. So if the same symbol has different signatures, this method will die horribly and the user can make it smarter or submit a bug report. So far it hasn't caused me problems.


radumereuta (approved):
You can have overloaded productions with the same KLabel but with different signatures.
First example that comes to mind is lists in simple-untyped.
Not the scope of these changes but something to keep in mind.
In the [java](https://github.com/runtimeverification/k/blob/master/kore/src/main/scala/org/kframework/definition/outer.scala#L145) implementation, we return a set of productions and we have to continuously check if we get zero/one/many results.

The rest of the code lgtm, but I'm not that familiar with pyk, so maybe you should get another approval.

tothtamas28 (approved):

dwightguth:
What Radu says is actually incorrect. Overloaded productions like he describes will have different symbols in both kast and kore despite having the same klabel attribute.

PR #85, opened 2022-11-07
Add method `run_kore_term` to `KRun`

tothtamas28:


src/pyk/ktool/krun.py
```diff
@@ -59,48 +28,181 @@ def __init__(self, definition_dir: Path, use_directory: Optional[Path] = None, p
             self.main_module = mm.read()
 
     def run(
-        self, pgm: KInner, depth: Optional[int] = None, expand_macros: bool = False, args: Iterable[str] = ()
```
tothtamas28, line 62: Notice that I removed `args` from the parameter list: if a command line option is missing, a PR should be opened to add  a corresponding parameter rather than worked around downstreams by passing `args`.

ehildenb: This adds a bit of friction downstream honestly, to not have the `args` escape hatch, but generally I agree.

Also, it's been a while, so downstream projects should know the CLI arguments they need.

ehildenb: Actually, we definitily still need `args` for several examples. They use `-cCONFIG_VAR` or `-pCONFIG_VAR`, which is hard to make a direct argumnet for here (I guess you could have a separate `config_vars: Dict[str, KInner]` or `config_vars: Dict[str, Pattern]` argument).

ehildenb: I can make that change in a follow-up PR.

ehildenb (approved):

src/integration_tests/test_krun.py
```diff
@@ -38,6 +40,39 @@ def test_run(self) -> None:
...
+        expected = self._state(k='.', state=f'{x} |-> 1')
+
+        # When
+        actual = self.krun.run_kore_term(pattern)
+
+        # Then
+        self.assertEqual(actual, expected)
+
+    def _state(self, k: str, state: str) -> Pattern:
+        pretty_text = f"""
```
ehildenb, line 56: I guess this sort of thing can be done with `KDefinition.init_config` followed by setting some cells.

src/pyk/ktool/krun.py
ehildenb: Ah wait, I see now that `_krun` handles `pmap` and `cmap`. I guess I'd prefer if thees were passed as just a single map, `config_vars: Mapping[str, KInner]` or `Mapping[str, Pattern]` as appropriate. Then it's up to the user externally to do parsing themselves, which is usually more appropriate because most uses of config vars for higher-performance applications don't want to leave handling them up to `krun` script or kast.

PR #92, opened 2022-11-10
Add method `run_kore_config` to `KRun`

tothtamas28:


src/pyk/ktool/krun.py
```diff
@@ -115,6 +115,31 @@ def run_kore_term(
...
+        config: Mapping[str, Pattern],
+        *,
+        depth: Optional[int] = None,
+        expand_macros: bool = False,
+    ) -> Pattern:
+        proc_res = _krun(
+            definition_dir=self.definition_dir,
+            output=KRunOutput.KORE,
+            pmap={var: 'cat' for var in config},
+            cmap={var: pattern.text for var, pattern in config.items()},
```
tothtamas28, line 129: As pointed out by @geo2a on [Slack](https://runtimeverification.slack.com/archives/C7E701MFG/p1668070363547619), this is not the ideal solution.

Currently, `krun` breaks if the `-pVAR` value is not the name of a command, as the whole string is passed as the first argument to `execute` (with the second argument being the temporary file storing the `-cVAR` value), which then just executes `"$@"` (so for example, `python` works, but `python -m script.py` doesn't):

https://github.com/runtimeverification/k/blob/5bee2516a26cfca51df25a7ce8c27caeec2ae725/k-distribution/src/main/scripts/bin/krun#L494

It would be nice though if `krun` also worked with more complex parser commands (similarly to how `kompile --post-process` works), that way it wouldn't be necessary to set up a command with permissions just so that we can pass a file as value for `-cVAR`.

ehildenb: Hmmm, seems like something that we can improve on later, but is also just a limitation of the current `krun` script.

ehildenb (approved):

PR #93, opened 2022-11-11
Improvements in Kast -> Kore conversions

ehildenb:
This PR builds up the functionality needed for KEVM to do KCFG exploration using RPC endpoints, instead of textual interface.

- Add Kast <-> Kore updates:
  - Conversion for `#Ceil` is added (and test).
  - Conversion `kast_to_kore` is modified to regularize sorts annotating variables before doing the conversion (and test).
- Keeps `build_rule` from renaming variables to themselves, so that it does not throw away any sort information that some of the variables may have (and test).
- Adds `KDefinition.sort_vars : KInner -> KInner` and `KDefinition.sort_vars_subst : KInner -> Subst`. These methods are used to normalize the types of variables showing up in terms to all be the same sort (needed for converting Kast -> Kore, so that all converted variables have the same sort).
- This uses new helper `var_occurances : KInner -> Dict[str, List[KInner]]`, which computes occurances of variables showing up in terms, but keeps the lists of occurances. This method is also used in `free_vars` and `count_vars`.
- `KCFG.create_node` is modified to not strip generated cells upon inserting new nodes. This is needed for being able to take the output and put it back into the backend. Now `build_rule` removes the cells directly before constructing the rule for the frontend (hold onto the cells as long as possible).
- A now unused function (checked here, KEVM, and KSummarize) `kastManip.mark_useless_variables` is removed. It's logic was inlined into `build_rule` a while ago.


src/pyk/kast.py
```diff
@@ -1780,6 +1801,21 @@ def top_down(f: Callable[[KInner], KInner], kinner: KInner) -> KInner:
     return f(kinner).map_inner(lambda _kinner: top_down(f, _kinner))
 
 
+# TODO: make method of KInner
+def var_occurances(term: KInner) -> Dict[str, List[KVariable]]:
```
tothtamas28, line 1805: Typo: `var_occurences`. Could also be called `var_occs` for conciseness.

tothtamas28, line 1805: I think `Set` would be a better fit here.

```suggestion
def var_occurances(term: KInner) -> Dict[str, Set[KVariable]]:
```

src/pyk/kast.py
```diff
@@ -1704,6 +1704,27 @@ def _compute_subsorts(self, sort: KSort) -> List[KSort]:
                 _subsorts.extend([_subsort] + self.subsorts(prod.items[0].sort))
         return _subsorts
 
+    def sort_vars_subst(self, kast: KInner) -> Subst:
```
tothtamas28, line 1707: What do you think about this approach? (Did not test it.)

```python3
name_to_occs = var_occurences(kast)
name_to_sorts = {name: {occ.sort for occ in occs} for name, occ in name_to_occs.items()}
conflicting_name = some(name for name, sorts in name_to_sorts.items() if len(sorts) > 1)
if conflicting_name:
    raise ValueError(f'Could not find common subsort among variable occurances: {name_to_occs[conflicting_name]}')
subst = {name: single(sorts) for name, sorts in name_to_sorts.items()}  # each sorts set has exactly one element
return Subst(subst)
```

where `some` is a helper that can be added to `utils`:

```python3
def some(iterable: Iterable[T]) -> Optional[T]:
    return next(iter(iterable), None)
```

src/pyk/kastManip.py
```diff
@@ -205,12 +205,9 @@ def _extract_subst(conjunct: KInner) -> Optional[Subst]:
...
     counter: typing.Counter[str] = Counter()
-
-    def count(term: KInner) -> None:
-        if type(term) is KVariable:
-            counter[term.name] += 1
-
-    collect(count, term)
+    occurances = var_occurances(term)
+    for vname in occurances:
+        counter[vname] = len(occurances[vname])
```
tothtamas28, line 210: I prefer the old approach, it is probably a bit more efficient. This however induces a good test for `var_occurences`.

tothtamas28 (approved):

src/pyk/kast.py
ehildenb: No, I want it to retain duplicates so that `count_vars` works properly.

src/pyk/kast.py
ehildenb: Well, I'm not sure. This method eventually needs to get a lot more complicated because it will have to do sort inference eventually (for variables with differing sort labels in the same connected component of sorts, it will have to find the _least_ sort for example, or for variables which show up in different contexts, it will have to find the least sort of all the contexts the variables show up in). So I'd rather not try to simplify it, because later it's going to have to turn into a normal term-traversal with `bottom_up` or `top_down` anyway.

src/pyk/kastManip.py
ehildenb: You think it's more effecient? I don't think it should be algorithmically any more efficient, this does one term traversal (for `var_occurances`), then does a bunch of list lengths (which I guess should be O(1) for each list length?)

tothtamas28: I mean more efficient in terms of memory: for each variable, the old implementation stores the count, whereas the new one builds a list (which then gets garbage collected).

src/pyk/kast.py
ehildenb: I don't think the above will work, most occurances of variables have no sort attached to them.

src/pyk/kastManip.py
ehildenb: Well, but the `var_occurrences` method is used elsewhere, so just in terms of not re-implementing the same algorithm I figured to factor it out. I tihnk I'll leave as I have in the PR for now, don't think it will be a big inefficiency anyway, each of the variables already has memory allocated for it since it's coming from an existing term, and the list wrappers don't seem like that much.

src/pyk/kast.py
tothtamas28: You're right, this does not take into consideration that `{None, some_sort}` is not an error case.

src/pyk/kast.py
ehildenb: I fixed the spelling error (and several others). Prefer long variable names usually.

PR #94, opened 2022-11-11
Better automated unparsing

ehildenb:
This tries to automatically insert parentheses in a smarter way so that unparsing leads to fewer parsing ambiguities.

**TODO**: Try adding brackets/parens as a Kast -> Kast transformation ahead of time.


ehildenb:
This algorithm (bracketing around left or right recursive productions), gives this reduction in KEVM's symbol-table patch:

Screenshot from 2022-11-04 16-35-53

But it has pretty unsatisfactory results on IMP, as demonstrated in this PR. Probably better to go for an approach that is less aggressive with parens, and does not insert them when the surrounding context makes it possible not to.

ehildenb:
Closing, as it hasn't been pressing.

Issue #45, opened 2022-05-27
Compartmentalize kinner.py and kouter.py


ehildenb:
Currently, we have `kast.py` which declares both K inner ASTs, and K outer ASTs.

These can be separated into separate files, `kast/inner.py` and `kast/outer.py`. This will allow us to move the definitions of `TRUE`, `FALSE`, and `KTOKENDOTS` to another file which can be included by `kouter.py` so that `TRUE` is available for where it's used as a default.

That way `inner.py` is more purely just the actual AST definition, and some with `outer.py`.

PR #96, opened 2022-11-12
Split kast module into kinner and kouter

ehildenb:
This makes a refactoring I've wanted for a while to organize the kast and kastManip modules a bit.

- New subdirectory `src/pyk/kast`, which contains modules:
  - `kinner.py`: part of what used to be `src/pyk/kast.py` (`KInner` datastructures)
  - `kouter.py`: part of what used to be `src/pyk/kast.py` (`KOuter` datastructures)
  - `manip.py`: what used to be `src/pyk/kastManip.py`.
- Adjust imports throughout the repository.
- In our `k-files`, adjust the `functional` attribute to the new `total` attribute instead.


src/pyk/kast/kinner.py
```diff
@@ -0,0 +1,797 @@
...
+
+_LOGGER: Final = logging.getLogger(__name__)
+
+
+class KAst(ABC):
+    @classmethod
+    @abstractmethod
+    def from_dict(cls: Type[T], d: Dict[str, Any]) -> T:
+        node = d['node']
+        return globals()[node].from_dict(d)
```
tothtamas28, line 42: This separation `KInner` and `KOuter` breaks `from_dict`, e.g.

```python3
KAst.from_dict({
    'node': 'KRequire',
    'require': 'imp.k',
})
```

will now fail. I tried to do this before, but couldn't find a solution (https://github.com/runtimeverification/pyk/issues/45).

ehildenb: Yeah, but we never need to do `KAst.from_dict`, we only need to ever be able to do `KInner.from_dict` or `KOuter.from_dict`, but usually we even know more specifically which one we want (a few such adjustments are made in this PR, for example `KFlatModuleList.from_dict` or `KDefinition.from_dict`).

Any time you do `KAst.from_dict`, you immediately cast it to the more specific one you want anyway (so type-checking works), so you might as well just say `MORE_SPECIFIC.from_dict` anyway.

src/integration_tests/test_emit_json_spec.py
```diff
@@ -27,8 +28,7 @@ def setUp(self) -> None:
         _kprove(spec_file, kompiled_dir=self.kompiled_dir, emit_json_spec=self.spec_json_file, dry_run=True)
 
         with open(self.spec_json_file, 'r') as f:
-            kfml = KAst.from_dict(json.load(f)['term'])
-            assert type(kfml) is KFlatModuleList
+            kfml = KFlatModuleList.from_dict(json.load(f)['term'])
```
ehildenb, line 32: This is one such example where we know we'll always be reading a `KFlatModuleList`, and we were casting it to that one anyway.

src/pyk/kast/kouter.py
```diff
@@ -1849,12 +1145,6 @@ def build_cons(unit: KInner, label: Union[str, KLabel], terms: Iterable[KInner])
...
-    with open(path, 'r') as _f:
-        return KAst.from_dict(json.loads(_f.read())['term'])
-
-
 def read_kast_definition(path: Union[str, PathLike]) -> KDefinition:
-    _defn = read_kast(path)
-    assert isinstance(_defn, KDefinition)
-    return _defn
+    with open(path, 'r') as _f:
+        return KDefinition.from_dict(json.loads(_f.read())['term'])
```
ehildenb, line 1150: Here we get rid of the generic `read_kast` which does not work with the new type structure (this is the only function that broke). But `read_kast` only use was in `read_kast_definition`, at least here and in KEVM, and `read_kast_definition` can just use the more specific `KDefinition.from_dict`.

src/pyk/kast/kinner.py
tothtamas28: Then `from_dict` and `from_json` need to be removed from `KAst`.

ehildenb: or made virtual

tothtamas28: You're right, replacing the body of `from_dict` with `...` makes more sense. Let's do that instead.

ehildenb: Done.

src/pyk/kast/kinner.py
```diff
@@ -0,0 +1,797 @@
...
+from ..utils import EMPTY_FROZEN_DICT, FrozenDict, hash_str
+
+T = TypeVar('T', bound='KAst')
+W = TypeVar('W', bound='WithKAtt')
+KI = TypeVar('KI', bound='KInner')
+
+_LOGGER: Final = logging.getLogger(__name__)
+
+
+class KAst(ABC):
```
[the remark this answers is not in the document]
ehildenb: Note that using the same KOuter datastructures means that all the methods in `KDefinition` (for example) will have to be made over a new abstract `KInner2`, which is a `Union[KInner,KoreInner]`. Then queries like `KDefinition.rules` will work over both KAst and Kore.

tothtamas28 (approved):

src/integration_tests/kompiled_test.py
```diff
@@ -5,7 +5,8 @@
 from typing import Iterable, Optional
 from unittest import TestCase
 
-from pyk.kast import KDefinition, KInner
+from pyk.kast.kinner import KInner
+from pyk.kast.kouter import KDefinition
```
tothtamas28, line 9: What's your opinion about calling these modules `pyk.kast.inner` and `pyk.kast.outer` instead?

src/pyk/ktool/krun.py
```diff
@@ -50,8 +50,7 @@ def run(
         if result.returncode != 0:
             raise RuntimeError('Non-zero exit-code from krun.')
 
-        result_kast = KAst.from_dict(json.loads(result.stdout)['term'])
-        assert isinstance(result_kast, KInner)
+        result_kast = KInner.from_dict(json.loads(result.stdout)['term'])
```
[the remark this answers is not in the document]
ehildenb: But why? Do you have a use case for it? Issue created: https://github.com/runtimeverification/pyk/issues/97

src/integration_tests/kompiled_test.py
ehildenb: Sounds good, can do that.

ehildenb:
I'll note that this better reflects the way that the Java handles these classes too.

PR #99, opened 2022-11-15
Add initial --bug-report functionality

ehildenb:
I've tested this with KEVM, and it produces bug reports for `krun` and for `kore-rpc` that are reproducible _on my machine_. The `kore-rpc` bug reports should be reproducible across machines, but the `krun` ones may not be (machine specific binaries etc...).

This PR:

- Introduces the `BugReport` class.
  - This class tracks a tarball to attach bug reports to.
  - It exposes facilities for adding files/commands to the bug report, so that if bug reporting is turned on, it can collect those files/commands directly.


ehildenb:
@tothtamas28 I have a few questions about the architecture here:

- Would it make sense to _always_ produce a `BugReport` class (instead of it being `Optional`), but have it store a `bool` flag about whether the various methods should be no-ops or should actually be adding thnigs to the tarball? This would lift a conditional out of each call-sight, potentially making the code simpler.
- If we do always produce a `BugReport` class, maybe it should be fleshed out a bit more to handle all sorts of error cases too. In the Java frontend, we have a `KEManager` class (for "K Exception Manager"), which basically handles logging and reporting errors in a unified way. Should we go for a similar approach? Or keep it simple to `BugReport` for the moment.

src/pyk/ktool/krun.py
```diff
@@ -195,6 +206,14 @@ def _krun(
         no_expand_macros=no_expand_macros,
     )
 
+    if bug_report is not None:
```
tothtamas28, line 209: What do you think about pushing this down to `run_process`, similarly to how profiling and logging are supported there?

On error:

```python3
try:
    proc_res = run_process(...)
except RunProcessError as err:
    tar_path = ...
    with err.bug_report(tar_path) as report:
        report.add_file(...)
        report.add_dir(...)
    raise RuntimeError(f'krun crashed, wrote bug report to {tar_path}') from err
```

On success:

```python3
tar_path = ...
proc_res = run_process(...)
with proc_res.bug_report(tar_path) as report:
    ...
```

(I'm not sure we need this functionality for success. If we do, we should find a name other than `BugReport`.)

This also means that `BugReport` is kept at a more abstract level. Domain specific wrappers can be still introduced for convenience:

```python3
with err.bug_report(tar_path) as report:
    k_report = KBugReport(report)
    k_report.add_definition(...)
```

tothtamas28: > Would it make sense to always produce a BugReport class

In the design above, a context manager is instantiated from the process result / exception on demand. The `tar` is written to disk on `__exit__`. 

> Should we go for a similar approach? Or keep it simple to `BugReport` for the moment.

I think we should keep `BugReport` simple and generic. Then we can wrap it in various useful ways.

tothtamas28:
> the method `Path.is_relative_to` [is not available until Python 3.9](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_relative_to). Should I rework the code to avoid using this functionality, or should we require >= Python 3.9?

Let's implement `is_relative_to(Path)` under `cli_utils` for now. But also, we should consider bumping Python version (perhaps even to `3.10`).

src/pyk/ktool/krun.py
ehildenb: We do want it for successful cases, for reproducing results which may not lead to crashes.

But I think we shouldn't try to make it generic in `run_process`. That was my first attempt, but then we lose the ability to do something like `add_definition`, which has slightly different logic/binning than `add_file`, for example. Maybe if we instead had `run_k_process`, where the first argument was always a `definition_dir`, then it could work....

I tried to internalize the logic of whether to record to the bug report or not to the BugReport class, but this also did not work out super well because then we have to make a _different_ BugReport instance for different contexts (for example, `KoreClient` has to have a separate `BugReport` instance to `KRun`), which seems like it will be harder to make it thread-safe. Not sure if we should worry about it being thread-safe yet.

I also want to avoid that someone _must_ include a bug reporter in their frontend. I wonder how that could be made to happen, maybe by integrating the bug report functionality directly into `KDefinition` or `KPrint`?

tothtamas28: > We do want it for successful cases

Makes sense.

> But I think we shouldn't try to make it generic in run_process. That was my first attempt, but then we lose the ability to do something like `add_definition`

But is this not a solution for that issue?

```python3
with err.bug_report(tar_path) as report:
    k_report = KBugReport(report)
    k_report.add_definition(...)
```

Maybe I missed some important detail, but to me this seems like a problem that can be handled nicely in an abstract way, as `BugReport`  could just be replaced by context manager for an archive. In a given context, we can initialize / wrap instances in various ways, e.g. `RunProcessError.bug_report` could return a context manager where a shell script, `stdout`, `stderr`, stack trace, etc. have already been added to the context manager. That instance can then be wrapped in some other class that provides `add_definition`. What do you think?

ehildenb: Not sure what I remember of the context of this conversation, but I'm pretty happy with the current functionality, it seems to behave as I expect. I've attached a bug report produced by turning on bug reporting on the tests added here: https://github.com/runtimeverification/pyk/pull/157

ehildenb:
@ana-pantilie and @jberthold here is a new RPC-style bug report that this PR can generated when using the RPC prover: 
[bug-reports.tar.gz](https://github.com/runtimeverification/pyk/files/10429555/bug-reports.tar.gz)

Can you confirm that this contains the information you need to reproduce the RPC interaction? First you should unpack the files (in a scratch directory), then run the commands in `commands/NNN.sh` in ascending order:

```
tar -xvf bug-reports.tar.gz
bash commands/000.sh &  # background this, it starts the kore-rpc
bash commands/001.sh    # should print out the result, which should match file results/001.json
bash commands/002.sh    # should print the result, which should match file results/002.json
...
```

jberthold:
> Can you confirm that this contains the information you need to reproduce the RPC interaction? First you should unpack the files (in a scratch directory), then run the commands in `commands/NNN.sh` in ascending order:

Works for me, except the `nc` invocations in the commands 001 - 007 do not close the socket/terminate automatically.
`nc -N` will close the socket automatically when the input reaches EOF.

I ran 
```
for num in $(seq -f "%03g" 1 7); do 
    echo $num
    nc -Nv localhost 52509 < rpc/${num}_request.json | diff -b -s rpc/${num}_response.json -
done
```
(obviously the `7` is specific to the example tarball).
Output 
```
001
Connection to localhost (127.0.0.1) 52509 port [tcp/*] succeeded!
Files rpc/001_response.json and - are identical
002
Connection to localhost (127.0.0.1) 52509 port [tcp/*] succeeded!
Files rpc/002_response.json and - are identical
003
...
```
(the `-b` in the diff is required because the response files do not have a final newline).

ehildenb:
@jberthold here is another tarball to try out: 
[bug-reports.tar.gz](https://github.com/runtimeverification/pyk/files/10430226/bug-reports.tar.gz)

It gives this output:

image

The reason to not build a single script is because we want to have it be that partial results (from crashes) still results in valid tarballs.

jberthold:
I think the tarball is the same as the one before... but your code change looks good :+1:

src/pyk/cli_utils.py
```diff
@@ -115,3 +122,55 @@ def run_process(
 
 def gen_file_timestamp(comment: str = '//') -> str:
     return comment + ' This file generated by: ' + sys.argv[0] + '\n' + comment + ' ' + str(datetime.now()) + '\n'
+
+
+class BugReport:
+    _bug_report: Path
+    _command_id: int
+    _defn_id: int
```
tothtamas28, line 130: Later on we'll probably need further categories, so consider dropping these attributes in favor of a more general solution.

E.g. `add_file` / `add_file_contents` could take some sort of name template:

```python3
def add_file(self, path: Path, template: Optional[str] = None) -> str:
    ...

def add_file_contents(self, content: str, template: Optional[str] = None) -> str:
    ...
```

Then in `add_command` you can:

```python3
self.add_file_contents(..., 'commands/{}.sh')
```

This also makes `add_definition` unnecessary.

src/pyk/cli_utils.py
```diff
@@ -115,3 +122,55 @@ def run_process(
...
+    _file_remap: Dict[str, str]
```
tothtamas28, line 131: Do we have a use case that makes this remapping necessary? In this PR `add_command` seems to always be called in a setting where the file names are known.

src/pyk/ktool/krun.py
```diff
@@ -28,8 +28,9 @@ def __init__(
         use_directory: Optional[Path] = None,
         profile: bool = False,
         command: str = 'krun',
+        bug_report: Optional[BugReport] = None,
```
tothtamas28, line 31: Is the intended use of a `BugReport` instance to collect all debug artifacts for multiple function calls during execution?

tothtamas28 (approved):

src/pyk/cli_utils.py
ehildenb: It's to relativize paths. So instead of getting, for example, `kore-rpc --definition /tmp/pyk-testing/fully/path/to/kompile/definition.kore ...`, the command that is saved is `kore-rpc --definition kompiled/definition.kore ...`, so the commands can be run without having to adjust a bunch of paths from whatever system the original was run on.

src/pyk/ktool/krun.py
ehildenb: It's to be able to fully replay a given scenario/interaction with the backends without having to do any setup. Note that an interaction may be many calls to kprove/kore-rpc/krun, so we need to save them all when it's enabled.

ehildenb: That can be used for providing reproduction instructions in a bug report, for example, which is where the name comes from (the haskell backend has supported this feature internally for a while, this is pulling the same idea into the frontend so it's supported in all backends).

src/pyk/kore/rpc.py
```diff
@@ -103,21 +106,44 @@ def request(self, method: str, **params: Any) -> Dict[str, Any]:
...
+            self._bug_report.add_command(
+                ['diff', '-b', '-s', f'rpc/{self._req_id:03}_actual.json', f'rpc/{self._req_id:03}_response.json']
+            )
+
         data = json.loads(resp)
         self._check(data)
 
-        assert data['id'] == old_id
+        assert data['id'] == self._req_id
+        self._req_id += 1
```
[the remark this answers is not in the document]
ehildenb: Fixed.

src/pyk/cli_utils.py
ehildenb: I'm not really sure we'll need more categories, there's just definitions and commands and rpc calls right now (kompile, kprove, krun, kore-rpc).

PR #105, opened 2022-11-20
Handle LeftAssoc and RightAssoc in kore_to_kast

ehildenb:
This PR:

- Adds `Assoc.pattern` and defines it for `LeftAssoc` and `RightAssoc`, which resolves to the full associative pattern.
- Adds (previously failing) tests of `kore_to_kast` on a `LeftAssoc` and a `RightAssoc` pattern.
- Adds the ability for `KPrint.kore_to_kast` to convert `LeftAssoc` and `RightAssoc` patterns (using `Assoc.pattern`).


src/integration_tests/test_kore_to_kast.py
```diff
@@ -227,3 +227,55 @@ def test_kast_to_kore(self) -> None:
...
+                    )
+                ),
+                KApply(
+                    '_Map_',
+                    [
+                        KApply('_Map_', [KVariable('X', sort=KSort('Map')), KVariable('Y', sort=KSort('Map'))]),
+                        KVariable('Z', sort=KSort('Map')),
+                    ],
+                ),
+            ),
```
ehildenb, line 254: I actually can't find any documentation if this is the correct associative order. That is, `LeftAssoc` means that parenthesis should be nested around left-more elements. @dwightguth can you confirm?

src/pyk/kore/syntax.py
```diff
@@ -1345,6 +1350,15 @@ def let(self, *, app: Optional[App] = None) -> 'LeftAssoc':
     def map_pattern(self: 'LeftAssoc', f: Callable[[Pattern], Pattern]) -> 'LeftAssoc':
         return self
 
+    @property
+    def pattern(self) -> Pattern:
+        assert len(self.app.sorts) == 0
+        assert len(self.app.patterns) > 0
```
tothtamas28, line 1356: I think these should raise `ValueError` instead.

src/pyk/kore/syntax.py
```diff
@@ -1324,6 +1324,11 @@ class MLSyntaxSugar(MLPattern):
 class Assoc(MLSyntaxSugar):
     app: App
 
+    @property
+    @abstractmethod
+    def pattern(self) -> Pattern:
```
tothtamas28, line 1329: What do you think about `desugared`?

tothtamas28 (approved):

src/pyk/kore/syntax.py
ehildenb: I don't like `pattern` very much, but I think I like `desugared` even less though. We can change it later, or add a wrapper later.

src/pyk/kore/syntax.py
ehildenb: Done.

src/integration_tests/test_kore_to_kast.py
ehildenb: @radumereuta perhaps you know?

dwightguth: Left associativity does indeed mean that nested elements occur under the left side.

PR #112, opened 2022-11-23
Steps towards RPC prover

ehildenb:
These changes are what I needed to have a first working version of KProve implemented fully here: https://github.com/runtimeverification/evm-semantics/pull/1443. In particular:

- Some KAst <-> Kore problems are addressed:
  - If the same variable shows up multiple times in `KDefinition.sort_vars`, but with different sorts, instead of just giving up we'll check if one sort is a subsort of the other. If so, we'll return the smaller sort. Test added to `test_kast_to_kore`.
  - The conversions for `Exists` are added. Test added to `test_bidirectional`.
- `KProve.execute` method now returns `CTerm` instead of `KInner`, as it shouldn't be able to return `mlTop()` or `mlBottom()` anyway.
- `KCFG.create_cover` now allows you to specify a witness manually instead of trying to compute it for you. (De)Serialization requires that we can do `Subst.{from_dict,to_dict}`, so that functionality is added too.
- `KProve.implies` method is added as a wrapper around the `implies` RPC endpoint. Tests are added that make sure it's giving back sane results. By default, we much existentially bind variables in the consequent that are not present in the antecedent to get anything useful out of the endpoint. Tests added to `test_implies` in `test_kprove`.


src/pyk/kcfg.py
```diff
@@ -95,20 +95,27 @@ class Cover(EdgeLike):
         subst: Subst
         constraint: KInner
 
-        def __init__(self, source: 'KCFG.Node', target: 'KCFG.Node'):
+        def __init__(self, source: 'KCFG.Node', target: 'KCFG.Node', witness: Optional[Tuple[Subst, KInner]] = None):
```
tothtamas28, line 98: Consider passing `subst` and `constraint` separately (with the necessary precondition checks).

src/pyk/kcfg.py
```diff
@@ -95,20 +95,27 @@ class Cover(EdgeLike):
...
-            subst, constraint = match_res
             object.__setattr__(self, 'subst', subst)
             object.__setattr__(self, 'constraint', constraint)
 
         def to_dict(self) -> Dict[str, Any]:
-            return {'source': self.source.id, 'target': self.target.id}
+            return {
+                'source': self.source.id,
+                'target': self.target.id,
+                'witness': {'subst': self.subst.to_dict(), 'constraint': self.constraint.to_dict()},
```
tothtamas28, line 117: Consider eliminating `witness` and storing `subst` and `constraint` directly under the root (that is more in line with the object structure).

tothtamas28 (approved):

src/pyk/kast/inner.py
```diff
@@ -92,6 +92,20 @@ def __mul__(self, other: 'Subst') -> 'Subst':
...
+        for k, v in d.items():
+            if type(k) is str:
+                _subst[k] = KInner.from_dict(v)
+        return Subst(_subst)
+
+    def to_dict(self) -> Dict[str, Any]:
+        _res = {}
+        for k, v in self._subst.items():
+            _res[k] = v.to_dict()
+        return _res
```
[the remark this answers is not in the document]
ehildenb: Done.

src/pyk/kast/inner.py
```diff
@@ -92,6 +92,20 @@ def __mul__(self, other: 'Subst') -> 'Subst':
     def __call__(self, term: KInner) -> KInner:
         return self.apply(term)
 
+    @classmethod
+    def from_dict(cls: Type['Subst'], d: Dict[str, Any]) -> 'Subst':
+        _subst = {}
+        for k, v in d.items():
+            if type(k) is str:
+                _subst[k] = KInner.from_dict(v)
+        return Subst(_subst)
```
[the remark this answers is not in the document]
ehildenb: Done.

src/pyk/kcfg.py
ehildenb: Done.

src/pyk/kcfg.py
ehildenb: Done.

PR #116, opened 2022-11-28
Refactor `kompile` and `KompiledTest`

tothtamas28:


src/integration_tests/kompiled_test.py
```diff
@@ -2,49 +2,57 @@
...
+    KOMPILE_BACKEND: ClassVar[Optional[str]] = None
+    KOMPILE_MAIN_MODULE: ClassVar[Optional[str]] = None
+    KOMPILE_SYNTAX_MODULE: ClassVar[Optional[str]] = None
+    KOMPILE_INCLUDE_DIRS: ClassVar[Iterable[str]] = []
+    KOMPILE_POST_PROCESS: ClassVar[Optional[str]] = None
 
-    kompiled_dir: Path
-    definition: KDefinition
+    kompiled_dir: ClassVar[Path]
+    definition: ClassVar[KDefinition]
```
tothtamas28, line 24: This ensures that kompilation occurs once per `KompiledTest` subclass, which results in a noticeable speedup for integration tests (around 30%).

Hopefully it won't cause issues for test parallelization. If it does, we can easily revert this change.

ehildenb (approved):

src/integration_tests/kompiled_test.py
ehildenb: Looks like it only affects the tests, rather than the main `kompile` method.

tothtamas28: Yes, the change here is that now `kompile` runs before the first test only, and not before each test.

Issue #82, opened 2022-11-04
Better handling of RPC client/server resources


ehildenb:
Our current handling of the RPC client/server resources is not amazing, it should be improved.

- `KoreClient` and `KoreServer` should implement `__del__` functionality, so that upon them being cleaned up they properly shut down their servers appropriately.
- `KProve` should properly clean up the `KoreClient` and `KoreServer` that it starts when the `execute` endpoint is called.
- Users should not have to specify the port to start a client/server on, and it should be able to pick one that does not conflict with other applications on the machine.

Goal is that when someone needs an RPC server:

- It is started automatically (even if they need multiple servers) without the user having to do any initialization. For example, on the first call to `KProve.execute`, the client and server need to be booted.
- It is cleaned up automatically (when the holding object goes out of scope, or the application crashes, or application ends normally).
- It is not needing to be restarted on subsequent calls (it persists for as long as needed).

Issue #76, opened 2022-10-26
Implement kast_to_kore and kore_to_kast


ehildenb:
Several projects are using `kast_to_kore` and `kore_to_kast` to feed terms directly into the backend. Currently these routines call the `kast` binary, but in many cases we can have a simple and efficient direct translation in Python between the [KAst datastructures](https://github.com/runtimeverification/pyk/blob/master/src/pyk/kast.py) and the [Kore datastructures](https://github.com/runtimeverification/pyk/blob/master/src/pyk/kore/syntax.py). This will facilitate faster communication with the backends, and eventually FFI.

At the moment, it seems the only "documentation" of how conversion between KAst and Kore works is here: https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java. This is obviously not desirable, and leads to us believing that KAst is somehow an "ambiguous" format (though I personally disagree with this).

We'll begin by implementing just the cases needed for:

- [ ] Translating constraints that show up during [symbolic execution of KEVM](https://github.com/runtimeverification/hs-backend-booster/issues/3).
- [ ] Translating the components of KAVM configurations that are needing to go between KAst and Kore.

ehildenb:
There is also a munger/unmunger here that I'm using as reference: https://github.com/runtimeverification/k/blob/382a862f0fe5fd1ca00c4c5d187939db2b4b0506/kore/src/main/scala/org/kframework/parser/kore/parser/KoreToK.scala

PR #118, opened 2022-11-28
Better sort inference, Kast/Kore conversions, check-implication behavior

ehildenb:
Fixes: #82
Fixes: #76 

This PR makes steps towards RPC prover in KEVM. In particular:

- `KCFG` updates:
  - Adds an option `omit_large_subst=True` to `KCFG.pretty` which is threaded through to `KCFG.Edge.pretty`.
  - Removes unused methods `_is_rewrite_edge` and `_is_case_split_edge`.
- Updates to Kast <-> Kore conversions:
  - Tests (previously failing) of `#Top`, `#Bottom`, `#Implies`, and a more complicated `#Exists` are added.
  - `KVariable.let_sort` is added, which lets you update just the sort information of a given variable.
  - `KDefinition.sort_vars_subst` now also takes into account variable contexts (the symbol that shows up above a variable) when finding least sorts for each variable.
  - `KProduction.argument_sorts` is added, and `KDefinition.argument_sorts` makes use of this.
  - When retrieving a production for a klabel, the canonical one is selected by ignoring ones (with a warning) that have the `unparseAvoid` attribute. This is needed for [`#Top` and `#Bottom`](https://github.com/runtimeverification/k/blob/master/k-distribution/include/kframework/builtin/kast.md#syntax-of-matching-logic), which have multiple representations.
  - Conversions for `#Top`, `#Bottom`, and `#Implies` are added.
  - Adds smart constructors to `prelude.py` for `mlExists` and `mlCeil`.
  - Switches over `kore_to_kast` to use only the smart constructors.
- RPC prover changes (`KProve` class):
  - Test (previously failing) of `check_implication` on trivial subsumption is called.
  - Methods `KProve.__enter__`, `KProve.__exit__`, and `KProve.close` are added for better handling of RPC server.
  - Method `KProve.set_kore_rpc_port` is added for updating the Kore RPC port used.
  - Debugging messages added to each RPC endpoint in `KProve` (for better timing/debugging info).
  - Adds a warning when `check_implication` returns a non-trivial implication (this warning currently _always_ fires, because the backend does not minimize the implications it gives back).
  - Adds an error when `check_implication` returns a non-empty predicate back (currently not triggered yet on any examples I have, this will have to be changed as we get more examples).
  - Handles the case where `mlTop()` is returned as the substitution.


src/pyk/ktool/kprint.py
```diff
@@ -311,13 +328,26 @@ def _kore_to_kast(self, kore: Pattern) -> Optional[KInner]:
...
             rarg = self._kore_to_kast(kore.right)
             if larg is not None and rarg is not None:
                 return KApply(KLabel('#And', [psort]), [larg, rarg])
 
+        elif type(kore) is Implies:
+            psort = KSort(kore.sort.name[4:])
+            larg = self._kore_to_kast(kore.left)
+            rarg = self._kore_to_kast(kore.right)
+            if larg is not None and rarg is not None:
+                return KApply(KLabel('#Implies', [psort]), [larg, rarg])
```
tothtamas28, line 349:
```suggestion
                return mlImplies(larg, rarg, psort)
```

src/pyk/ktool/kprint.py
```diff
@@ -435,6 +465,15 @@ def _get_sort(_ki: KInner) -> Optional[KSort]:
                             _and = self._add_sort_injection(_and, psort, sort)
                         return _and
 
+                elif kast.label.name == '#Implies' and kast.arity == 2:
```
tothtamas28, line 468:
```suggestion
                elif kast.label.name == '#Implies':
                    assert kast.arity == 2
```

src/pyk/ktool/kprove.py
```diff
@@ -142,6 +143,12 @@ def __init__(
             self.main_module = mm.read()
         self._kore_rpc = None
 
+    def __enter__(self) -> 'KProve':
```
tothtamas28, line 146: You can make this class

```python3
class KProve(KPrint, ContextManager['KProve']):
```

src/pyk/ktool/kprove.py
```diff
@@ -154,6 +161,18 @@ def close_kore_rpc(self) -> None:
             _kore_server, _kore_client = self._kore_rpc
             _kore_client.close()
             _kore_server.close()
+            self._kore_rpc = None
+
+    def set_kore_rpc_port(self, p: int) -> None:
```
tothtamas28, line 166: What's the use case for this method?

src/pyk/ktool/kprove.py
```diff
@@ -360,17 +382,27 @@ def implies(self, antecedent: CTerm, consequent: CTerm, bind_consequent_variable
...
+            return Subst({})
         subst_pattern = mlEquals(KVariable('###VAR'), KVariable('###TERM'))
         _subst: Dict[str, KInner] = {}
         for ml_pred in flatten_label('#And', ml_subst):
             m = subst_pattern.match(ml_pred)
             if m is not None and type(m['###VAR']) is KVariable:
                 _subst[m['###VAR'].name] = m['###TERM']
             else:
-                raise ValueError(f'Recieved back a non-substitution from implies endpoint: {ml_pred}')
+                raise ValueError(f'Received a non-substitution from implies endpoint: {ml_pred}')
```
tothtamas28, line 405: If this assumption is part of the contract, this can even be an `AssertionError` I think.

```suggestion
                raise AssertionError(f'Received a non-substitution from implies endpoint: {ml_pred}')
```

tothtamas28 (approved):

src/pyk/ktool/kprove.py
ehildenb: In KEVM, we have [this parallelism](https://github.com/runtimeverification/evm-semantics/pull/1443/files#diff-8ef99db40b2010c0b478f54603fda2b277c332347b8945df0e83bb686a6eafc1R466).

We start a bunch of parallel processes to begin discharging proofs. But we already launched an RPC server [here](https://github.com/runtimeverification/evm-semantics/pull/1443/files#diff-8ef99db40b2010c0b478f54603fda2b277c332347b8945df0e83bb686a6eafc1R353).

So the pickler passes through the `KProve` object to each different parallel process, but then they are all trying to talk to the same Kore RPC server.

This allows each process to [set its own RPC server port, and restart the server automatically](https://github.com/runtimeverification/evm-semantics/pull/1443/files#diff-8ef99db40b2010c0b478f54603fda2b277c332347b8945df0e83bb686a6eafc1R367). Probably a better solution is to make the RPC server "parallel ready" itself, but that is not what it currently gives us.

src/pyk/ktool/kprove.py
ehildenb: Instead of doing `__enter__` and `__exit__` or in addition to that?

src/pyk/ktool/kprint.py
ehildenb: Done, I changed them all.

src/pyk/ktool/kprove.py
tothtamas28: In addition, just to declare that this class has `__enter__` and `__exit__`.

src/pyk/ktool/kprint.py
ehildenb: Done, I changed them all (and added `mlExists` and `mlCeil`).

src/pyk/ktool/kprove.py
ehildenb: Done.

src/pyk/ktool/kprove.py
ehildenb: Done.

PR #122, opened 2022-11-30
Proper implication handling

ehildenb:
This PR makes a few changes for rpc proving:

- `KProve.implies` endpoint returns both the substitution and the predicate (tests are updated).
~- `KCFG` now stores a `_splits` variable, which records when a given nodes is split into several other nodes.~
~- Frontier calculation is updated.~
~- Node removal is updated.~
~- We now use `Cover` for storing case-splits, as that is what they are.~
- Updates to the kit-tests:
  - Allow passing through `CHECK=...` through from main makefile to submakefile.
  - Update the expected output (for the new case splits) for sum-to-n test.


src/pyk/kcfg.py
```diff
@@ -238,6 +240,7 @@ def to_dict(self) -> Dict[str, Any]:
         expanded = sorted(self._expanded)
         verified = [{'source': source_id, 'target': target_id} for source_id, target_id in sorted(self._verified)]
         aliases = dict(sorted(self._aliases.items()))
+        splits = dict(sorted([(k, sorted(v)) for k, v in self._splits.items()]))
```
tothtamas28, line 243:
```suggestion
        splits = dict(sorted((k, sorted(v)) for k, v in self._splits.items()))
```

src/pyk/kcfg.py
```diff
@@ -581,15 +590,14 @@ def split_node(self, source_id: str, constraints: Iterable[KInner]) -> List[str]
...
         source = self.node(source_id)
 
-        def _add_case_edge(_constraint: KInner) -> str:
-            _cterm = CTerm(mlAnd([source.cterm.kast, _constraint]))
+        def _add_case_cover(_constraint: KInner) -> str:
+            _cterm = source.cterm.add_constraint(_constraint)
             _node = self.get_or_create_node(_cterm)
-            self.create_edge(source.id, _node.id, _constraint, 0)
-            self.add_verified(source.id, _node.id)
+            self.create_cover(source.id, _node.id, constraint=_constraint)
```
tothtamas28, line 596: Shouldn't the split node be the target of each of these covers? If I remember correctly, `source` is the covered node and `target` is the one that covers.

tothtamas28 (approved):

src/pyk/kcfg.py
ehildenb: Yeah, that's true.... Also means that graph traversals and pretty printing needs to be changed.

ehildenb: Maybe splits should be stored as their own type of edge then.

ehildenb: I've removed the splits from this PR, only kept the implication checking.

src/pyk/kcfg.py
ehildenb: Done.

src/pyk/kcfg.py
ehildenb: https://github.com/runtimeverification/pyk/pull/124

Issue #121, opened 2022-11-29
Factor out Kast <-> Kore conversions


ehildenb:
We currently have `KPrint.kast_to_kore` and `KPrint.kore_to_kast`. These methods have to be in `KPrint` because they require a `KDefinition`, and fall back to binary `kast` when the methods `KPrint._kast_to_kore` and `KPrint._kore_to_kast` fail to do the conversions.

Now that `KPrint._kast_to_kore` and `KPrint._kore_to_kast` have matured, we should try to factor them out of `KPrint` class. In particular:

- [ ] (Optional) Split [kore.syntax](https://github.com/runtimeverification/pyk/blob/master/src/pyk/kore/syntax.py) into Inner and Outer similar to how Kast is (reduce file size, make it easier to do targeted imports).
- [x] Add a `injections` table to [`Definition`](https://github.com/runtimeverification/pyk/blob/master/src/pyk/kore/syntax.py#L1899) similar to the subsorts table in [`KDefinition`](https://github.com/runtimeverification/pyk/blob/master/src/pyk/kast/outer.py#L839). It should scan the `Definition` and find Injection productions and store them as a dictionary with the key being the supersort of the injection and the values being the various subsorts it maps to. #126 
- [x] Implement `Definition.add_sort_injections`, which traverses a given Kore term and adds sort injections where needed (or fails with value error if no injections exist for the given sort pair don't exist). This should basically replace [`KPrint._add_sort_injection`](https://github.com/runtimeverification/pyk/blob/master/src/pyk/ktool/kprint.py#L552). May want to implement `bottom_up` on `Pattern` to facilitate this. https://github.com/runtimeverification/pyk/pull/134
- [ ] Add to each `KInner` subclass a method `KInner.kore`, which implements the logic of [`KPrint._kast_to_kore`](https://github.com/runtimeverification/pyk/blob/master/src/pyk/ktool/kprint.py#L403) directly, except it doesn't add any sort injections.
- [ ] Implement a new method `KInner.from_kore`, which reproduces [`KPrint._kore_to_kast`](https://github.com/runtimeverification/pyk/blob/master/src/pyk/ktool/kprint.py#L292).
- [ ] Re-implement `KPrint.{kast_to_kore,kore_to_kast}` in terms of these new methods, with the difference being taht we use `KInner.kore` for producing the kore, and then do a second traversal to add the injections.

tothtamas28:
[...]

PR #126, opened 2022-12-02
Implement KORE subsort table

tothtamas28:
Related: https://github.com/runtimeverification/pyk/issues/121

Extract a table for the subsort relation from `definition.kore` based on the `subsort` attribute.


src/pyk/kompile.py
```diff
@@ -0,0 +1,57 @@
...
+        timestamp_file = definition_dir / 'timestamp'
+        check_file_path(timestamp_file)
+        timestamp = timestamp_file.stat().st_mtime_ns
+
+        object.__setattr__(self, 'path', path)
+        object.__setattr__(self, 'timestamp', timestamp)
+
+    @cached_property
+    def definition(self) -> Definition:
+        return KoreParser(self.path.read_text()).definition()
```
ehildenb, line 35: We should store this so we don't read it more than once. Reading/parsing definitions takes a few seconds.

src/pyk/kompile.py
```diff
@@ -0,0 +1,57 @@
...
+    # Strict subsort, i.e. not reflexive
+    def _is_subsort(self, subsort: Sort, supersort: Sort) -> bool:
+        return subsort in self._subsort_dict.get(supersort, set())
+
+
+def _subsort_dict(definition: Definition) -> Dict[Sort, Set[Sort]]:
+    axioms = (axiom for module in definition for axiom in module.axioms)
+    attrs = (attr for axiom in axioms for attr in axiom.attrs)
+    subsort_attrs = (attr for attr in attrs if attr.symbol == 'subsort')
+    subsort_attr_sorts = (attr.sorts for attr in subsort_attrs)
```
ehildenb, line 51: I would prefer if we inspected the axioms directly rather than rely on the attributes.

src/tests/kore/test-data/definitions/pass/test-axiom-subsort.kore
```diff
@@ -0,0 +1,12 @@
...
+
+module TEST-AXIOM-SUBSORT
+    axiom{R}
+        \exists{R} (
+            Val:SortAExp{},
+            \equals{SortAExp{}, R} (
+                Val:SortAExp{},
+                inj{SortId{}, SortAExp{}} (From:SortId{})
+            )
+        ) [subsort{SortId{}, SortAExp{}}()] // subsort
```
ehildenb, line 11: Hmmmm, I see that inpsecting the axiom is somewhat difficult. We could also just assert the structure of the axiom, after we have found the subsort attribute. Basically check that the axiom has the expected form.

ehildenb: But I guess that will be slower.

src/tests/test_kompile.py
```diff
@@ -1,43 +1,31 @@
-from pathlib import Path
+from pyk.kompile import _subsort_dict
+from pyk.kore.syntax import Attr, Axiom, Definition, Module, Sort, SortApp, SortVar, Top
 
-from pyk.ktool.kompile import KompileBackend, _build_arg_list
 
+def test_subsort_dict() -> None:
+    def sort_axiom(subsort: Sort, supersort: Sort) -> Axiom:
+        r = SortVar('R')
+        return Axiom((r,), Top(r), attrs=(Attr('subsort', (subsort, supersort)),))
```
ehildenb, line 8: Ah, here we build the sort_axiom anyway, I guess it's not too hard to construct.

ehildenb (approved):
We should definitely only read the definition in once (but continue to do so lazily like we are). There will eventually be many places we do `defn.definition` later (we do so a lot in the Kast datastructures), and we don't want to re-read it each time.

I would _prefer_ that we inspected the axioms directly for determining what type of axiom it is, but relying on the attributes is OK.

PR #128, opened 2022-12-04
Improve Kast -> Kore for KSequence, fix bytestoken quoting

ehildenb:
We had the incorrect enquoting of bytes/string tokens before, and also the incorrect conversion from Kast -> Kore for KSequence. This PR addresses that:

- Corrections to bytes/string literals:
  - `enquoteKString` is partially ported from [the Java implementation](https://github.com/runtimeverification/k/blob/8b2e215fb46901b68532f9f3cb9656bfed7ed504/kore/src/main/java/org/kframework/utils/StringUtil.java#L224) to here.
  - ~`bytesToken` and `stringToken` store their tokens directly, and not with the surrounding quotes/decorators. This makes conversion to Kore more straightforward.~ See https://github.com/runtimeverification/k/issues/3071
  - `bytesToken` and `stringToken` are correctly enquoted when pretty-printing (test fixed to reflect this for bytesToken).
- Corrections to KSequence -> kseq:
  - `KSequence` is taken into account as a sort context in `KDefinition.sort_vars`.
  - The fact that `kseq` is _not_ associative, but `KSequence` is associative, is now correctly reflected in `KPrint._kast_to_kore` conversion.
- Improvements to KCFG pretty printing:
  - Option `omit_large_subst` renamed to `minimize`, and set to `True`.
  - Also minimize large constraints on subsumption edges, to avoid printing really big `Ceil` conditions in KEVM.

The proof I've been working with in KEVM (to do RPC proving instead of `kprove` interface), is now passing!


src/pyk/utils.py
```diff
@@ -230,3 +230,32 @@ def compare_short_hashes(lhs: str, rhs: str) -> bool:
...
+            if 32 <= ord_c and ord_c < 127:
+                return orig_c
+            elif ord_c < 16:
+                return '\\x0' + hex(ord(orig_c))[2:]
+            elif ord_c <= 255:
+                return '\\x' + hex(ord(orig_c))[2:]
+            else:
+                raise ValueError(f'Unsupported character for enquoting: {orig_c}')
+
+    return ''.join(enquote_char(c) for c in orig_s)
```
tothtamas28, line 261: You can use raw strings here.

```suggestion
def enquote_str(orig_s: str) -> str:
    def enquote_char(orig_c: str) -> str:
        if orig_c == '"':
            return r'\"'
        elif orig_c == '\\':
            return r'\\'
        elif orig_c == '\n':
            return r'\n'
        elif orig_c == '\t':
            return r'\t'
        elif orig_c == '\r':
            return r'\r'
        elif orig_c == '\f':
            return r'\f'
        else:
            ord_c = ord(orig_c)
            if 32 <= ord_c and ord_c < 127:
                return orig_c
            elif ord_c < 16:
                return r'\x0' + hex(ord(orig_c))[2:]
            elif ord_c <= 255:
                return r'\x' + hex(ord(orig_c))[2:]
            else:
                raise ValueError(f'Unsupported character for enquoting: {orig_c}')

    return ''.join(enquote_char(c) for c in orig_s)
```

tothtamas28 (approved):

PR #132, opened 2022-12-07
Add parameter `command` to `KoreServer.__init__`

tothtamas28:
Enables

```python
server = KoreServer('k-files/imp-kompiled', module_name='IMP', port=3000, command='hs-backend-booster')
```


ana-pantilie (approved):
Thank you!

PR #134, opened 2022-12-08
Refactor KAST-to-KORE transformation

tothtamas28:
~Blocked on: #186~

Introduces:

* `kast_to_kore`: Naive KAST-to-KORE transformation where each `EVar.sort` is fixed based on the corresponding `KVariable.sort` if provided, or otherwise will be `SortK{}`. Hence in general the transformation only results in a well-typed KORE pattern if all variables in the KAST term are already annotated with sorts. Does not require a kompiled definition.
* `KompiledKore.add_injections`: adds sort injections.


src/pyk/kompile.py
```diff
@@ -35,16 +55,191 @@ def definition(self) -> Definition:
...
+
+    def is_subsort(self, sort1: Sort, sort2: Sort) -> bool:
+        if sort1 == sort2:
+            return True
+
+        if sort2 == SortApp('SortK'):
+            return True
+
+        if sort1 == SortApp('SortK'):
+            return False
```
ehildenb, line 114: We shouldn't need to special case these ones at all, in fact it's a bug if we do need to. `compiled.txt` and `definition.kore` should already have the subsorts into sort K added for every sort that _should_ have that subsort.

src/pyk/kompile.py
```diff
@@ -35,16 +55,191 @@ def definition(self) -> Definition:
...
+
+    def meet_sorts(self, sort1: Sort, sort2: Sort) -> Sort:
+        if sort1 == sort2:
+            return sort1
+
+        if sort1 == SortApp('SortK'):
+            return sort2
+
+        if sort2 == SortApp('SortK'):
+            return sort1
```
ehildenb, line 126: Once again, should not have to special case sort K.

Also, this meet sorts algorithm, have you found it's actually needed in practice? I would put a check at the beginning that says if `sort1 in substorts(sort2): return sort1` and the reverse situation too. That's a much faster check that works in all the cases we've seen in KEVM so far than actually computing a lattice meet.

src/pyk/kompile.py
```diff
@@ -55,3 +250,248 @@ def _subsort_dict(definition: Definition) -> Dict[Sort, Set[Sort]]:
...
+    if type(kast) is KToken:
+        return _ktoken_to_kore(kast)
+    elif type(kast) is KVariable:
+        return _kvariable_to_kore(kast)
+    elif type(kast) is KSequence:
+        return _ksequence_to_kore(kast)
+    elif type(kast) is KApply:
+        return _kapply_to_kore(kast)
+
+    raise ValueError(f'Unsupported KAst: {kast}')
```
ehildenb, line 303:
```suggestion
    raise ValueError(f'Unsupported KInner: {kast}')
```

Perhaps?

src/pyk/kompile.py
```diff
@@ -55,3 +250,248 @@ def _subsort_dict(definition: Definition) -> Dict[Sort, Set[Sort]]:
...
+
+    return 'Lbl' + munge(label)
+
+
+# --------------
+# Symbol munging
+# --------------
+
+
+UNMUNGE_TABLE: Final[FrozenDict[str, str]] = FrozenDict(
```
ehildenb, line 400: Can we remove the munge/unmunge table in `kprint.py` now too?

src/pyk/kompile.py
```diff
@@ -35,16 +55,191 @@ def definition(self) -> Definition:
...
+                    scope[pattern.name].add(sort)
+                    scope[pattern.name].add(pattern.sort)
+                else:
+                    root[pattern.name] = {sort, pattern.sort}
+
+            else:
+                for subpattern, subsort in zip(pattern.patterns, self.pattern_sorts(pattern)):
+                    collect(scope, subpattern, subsort)
+
+        def rewrite(scope: Scope[Sort], pattern: Pattern) -> Pattern:
```
ehildenb, line 206: This is because we don't have substitutions at the Kore level?

src/pyk/kompile.py
```diff
@@ -55,3 +250,248 @@ def _subsort_dict(definition: Definition) -> Dict[Sort, Set[Sort]]:
...
+        'Pipe': '|',
+        'RBra': '}',
+        'Tild': '~',
+    }
+)
+
+MUNGE_TABLE: Final[FrozenDict[str, str]] = FrozenDict({v: k for k, v in UNMUNGE_TABLE.items()})
+
+
+def munge(label: str) -> str:
```
ehildenb, line 440: We should steal the implementation from `kprint.py` directly. It was not trivial to get right, there are unexpected corner cases. Also should steal the tests from there as well.

src/pyk/utils.py
```diff
@@ -75,6 +78,77 @@ def __repr__(self) -> str:
 EMPTY_FROZEN_DICT: Final[FrozenDict] = FrozenDict()
 
 
+class Scope(MutableMapping[str, T]):
```
ehildenb, line 81: I see this is used in the tests, is it used elsewhere?

ehildenb: Ah I see it's used in the inlined rewrite function of kast_To_kore.

ehildenb:
I still want to see this broken up.

- No need to bring in Kore-level sort inference yet, and the sort inference algorithm used there is pretty high powered when it probably does not need to be (as evidenced that a much simpler inference algorithm has been working on the original kast_to_kore just fine). Also, it appears that the algorithm there will benefit from having a notion of substitutions, because it shouldn't have to do any rewriting (substitutions are simpler).
- Move the munging/unmunging _out_ of `KPrint` class, and also make the `KPrint.kast_to_kore` just a stub function that:
  - Calls the original sort inference at the kast level, then
  - Calls the new kast_to_kore in kompile, then
  - Calls the new sort injection pass in kompile.

That alone will be a large change to review.

src/pyk/kompile.py
tothtamas28: Earlier we had a discussion with @radumereuta about this. Right now, `K` is implicitly top sort, and `Bottom` is implicitly bottom sort. Should those relations be made explicit in K?

src/pyk/kompile.py
tothtamas28: A substitution in itself would not solve the problem here. `Scope` has a hierarchy of dictionaries so that symbols can be resolved in their respective scope. So for example, in

```
\and{S} (X:S, \exists{S} (X:S, X:S))
```

symbol `X` might have to resolve to separate values in the two conjuncts.

src/pyk/kompile.py
tothtamas28: Will do that in the `munge`/`unmunge` PR I pull from this one.

src/pyk/kompile.py
radumereuta: For some reason, Dwight decided not to include the subsorts to K in the generated kore file: https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L232

It may be related to the fact that user syntax should be considered KItem. But when doing parsing, type inferencing and other things like calculating injections we want to know about this relation.

src/pyk/kompile.py
tothtamas28: Done.

src/pyk/kompile.py
tothtamas28: Done.

src/pyk/kompile.py
```diff
@@ -35,16 +55,191 @@ def definition(self) -> Definition:
...
 
+        else:
+            sorts = ()
 
-def _subsort_dict(definition: Definition) -> Dict[Sort, Set[Sort]]:
+        assert len(sorts) == len(pattern.patterns)
+        return sorts
+
+    def kast_to_kore(self, kast: KInner, *, sort: Optional[Sort] = None, with_inj: bool = False) -> Pattern:
+        if not sort:
```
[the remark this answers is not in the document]
tothtamas28: Done.

src/pyk/konvert.py
```diff
@@ -35,16 +55,188 @@ def definition(self) -> Definition:
...
+            r'\bottom': SymbolDecl(Symbol(r'\bottom', (S,)), (), S),
+            r'\not': SymbolDecl(Symbol(r'\not', (S,)), (S,), S),
+            r'\and': SymbolDecl(Symbol(r'\and', (S,)), (S, S), S),
+            r'\or': SymbolDecl(Symbol(r'\or', (S,)), (S, S), S),
+            r'\implies': SymbolDecl(Symbol(r'\implies', (S,)), (S, S), S),
+            r'\iff': SymbolDecl(Symbol(r'\iff', (S,)), (S, S), S),
+            r'\ceil': SymbolDecl(Symbol(r'\ceil', (S, T)), (S,), T),
+            r'\floor': SymbolDecl(Symbol(r'\floor', (S, T)), (S,), T),
+            r'\equals': SymbolDecl(Symbol(r'\equals', (S, T)), (S, S), T),
+            r'\in': SymbolDecl(Symbol(r'\in', (S, T)), (S, S), T),
```
ehildenb, line 71: Why do we need to hardcode these symbols? Are they not present in the kore definitions we receive?

src/pyk/ktool/kprint.py
```diff
@@ -342,195 +330,25 @@ def _kore_to_kast(self, kore: Pattern) -> Optional[KInner]:
...
-            else:
-                item0_kore = self._kast_to_kore(item0, sort=KSort('KItem'))
-                items_kore = self._kast_to_kore(KSequence(kast.items[1:]), sort=KSort('K'))
-                if item0_kore is not None and items_kore is not None:
-                    return App('kseq', (), [item0_kore, items_kore])
-
-        _LOGGER.warning(f'KPrint._kast_to_kore failed on input: {kast}')
-        return None
+        if sort is None:
+            sort = K
```
ehildenb, line 334: I think in general we'll find that `GeneratedTopCell` is a more useful default sort, but we can make that adjustment later.

src/pyk/konvert.py
```diff
@@ -57,6 +249,150 @@ def _subsort_dict(definition: Definition) -> Dict[Sort, Set[Sort]]:
...
 
+def _symbol_table(definition: Definition) -> Dict[str, SymbolDecl]:
+    symbol_decls = (symbol_decl for module in definition for symbol_decl in module.symbol_decls)
+    return {symbol_decl.symbol.name: symbol_decl for symbol_decl in symbol_decls}
+
+
+# ------------
+# KAST-to-KORE
+# ------------
+
```
ehildenb, line 260: Everything before this (the `KompiledKore` especially), feels like it belongs in `src/pyk/kore/outer.py` (or `definition.py` if naming gets confusing) instead of here. It's all Kore specific stuff, similar to how `outer.py` is kast specific stuff. `konvert.py` should be just what's needed to go back and forth between the two formats.

src/pyk/konvert.py
```diff
@@ -35,16 +55,188 @@ def definition(self) -> Definition:
...
+        assert len(sorts) == len(pattern.patterns)
+        return sorts
+
+    def kast_to_kore(self, kast: KInner, *, sort: Optional[Sort] = None, with_inj: bool = False) -> Pattern:
+        if sort is None:
+            sort = SortApp('SortK')
+
+        pattern = kast_to_kore(kast)
+        pattern = self._strengthen_sorts(pattern, sort)
```
ehildenb, line 175: Do we have any tests that break without this call?

Right now `KPrint` is handling sorting variables. I would prefer if this file (`konvert.py`) was also passed the Kast definition (this file should be the bridge between Kast and Kore formats). So then this function should be pulled out of `KompiledDefinition`, and just be a flat utility which recieves a `KDefiniition` (Kast) and a `Definition` (Kore), and reads:

- Sort variables (at Kast level): `kast_defn.sort_vars(kast)`
- Call simple kast_to_kore (no Kast or Kore definition needed). `_kast_to_kore(kast)`
- Call sort injection: `kore_defn.add_injections(kore)`

That will be closer to a straight refactoring of the current codebase. Then you can remove kore variable sort inference from this PR (which perhaps we can remove all the `Scope` stuff too then?). I'd rather keep the change small. Note that right now this kore sort variable inference is a more expensive algorithm than the kast sort variable inference we're doing, and almost certainly is actually not doing any more sort inference than the original algorithm (or else the tests would be failing). So we're doing more work for nothing.

ehildenb:
Overall feedback:

- Move `KompiledDefinition` utilities to `Definition` class in `syntax.py`. Consider pulling out `Definition` and other "outer-like" classes into their own module, similar to how we do for the Kast data-structures.
- Make `konvert.kast_to_kore` a simple function which takes as input a `KDefinition` (Kast), and a `Definition` (Kore).
- Pull out the Kore sort inference, and the `Scope` class.
- No need for the extra test harness yet, just adapt it to use the underlying `konvert.kast_to_kore` instead of a `KPrint`.
- Wire `KPrint` to use this method directly, then fallback to `kast` binary.

The idea is that _all_ Kore data-structure pure things should be in `kore/` module, and all Kast should be in `kast/` module. The only file which should mention both will be `konvert.py`.

ehildenb:
If you do decide to factor out an outer-like module for Kore, that probably deserves it's own PR.

tothtamas28:
> Move most of the `KompiledDefinition` utilities to `Definition` class in `syntax.py`.

My problem with merging `KompiledKore` into `Definition` is that `KompiledKore` relies on the particular encoding of the subsort relation that the frontend produces, and this change gives the false impression that the methods work on an arbitrary `Definiton` instance. So in this case I think wrapping a `Definition` read from disk is more appropriate.

ehildenb:
Well, we are really only trying to support definitions that come from the frontend for the moment.... So I guess that's OK.

But we can also change the way we read in subsorts later if that's needed, right? Is the problem that you don't like relying on the attribute `subsort`? Could instead look for injections in the definition I suppose. But I think it's another case of "if it's working (the `subsort` attribute), then assume it will keep working until it doesn't, then we make a test for it".

But if you'd like to keep this functionality separate, that's fine. I would think it still belongs in the `kore/` subdirectory though, because it has to do with Kore definitions from disk specifically. Then `konvert.py` can just be about converting between the two formats (and be passed in all the information it needs to be able to do that).

tothtamas28:
> Is the problem that you don't like relying on the attribute `subsort`?

Basically yes, in particular I don't like that `kompile` attaches it to an axiom like

```
axiom{R} \exists{R} (Val:SortKItem{}, \equals{SortKItem{}, R} (Val:SortKItem{}, inj{SortString{}, SortKItem{}} (From:SortString{}))) [subsort{SortString{}, SortKItem{}}()]
```

which, when instantiating a `Definition` manually, you'd need to reproduce just so that `pyk` can read the `subsort` attribute and build the sort lattice. (Alternatively, you can add dummy axioms instead, but that's not better.) So I figured, let's restrict type inference to definitions read from disk to avoid this, an emphasize it with a wrapper class.

I guess the situation would be better (i.e. less noisy) if `kompile` had a `supersorts` or `subsorts` attribute attached directly to the `sort` sentence. But that on the other hand would break modularization I guess, e.g. module `INT` would look different based on the modules it is imported from.

> I would think it still belongs in the kore/ subdirectory though

I'm fine with that.

src/pyk/konvert.py
tothtamas28: These are matching logic symbols, not theory symbols, and thus are not declared in the definition. Adding them to a common symbol table is useful though as it makes working with ML patterns easier.

tothtamas28:
@ehildenb I addressed the review comments above.
* Dropped KORE sort inference (from this PR, see below)
* Extracted part of `KPrint.kast_to_kore` to `pyk.konvert.kast_to_kore`
* Moved methods not dealing with the subsort table from `KompiledKore` to `Definition`
* Moved `KompiledKore` to `pyk.kore.kompiled`
* Changed `TestKoreToKast` to use `pyk.konvert` instead of `KPrint`

Moreover, I pulled out KORE sort inference for further discussion: https://github.com/runtimeverification/pyk/pull/189. The PR addresses the following point.

> Note that right now this kore sort variable inference is a more expensive algorithm than the kast sort variable inference we're doing, and almost certainly is actually not doing any more sort inference than the original algorithm (or else the tests would be failing). So we're doing more work for nothing.

* The first commit adds a simple version of KORE sort inference - comparable to KAST sort inference as implemented - that induces a variant of KAST-to-KORE which does not rely on `KDefinition` while still passes all tests.
* The second commit adds a test case that breaks both variants of KAST-to-KORE; then replaces the simple sort inference algorithm with the conceptually (but not computationally) more complex one, which fixes the failing test.

src/pyk/konvert.py
```diff
@@ -1,61 +1,166 @@
...
-    timestamp: int
 
-    def __init__(self, definition_dir: Union[str, Path]):
-        definition_dir = Path(definition_dir)
-        check_dir_path(definition_dir)
+def _kapply_to_ml_quant(kapply: KApply) -> MLQuant:
+    label = kapply.label
+    symbol = ML_QUANT_LABELS[label.name]
+    sorts = tuple(_ksort_to_kore(ksort) for ksort in label.params)
+    (_,) = sorts
```
ehildenb, line 132: What is this line doing?

src/pyk/konvert.py
```diff
@@ -1,61 +1,166 @@
...
+    if kvar.sort:
+        sort = _ksort_to_kore(kvar.sort)
+    else:
+        sort = SortApp('SortK')
+    return EVar('Var' + munge(kvar.name), sort)
+
+
+def _ksequence_to_kore(kseq: KSequence) -> Pattern:
+    if not kseq:
+        return App('dotk')
```
ehildenb, line 104:
```suggestion
    if len(kseq) == 0:
        return App('dotk')
```

I prefer this more explicit check.

ehildenb: Same elsewhere, I prefer not relying on "truthiness" of values, simplify because then I can't tell looking at this line if it's something that I should expect to be an optional type or a iterable/list-ish type.

src/pyk/konvert.py
```diff
@@ -1,61 +1,166 @@
...
+def _kapply_to_ml_quant(kapply: KApply) -> MLQuant:
+    label = kapply.label
+    symbol = ML_QUANT_LABELS[label.name]
+    sorts = tuple(_ksort_to_kore(ksort) for ksort in label.params)
+    (_,) = sorts
 
-        path = (definition_dir / 'definition.kore').resolve()
-        check_file_path(path)
+    kvar, kast = kapply.args
+    var = _kast_to_kore(kvar)
```
ehildenb, line 135: We should probably be asserting that `type(Var) is EVar` here or something? Even though the syntax of Kore allows a pattern in the first position, it's definitely not supported in any tools or in the logic.

ehildenb (approved):

ehildenb:
I love how this looks now! Left a few comments.

src/pyk/konvert.py
tothtamas28: Fails if `len(sorts) != 1`.

src/pyk/konvert.py
tothtamas28: `MLQuant.of(symbol, sorts, patterns)` checks this.

src/pyk/konvert.py
tothtamas28: This is a recommendation in [PEP8](https://peps.python.org/pep-0008/#programming-recommendations).

```python3
# Correct:
if not seq:
if seq:
```

```python3
# Wrong:
if len(seq):
if not len(seq):
```

But I see your point: if `kseq` is `Optional`, then `mypy` flags `len(kseq) == 0` as a type error, so the reviewer can be confident that there is no bug coming from an unhandled `None` case. In this case though, I think the context is simple enough to follow the recommendation.

src/pyk/konvert.py
ehildenb: I guess I would prefer tha tto be `assert len(sorts) == 1` then, which is more explicit. I usually don't like fancy short-hand :sweat_smile:

tothtamas28: Dropped that line altogether (realized that `MLQuant.of` checks this anyways).

PR #142, opened 2022-12-10
Test of execute endpoint with cell map

ehildenb:
There was a bug in how KEVM was handling cell maps, and I could not figure it out. So I made a minimal example to reproduce the bug, and the last commit here demonstrates the fix.

So somehow the kore terms being generated made it past validation in the backend (even without the wrapper mapitem labels), but this last commit shows how to correctly build the terms so they can execute properly.


src/integration_tests/test_cell_map_proof.py
```diff
@@ -0,0 +1,92 @@
...
+        1,
+        ('#accountNonexistent(1)', 'SetItem(1)', [('1', '2')]),
+        1,
+        ('false', 'SetItem(1)', [('1', '2')]),
+        [],
+    ),
+)
+
+
+class TestSimpleProof(KProveTest):
```
tothtamas28, line 26:
```suggestion
class TestCellMap(KProveTest):
```

src/integration_tests/test_cell_map_proof.py
```diff
@@ -0,0 +1,92 @@
...
+import pytest
+
+from pyk.cterm import CTerm
+from pyk.kast.inner import KApply, KInner, KSequence, KToken, KVariable, build_assoc
+from pyk.kast.manip import get_cell
+from pyk.ktool import KProve
+
+from .utils import KProveTest
+
+State = Tuple[str, str, Iterable[Tuple[str, str]]]
```
tothtamas28, line 12: After

```python3
from typing import NamedTuple
```

you can

```suggestion
class State(NamedTuple):
    pgm: str
    active_accounts: str
    accounts: Iterable[Tuple[str, str]]
    
```

tothtamas28 (approved):

src/integration_tests/test_cell_map_proof.py
```diff
@@ -0,0 +1,96 @@
...
+    accounts: Iterable[Tuple[str, str]]
+    
+
+EXECUTE_TEST_DATA: Final[Iterable[Tuple[str, int, State, int, State, Iterable[State]]]] = (
+    (
+        'account-nonexistent',
+        1,
+        ('#accountNonexistent(1)', 'SetItem(1)', [('1', '2')]),
+        1,
+        ('false', 'SetItem(1)', [('1', '2')]),
```
ehildenb, line 24:
```suggestion
        State('#accountNonexistent(1)', 'SetItem(1)', [('1', '2')]),
        1,
        State('false', 'SetItem(1)', [('1', '2')]),
```

PR #151, opened 2022-12-16
Add a few useful methods to `Pattern`

tothtamas28:
Namely: `patterns` (property), `let_patterns`, `bottom_up`, `top_down`.


ehildenb:
Can you add a simple test? An example would be a routine that does [`var_occurrences`](https://github.com/runtimeverification/pyk/blob/e98c91c79f4c240b465c25d2bbee28782112e887/src/pyk/kast/inner.py#L662) but on `Kore`.

ehildenb (approved):

PR #157, opened 2023-01-13
RPC based prover for KCFG exploration

ehildenb:

PR #158, opened 2023-01-17
Factor out KCFGExplore class

ehildenb:
This PR broadly remove RPC server handling from the KProve class. This refactoring has been pulled out of this PR: https://github.com/runtimeverification/pyk/pull/157

- Kore RPC log messages about requests and responses are downgraded from `info` to `debug`.
- `main_module` and `backend` fields are moved from `KProve/KRun` classes to `KPrint` class.
- `KCFGExplore` class is introduced which manages the RPC server:
  - All uses of RPC server are removed from `KProve` and moved to this class (`execute`, `implies`, and `simplify`).
  - This class now gets the `port` at construction time, and stays with the same port throughout execution (different than how `KProve` class used to let you set the port and restart the server).
- Tests are updated to reflect new structure:
  - New folder `src/integration_tests/kcfg` is added.
  - Tests which exercised the RPC server are moved into this subdirectory.


src/pyk/kcfg/explore.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def close_kore_rpc(self) -> None:
+        if self._kore_rpc is not None:
+            _kore_server, _kore_client = self._kore_rpc
+            _kore_client.close()
+            _kore_server.close()
+            self._kore_rpc = None
+
+    def close(self) -> None:
+        self.close_kore_rpc()
```
tothtamas28, line 52: Consider inlining this call and removing `close_kore_rpc`.

src/pyk/kcfg/explore.py
```diff
@@ -0,0 +1,119 @@
...
+        self._bug_report = bug_report
+        self._kore_rpc = None
+
+    def __enter__(self) -> 'KCFGExplore':
+        return self
+
+    def __exit__(self, *args: Any) -> None:
+        self.close()
+
+    def kore_rpc(self) -> Tuple[KoreServer, KoreClient]:
```
tothtamas28, line 35: Ideally this method should not be part of the public interface. Moreover, after `close` calls to `kore_rpc` should raise an exception.

tothtamas28 (approved):

src/pyk/kcfg/explore.py
ehildenb: Done.

src/pyk/kcfg/explore.py
ehildenb: Done.

PR #160, opened 2023-01-20
Expect non-zero return code from `krun`

Baltoli:
I have a use-case for the C semantics / FRET project where I'd like to be able to use `krun` via Pyk to take a few steps from an initial configuration. Our semantics initializes the return-code cell with a non-zero value so that we can identify when rewriting stops before a program has completed execution; this causes `krun` to raise an exception when run via Pyk.

This PR adds an extra option to the `KRun` run methods that allows a specific return code to be expected in such situations, as well as an integration test-case for this.


src/integration_tests/test_krun.py
```diff
@@ -101,3 +103,20 @@ def test_run_kore_config(self, krun: KRun) -> None:
 
         # Then
         assert actual == expected
+
+
+class TestReturnCodeRun(KRunTest):
+    KOMPILE_MAIN_FILE = 'k-files/return-code.k'
+
+    def _input(self, value: int) -> KToken:
```
tothtamas28, line 111:
```suggestion
    @staticmethod
    def _input(value: int) -> KToken:
```

src/pyk/ktool/krun.py
```diff
@@ -58,10 +59,11 @@ def run(
                 pmap=pmap,
                 profile=self._profile,
                 bug_report=self._bug_report,
+                check=(expect_rc == 0),
             )
 
-        if result.returncode != 0:
-            raise RuntimeError('Non-zero exit-code from krun.')
+        if result.returncode != expect_rc:
+            raise RuntimeError('Unexpected exit-code from krun.')
```
tothtamas28, line 66:
```suggestion
            raise RuntimeError(f'Expected {expect_rc} as exit-code for krun, got: {result.returncode}')
```

tothtamas28: I see this check has multiple occurrences, consider extracting a method:

```python3
@staticmethod
def check_returncode(actual: int, expected: int) -> None:
    raise RuntimeError(f'Expected {expected} as exit-code for krun, got: {actual}')
```

src/pyk/ktool/krun.py
```diff
@@ -38,6 +38,7 @@ def run(
         config: Optional[Mapping[str, KInner]] = None,
         depth: Optional[int] = None,
         expand_macros: bool = False,
+        expect_rc: int = 0,
```
tothtamas28, line 41: Maybe this can be generalized to `expect_rc: Union[int, Iterable[int]] = 0`.

tothtamas28 (approved):
LGTM.

Baltoli:
Thanks @tothtamas28, much appreciated. I'll extract the check out into its own method as you suggest, and generalise it to take an iterable at the same time. I'll also update the test to reflect that.

src/pyk/ktool/krun.py
ehildenb: Seems to me this method should belong in `cli_utils.py`. We can extract it later though. It would have to take `krun` as an argument.

ehildenb: Or perhaps just built into `run_process` in `cli_utils.py`?

Baltoli:
I've addressed your comments @tothtamas28. @ehildenb I think that extracting the code into `cli_utils` is worth doing but outside the scope of this PR perhaps - I'll open an issue.

PR #161, opened 2023-01-20
Add class `Chainable` and related KORE utilities

tothtamas28:


tothtamas28:
This let's us define functions that navigate and extract data from KORE patterns by combining a few simple building blocks:


```python
def get_locations(pattern: Pattern) -> Optional[Tuple[int, int, int, int]]:
    def locations(pattern: Pattern) -> Tuple[int, int, int, int]:
        return (chain >> args(1, 2, 3, 4) >> tuple_of(kore_int, kore_int, kore_int, kore_int))(pattern)

    return (
        chain
        >> app("Lbl'-LT-'generatedTop'-GT-'")
        >> arg(0)
        >> app("Lbl'-LT-'clarity'-GT-'")
        >> arg(18)
        >> app("Lbl'-LT-'currentLoc'-GT-'")
        >> arg(0)
        >> case_symbol(
            (LBL_EMPTY_LOCATIONS, none),
            (LBL_LOCATIONS, locations)
        )
    )(pattern)
```

tothtamas28:
Related PR: https://github.com/runtimeverification/algoclarity-repl/pull/32

src/pyk/kore/match.py
```diff
@@ -99,11 +78,105 @@ def match_map(
...
+    ...
+
+
+@overload
+def args(n1: int, n2: int, n3: int, /) -> Callable[[App], Tuple[Pattern, Pattern, Pattern]]:
+    ...
+
+
+@overload
+def args(n1: int, n2: int, n3: int, n4: int, /) -> Callable[[App], Tuple[Pattern, Pattern, Pattern, Pattern]]:
```
ehildenb, line 127: What is this last argument `/` meaning?

src/pyk/kore/match.py
```diff
@@ -99,11 +78,105 @@ def match_map(
...
+    ...
+
+
+def args(*ns: int) -> Callable[[App], Tuple]:
+    def res(app: App) -> Tuple:
+        return tuple(app.args[n] for n in ns)
+
+    return res
+
```
ehildenb, line 136: Would it be possible to have `def args(*ss: str) -> Callable[[App], Tuple]:`, and it grabs all subarguments that match a given symbol?

Then we could select sub-cells by name instead of by index, which would be more robust against refactorings in the semantics, and more readable as well.

ehildenb: I understand this would be a bit slower, but I think the readability is worth it here. Also, it should still be faster than the `get_cell` method over Kast, which makes a call to `split_config_from`.

ehildenb:
Please add a testing harness and tests which exercise each type of chaining operator.

sskeirik (commented):
I agree with Everett's comments --- I too am curious about the meaning of the slash.

I think the `arg` combinator that works by index or by name is especially cool. Thanks for that.

A few tests, even just as examples to demonstrate how the combinators should be used, would be great.

src/pyk/kore/match.py
tothtamas28: Parameters preceding this symbol are [positional-only parameters](https://peps.python.org/pep-0570/).

src/pyk/kore/match.py
```diff
@@ -99,11 +74,185 @@ def match_map(
...
+def arg(id: Union[int, str]) -> Callable[[App], Union[Pattern, App]]:
+    def res(app: App) -> Union[Pattern, App]:
+        if type(id) is int:
+            if len(app.args) <= id:
+                raise ValueError('Argument index is out of range')
+
+            return app.args[id]
+
+        try:
+            arg, *_ = (arg for arg in app.args if type(arg) is App and arg.symbol == id)
```
ehildenb, line 124: This looks like we could have it be multiple elements, but that we'll just take the first? I think we should be asserting that there is only a single argument match (that `len(_) == 0`).

ehildenb (approved):

PR #167, opened 2023-01-23
Variables store their sorts directly

ehildenb:
Instead of storing sorts as KVariable attributes, we store them as a field directly on the class.

- Packing variable sorts as attributes (and unpacking) happens at `{from|to}_dict` time, instead of as a property.
- The `att` field is removed completely from `KVariable`.
- The tests are updated accordingly.


src/pyk/kast/inner.py
```diff
@@ -233,50 +233,36 @@ def match(self, term: KInner) -> Optional[Subst]:
...
+        _att = KAtt({})
+        if self.sort is not None:
+            _att = _att.update({KAtt.SORT: self.sort.to_dict()})
+        return {'node': 'KVariable', 'name': self.name, 'att': _att.to_dict()}
 
-    def let(
-        self, *, name: Optional[str] = None, sort: Optional[KSort] = None, att: Optional[KAtt] = None
-    ) -> 'KVariable':
-        # TODO: We actually want `sort: Optional[Optional['KSort']]`
+    def let(self, *, name: Optional[str] = None, sort: Optional[KSort] = None) -> 'KVariable':
```
tothtamas28, line 259: Right now, `var.let(sort=None) == var`, even if `var.sort is not None`. But with some custom decorator it should be possible to distinguish `var.let()` from `var.let(sort=None)`. Should we create an issue for this?

src/pyk/kast/inner.py
```diff
@@ -233,50 +233,36 @@ def match(self, term: KInner) -> Optional[Subst]:
 
 @final
 @dataclass(frozen=True)
-class KVariable(KInner, WithKAtt):
```
tothtamas28, line 236: Is `sort` the only variable attribute?

tothtamas28 (approved):

src/pyk/kast/kast.py
```diff
@@ -125,6 +126,9 @@ def let(self, *, atts: Optional[Mapping[str, Any]] = None) -> 'KAtt':
     def update(self, atts: Mapping[str, Any]) -> 'KAtt':
         return self.let(atts={k: v for k, v in {**self.atts, **atts}.items() if v is not None})
 
+    def remove(self, remove_atts: Iterable[str]) -> 'KAtt':
+        return KAtt({k: v for k, v in self.atts.items() if k not in remove_atts})
```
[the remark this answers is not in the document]
ehildenb: Done.

src/pyk/utils.py
```diff
@@ -58,6 +58,11 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return f'FrozenDict({repr(self._dict)})'
 
+    def __lt__(self, other: 'FrozenDict') -> bool:
+        if len(self._dict) == 0 and len(other._dict) == 0:
```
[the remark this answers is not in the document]
ehildenb: I just removed `__lt__` implementation altogether from `FrozenDict`, since it's no longer needed when `KVariable` does not hold a `KAtt`.

src/pyk/kast/inner.py
```diff
@@ -233,50 +233,36 @@ def match(self, term: KInner) -> Optional[Subst]:
...
-        if sort is not None:
-            att = att.update({KAtt.SORT: None})
-        return KVariable(name=name, sort=sort, att=att)
-
-    def let_att(self, att: KAtt) -> 'KVariable':
-        return self.let(att=att)
+        sort = sort if sort is not None else self.sort
+        return KVariable(name=name, sort=sort)
 
     def let_sort(self, sort: KSort) -> 'KVariable':
```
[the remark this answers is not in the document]
ehildenb: done.

src/pyk/kast/inner.py
ehildenb: I'm actually skeptical this functionality will ever be used. Here is the issue: https://github.com/runtimeverification/pyk/issues/168

src/pyk/kast/inner.py
ehildenb: It's the only one in our test-suite! I can add an assertion that we don't read any others from the frontend, so at least we know about it if there are others.

PR #170, opened 2023-01-27
Cell map sort inference

ehildenb:
Currently, we don't do very good sort inference on a variable that could sort as `*CellMap` or as `Map`.

This PR:

- Makes a refactoring to the sort inference to pull out the `_greatest_common_subsort` method as standalone.
- Moves tests of `kast_to_kore` and `kore_to_kast` into their own subdirectory, so we can have tests against multiple definitions.
- Adds tests of `kast_to_kore` for a term with a `*CellMap` variable showing up in a context where it could also be a `Map`, which is currently failing.


ehildenb:
Closing, as it hasn't been pressing.

PR #173, opened 2023-01-27
Improvements to KCFG exploration and TUI

ehildenb:
This PR adds several improvements to the UI and capabilities of KCFGExplore, including:

- Additional proof exploration strategies added to KCFGExplore:
  - `KCFGExplore.step` will expand a given node by 1 step, it can either be a frontier node, or a node with a single successor, which will break the edge of depth N coming out of the node into an edge of depth 1, and an edge of depth `N - 1`.
  - `KCFGExplore.section_edge` will take as input a given edge, and some number of new sections you want to break that edge into, and it will find the intermediate nodes to split that edge into multiple edges. So for example if you have an edge of length 99, and ask it to bisect it into 4 pieces, it will replace it with 5 edges, the first 4 of length 24 and the last one of length 3.
- Bug in implication endpoint fixed, and test added.
  - Slack discussion here: https://runtimeverification.slack.com/archives/C045G8HCK99/p1674730840244839
  - Issue submitted to Haskell backend here: https://github.com/runtimeverification/haskell-backend/issues/3469
  - Quieted down a warning that is not needed: https://github.com/runtimeverification/haskell-backend/issues/3466
- Beginning of KCFG management brought into KCFGExplore class:
  - Static methods `read_cfg` and `write_cfg` which both take a directory, and a KCFG identifier, and decide where to store/read that KCFG from the directory.
  - Eventually this can be extended to take a database object instead of a directory, for example.
- Interaction from TUI, including:
  - Press `h` after selecting a node to hide it from behavior view.
  - Press `H` to unhide all nodes.


ehildenb:
Hiding and unhiding nodes:

[Screencast from 01-27-2023 02:20:47 PM.webm](https://user-images.githubusercontent.com/12721972/215203350-274ae4a0-28f1-41e3-8370-10f22952e28b.webm)

tothtamas28 (approved):

PR #174, opened 2023-01-28
Cache properties in `KFlatModule` and `KDefinition`

### LINKED ISSUE #127: kast.outer.KDefinition is manually caching many fields
Because of my misunderstanding about how `@cached_property` decorator works, I avoided it for class `kast.outer.KDefinition`. So there are many fields which are `@property` decorated which I manually am caching the results of.

These should be switched to `@cached_property` instead, and the manual caching removed.

tothtamas28:
Fixes #127


src/pyk/kast/outer.py
```diff
@@ -755,38 +756,40 @@ def __init__(
...
-        ]
-        _functions = [f for f in _functions if not (f.klabel and _is_non_free_constructor(f.klabel.name))]
-        return _functions
-
-    @property
-    def constructors(self) -> List[KProduction]:
-        return [prod for prod in self.syntax_productions if prod not in self.functions]
+        return ('function' in prod.att.atts or 'functional' in prod.att.atts) and (
+            not prod.klabel or not is_non_free_constructor(prod.klabel.name)
+        )
```
tothtamas28, line 784: This refactoring is not completely trivial, please check.

src/pyk/kast/outer.py
```diff
@@ -957,47 +960,48 @@ def let(
...
+        return tuple(rule for rule in self.rules if 'macro' in rule.att) + self.alias_rules
+
+    @cached_property
+    def semantic_rules(self) -> Tuple[KRule, ...]:
+        def is_semantic(rule: KRule) -> bool:
+            return (type(rule.body) is KApply and rule.body.label.name == '<generatedTop>') or (
+                type(rule.body) is KRewrite
+                and type(rule.body.lhs) is KApply
+                and rule.body.lhs.label.name == '<generatedTop>'
+            )
```
tothtamas28, line 1002: This is also nontrivial, please check.

tothtamas28:
I also tried to add the `lru_cache` decorator to methods like `production_for_klabel`, but `flake8-bugbear` flags that as a potential memory leak:

```
B019 Use of `functools.lru_cache` or `functools.cache` on methods can lead to memory leaks. The cache may retain instance references, preventing garbage collection.
```

which makes sense. Here's a potential [workaround](https://stackoverflow.com/a/33672499) though. Should we add this decorator to `pyk.utils`?

src/pyk/kast/outer.py
```diff
@@ -755,38 +756,40 @@ def __init__(
     def __iter__(self) -> Iterator[KSentence]:
         return iter(self.sentences)
 
-    @property
-    def productions(self) -> List[KProduction]:
-        return [sentence for sentence in self.sentences if type(sentence) is KProduction]
+    @cached_property
+    def productions(self) -> Tuple[KProduction, ...]:
+        return tuple(sentence for sentence in self if type(sentence) is KProduction)
```
ehildenb, line 761: Remind me why we're preferring `Tuple` in general to `List`? Just for immutability?

src/pyk/kast/outer.py
ehildenb: Hmmmm, I'm actually looking at this now, and I'm not sure why `CellMapItem`, `SetItem`, and `ListItem` are considered non-free-constructors.... I would think they are free constructors. But I do remember having a conversation with Dwight about it.

ehildenb: Here is a discussion about it: https://github.com/runtimeverification/k/issues/2683

ehildenb: Ah, it's because they are marked as `function` in the prelude, but we don't want to consider them for defunctionalization.... Maybe a better name for this method is `is_fake_function`, or `is_not_actually_function`.

ehildenb: Also,  I would leave it stated as `not (prod.klabel is not None and is_not_actually_function(prod.klabel.name))`, since the first check of `prod.klabel` is just to guard doing a looking of a `None`, so somehow reading them as `not (label exists    _and_   predicate on label)` seems clearer than `not label exists        or     not predicate on label`

Either that, or lift `is_not_actually_function` to take type `KProduction` instead of type `str`, and just pass it the entire `prod`, which would make this boolean condition clearer.

src/pyk/kast/outer.py
ehildenb: It looks good to me!

src/pyk/kast/outer.py
ehildenb: The logic looks good to me, I'd just make the above adjustments.

src/pyk/kast/outer.py
tothtamas28: Yes, the main difference between `Tuple[T, ...]` and `List[T]` is that the first is immutable and the second is not. Here, we need immutability because we cache the value.

src/pyk/kast/outer.py
tothtamas28: Done.

ehildenb (approved):

PR #184, opened 2023-02-03
Handle bottom antecedent in cterm_implies better

ehildenb:
Blocked on: https://github.com/runtimeverification/haskell-backend/issues/3484

This is to have better reporting of when branches trivially go to `#Bottom` when doing all-path reachability proofs. We should have a way of signaling that the antecedent to an `implies` check is `#Bottom` to the consumer directly.


ehildenb:
Subsumed: https://github.com/runtimeverification/pyk/pull/190

PR #185, opened 2023-02-03
Allow overriding unparser for KPrint/KRun/KProve

ehildenb:
https://github.com/runtimeverification/pyk/pull/183 changes the way that we cache properties in the `KPrint` class (to use more standard Python features). But this breaks some functionality in KEVM where we were manually overriding the unparsing definition and unparsing symbol table.

Here is a proposed solution:

- Thread optional arguments `definition` and `patch_symbol_table` into each constructor.
- If these arguments are present, ignore the definition on disk and use these ones.

Here is where I adapt KEVM to this scheme: https://github.com/runtimeverification/evm-semantics/pull/1546

Some alternatives (perhaps better?)

- Instead of allowing overriding the entire `KDefinition`, allow specifying `extra_modules` that should be included into the one read on disk. This should be compatible with all the uses in KEVM.
- Instead of allowing to pass in `patch_symbol_table` as an argument, make it a method with default implementation the identity. Then subclases can override it as needed. This may be cleaner code-wise.


tothtamas28:
> Instead of allowing to pass in `patch_symbol_table` as an argument, make it a method with default implementation the identity. Then subclases can override it as needed.

I think that's the idiomatic solution here. So something like

```python3
class KPrint
    @cached_property
    def symbol_table(self) -> SymbolTable:
        symbol_table = build_symbol_table(self.definition, opinionated=True)
        self._patch_symbol_table(symbol_table)
        return symbol_table

    @classmethod
    def _patch_symbol_table(cls, symbol_table: SymbolTable) -> None:
        pass
```

> Instead of allowing overriding the entire `KDefinition`, allow specifying `extra_modules` that should be included into the one read on disk. This should be compatible with all the uses in KEVM.

That sounds good too. If we need a more flexible approach, we can add a method `_patch_definition` and deal with it in a similar way as we do with `symbol_table`.

ehildenb:
I've tested this with KEVM, and it seems to behave as expected (in the linked pyk update: https://github.com/runtimeverification/evm-semantics/pull/1546).

src/pyk/ktool/kprint.py
```diff
@@ -152,6 +153,7 @@ class KPrint:
     main_module: str
     backend: str
     _profile: bool
+    _extra_unparsing_modules: Optional[KFlatModuleList]
```
tothtamas28, line 156: Why not simply

```python
_extra_unparsing_modules: Iterable[KFlatModule] = ()
```

tothtamas28 (approved):

src/pyk/ktool/kprint.py
ehildenb: Hmmmmm, probably is better. I'll make the switch. We don't need the name attached to the FlatModuleList anyway.

PR #194, opened 2023-02-09
Fix `_subsort_table` in `KompiledKore`

tothtamas28:


src/pyk/kore/kompiled.py
```diff
@@ -42,11 +42,21 @@ def _subsort_table(self) -> FrozenDict[Sort, FrozenSet[Sort]]:
...
-        return FrozenDict({supersort: frozenset(subsorts) for supersort, subsorts in subsort_table.items()})
+        supersorts = direct_subsorts.keys()
+        sorts = set(direct_subsorts).union(*direct_subsorts.values())
+
+        subsort_table = dict(direct_subsorts)
+        for sort_k in supersorts:
+            for sort_i in sorts:
+                for sort_j in supersorts:
+                    if sort_i in subsort_table[sort_k] and sort_k in subsort_table[sort_j]:
+                        subsort_table[sort_j].add(sort_i)
```
ehildenb, line 57: This is doing, effectively, a pairwise search between all potential subsort relations? That seems slow..... But I suppose it works. It seems that computing it with a separate recursive function and a memo table is easier to read perhaps? https://github.com/runtimeverification/pyk/blob/master/src/pyk/kast/outer.py#L1047. Not sure.

If anything, I would re-order the loops a bit:

```
for sort_k in supersorts:
    for sort_j in supersorts:
        if sort_k != sort_j:
            for sort_i in sorts:
                ....
```

Seems a little more readable to me perhaps? Not sure honestly.

ehildenb: This way of doing it though, it's very expensive in terms of how many sorts are in the definition. Definitely think the recursive function + memo table is cheaper..... This one will scale at least cubically with the number of sorts, and on top of that quadratically with the number of direct subsorts. We should be able to achieve something that is closer to linear in the number of direct subsorts I believe.

ehildenb (approved):

src/pyk/kore/kompiled.py
tothtamas28: > This is doing, effectively, a pairwise search between all potential subsort relations?

This implements Warshall's algorithm (a special case of the Floyd–Warshall algorithm). The outermost loop maintains invariant that after the `k`-th iteration, all paths have been considered where nodes along the path - except perhaps for the start node `n_i` and end node `n_j` -  are in `{n_1, n_2, ..., n_k}`.

> This way of doing it though, it's very expensive in terms of how many sorts are in the definition.

Yeah, good point. I optimized the loops a bit, that got cumulative time spent in that function from 2.66s down to 0.02s on the example that was failing before.

I also tried a recursive variant, execution time there was around 0.07, so it's in the same order of magnitude as the optimized looping variant, but a bit slower, so I'd go with the later for now.

Please check if you find the changes sound. For the set update in the innermost loop to not cause problems, you need `sort_k != sort_j`, which I think holds by invariant `x not in subsort_table[x]`.

ehildenb (approved):

PR #197, opened 2023-02-10
Drop `Attr`

tothtamas28:
Related: https://github.com/runtimeverification/k/issues/3137


radumereuta:
Can you help me understand this PR?
Why do you need to remove `Att` completely?

tothtamas28:
> Can you help me understand this PR? Why do you need to remove `Att` completely?

In accordance with: https://github.com/runtimeverification/k/issues/3137#issuecomment-1424486644

Attributes are to be represented by `App` instances, `Att`, which is more restrictive, will no longer be used.

radumereuta:
In the front-end we have this [helper class](https://github.com/runtimeverification/k/blob/master/kore/src/main/scala/org/kframework/attributes/Att.scala) 
to store attributes and help retrieve them faster. It's very convenient since it acts like a Map. You can add elements, remove them, query...
So I don't understand why you want to remove the class completely. I would have expected you to add more parameters to support full ASTs, but not remove it altogether.

tothtamas28:
> It's very convenient since it acts like a Map.

Right now we don't need these helpers, but will definitely add such functions once the need arises.

> I would have expected you to add more parameters to support full ASTs,

Wouldn't that be just a reproduction of `App`?

> but not remove it altogether

Just to avoid confusion, this PR does not remove KORE attributes from `pyk`.
* https://github.com/runtimeverification/pyk/pull/186  -- this PR adjusted the AST and the parser to use `App` instead of `Attr` to represent attributes.
* Since the decision has been made to represent attributes as `App`, there is no reason to keep `Attr` around. This PR just removes dead code.

radumereuta (approved):
I get it now. Thanks.
Approved.

PR #200, opened 2023-02-11
Correct Kore encoding of Cell maps

ehildenb:
A bug discovered by KAVM getting stuck: https://github.com/runtimeverification/avm-semantics/pull/284

- Adds test harness for all-path-reachability prover for cell-map definition.
  - Adds failing test which mimics the failure in the above linked PR.
- Adds test harness for konvert for cell-map definition (and move konvert tests into subdirectory).
  - Adds failing test which minimizes the failure in the all-path-reachability proof to a failure in konvert.
- Adds methods to `KDefinition` for adding/removing `*CellMapItem` wrappers:
  - `KDefinition.cell_collection_productions` is a cached property which returns all the productions with `cellCollection` attribute.
  - `KDefinition.add_cell_map_items` adds in the wrappers `*CellMapItem` around cell-map collection items.
  - `KDefinition.remove_cell_map_items` remove those wrappers from cell-map collection items.
- `konvert.kast_to_kore` adds in cell-map-item collection wrappers.
- `konvert.kore_to_kast` strips the cell-map-item collection wrappers.
- Modifies `kast_to_kore` to not completely suppress `ValueError`, rather turn them into warnings.


src/tests/integration/konvert/test_cell_map.py
```diff
@@ -0,0 +1,169 @@
...
+            [
+                KApply(
+                    '_AccountCellMap_',
+                    [
+                        KApply('<account>', [KApply('<id>', [intToken(3)]), KApply('<balance>', [intToken(0)])]),
+                        KVariable('_DotVar2', sort=KSort('AccountCellMap')),
+                    ],
+                )
+            ],
+        ),
```
ehildenb, line 77: As an aside, this test alone should be convincing evidence that keeping Kast format around is valuable.

ehildenb:
I've confirmed that this gets rid of the spurious early branching on the KAVM semantics that we were observing. Now we see some other failure happening.

ehildenb:
Another solution would be for the Java frontend to always insert the `*CellMapItem` wrappers at the kast level, because as far as I can tell neither `compiled.json` or the output JSON specifications from `kprove --emit-json-spec ...` have them.

ehildenb:
Actually the proof is passing now: 
image


I had made some modifications to the proof for debugging which caused it to fail.

I also changed it to add `execute_depth=1` to the call to `all_path_reachability_prove`. This ensures we don't execute off the end of the proof (and we check implication every step), but it makes it very slow, because we're storing every state.

Better would be to make sure that every transaction inserts a dummy `#endTx()` marker, with a simple rule like this:

```
rule [endtx]: <k> endTx() => . ... </k>
```

Then we can supply that as a `terminal_rule` argument to `all_path_reachability_prove`, and it will make sure to save final states that we need to compare against the target node.

src/tests/integration/konvert/test_cell_map.py
```diff
@@ -0,0 +1,169 @@
...
+    (
+        'domain-value-int',
+        INT,
+        int_dv(3),
+        intToken(3),
+    ),
+    (
+        'account-cell-map',
+        KSort('AccountsCell'),
+        App(
```
tothtamas28, line 26: I think we should modify these tests to take textual KORE and do the parsing themselves. Consider creating an issue for this.

tothtamas28 (approved):

src/tests/integration/konvert/test_cell_map.py
ehildenb: Done: https://github.com/runtimeverification/pyk/issues/203

PR #210, opened 2023-02-17
Implement `kbuild which TARGET`

Baltoli:
This PR implements a command `kbuild which`, that prints the definition directory for a `kbuild` target.

More than happy to change the UI of the command to make it more consistent / idiomatic; currently, it prints an error and exits with error code 1 if the build directory doesn't exist.


src/pyk/kbuild/__main__.py
```diff
@@ -36,20 +40,41 @@ def _argument_parser() -> ArgumentParser:
     kompile_parser = command_parser.add_parser('kompile', help='kompile target')
     kompile_parser.add_argument('target_name', metavar='TARGET', help='target to build')
 
+    which_parser = command_parser.add_parser('which', help='print definition directory for target')
+    which_parser.add_argument('target_name', metavar='TARGET', help='target to print definition for')
```
tothtamas28, line 44:
```suggestion
    which_parser.add_argument('target_name', metavar='TARGET', help='target to print definition directory for')
```

tothtamas28 (approved):

PR #215, opened 2023-02-21
 Ignore attribute `Production` in `KVariable`

tothtamas28:


src/pyk/kast/inner.py
```diff
@@ -244,26 +244,33 @@ def __init__(self, name: str, *, sort: Optional[KSort] = None):
...
         att = KAtt.from_dict(d['att']) if d.get('att') else EMPTY_ATT
-        for a in [
+
+        sort: Optional[KSort]
+        if KAtt.SORT in att:
+            sort = KSort.from_dict(att[KAtt.SORT])
+        else:
+            sort = None
+
+        ignored_atts = {
```
tothtamas28, line 255: @ehildenb shouldn't we just drop `KAtt.SORT` and leave the rest?

ehildenb: The goal for now is to track which attributes are problems here: https://github.com/runtimeverification/k/issues/3120

ehildenb: so having the explicit list here (with an assertion failure) makes it easier to discover more problem attributes later.

I've added `Production` and `Sort` to that list.

ehildenb (approved):

PR #216, opened 2023-02-21
kcfg/explore: actually check result of implication

ehildenb:
As discovered by @ana-pantilie here: https://github.com/runtimeverification/haskell-backend/issues/3469

We actually are currently ignoring the `satisfiable` field of the `implies` endpoint. So this PR:

- Actually checks that field, returning `None` immediately if it's `false`.
- Removes the extra call to the simplifier at the end to check the instantiated consequent (the test `consequent-constraint` still passes).
- Adds assertions that if the result is `satisfiable`, there is a `predicate` and `substitution` present.
- Fixes two tests of `test_implies` that had the antecedent/consequent backwards, but were passing because we were ignoring the `satisfiable` field of the `implies` endpoint.


src/pyk/kcfg/explore.py
```diff
@@ -147,8 +147,12 @@ def cterm_implies(
         consequent_kore = self.kprint.kast_to_kore(_consequent, GENERATED_TOP_CELL)
         _, kore_client = self._kore_rpc
         result = kore_client.implies(antecedent_kore, consequent_kore)
-        if result.substitution is None:
+        if not result.satisfiable:
             return None
+        if result.substitution is None:
+            raise ValueError('Received empty substutition for satisfiable implication.')
+        if result.predicate is None:
+            raise ValueError('Received empty predicate for satisfiable implication.')
```
tothtamas28, line 155: Is this part of the `kore-rpc` contract? If so, we can enforce it in the client, even on the type level.

tothtamas28 (approved):

src/pyk/kcfg/explore.py
ehildenb: I'm really not sure, figured I raise a value error and see if it's every triggered.

ehildenb: Issue created: https://github.com/runtimeverification/pyk/issues/218

PR #219, opened 2023-02-22
Allow creating dir_path at CLI read time

ehildenb:


src/pyk/cli_utils.py
```diff
@@ -25,17 +25,20 @@ def loglevel(level: str) -> int:
...
     if not path.is_dir():
         raise ValueError(f'Path is not a directory: {path}')
 
 
-def dir_path(s: str) -> Path:
+def dir_path(s: str, create: bool = False) -> Path:
     path = Path(s)
-    check_dir_path(path)
+    check_dir_path(path, create=create)
     return path
```
tothtamas28, line 42: In my opinion this should be its own function:

```python3
def ensure_dir(path: Union[str, Path]) -> Path:
    path = Path(path)

    if not path.exists():
        return path.mkdir(parents=True)

    if path.is_dir():
        return path

    raise ValueError(f'Path exists, but is not a directory: {path}')

   
```

tothtamas28: It can then be also used as `parser.add_argument(..., type=ensure_dir, ...)`

ehildenb: I've done this, and tested `ensure_dir_path` with KEVM.

tothtamas28 (approved):

PR #222, opened 2023-02-23
Remove CFGManager

ehildenb:
This is only used downstream in one project, and we're trying to switch over to using KCFGExplore class in most cases.

This is left over from: https://github.com/runtimeverification/pyk/pull/193


nishantjr:
How about we split the file management from modifying the CFG?
That is, keep the CFG reading/writing in CFGManager
and have the various functions that manipulate it in KCFGExplore.

tothtamas28 (approved):

PR #225, opened 2023-02-24
Only apply #Ceil to config, not to constraints

ana-pantilie:


ehildenb:
@ana-pantilie and @jberthold what problem does this solve? It would be good to add a minimal example to the test-cases present here: https://github.com/runtimeverification/pyk/tree/master/src/tests/integration/kcfg

In that test harness, you have 3 definitions you can work with, `k-files/simple-proofs.k`, `k-files/imp-verification.k`, and `k-files/cell-map.k`. You can add to `EXECUTE_TEST_DATA` in each one, respectively, to make the unit test (for example, here: https://github.com/runtimeverification/pyk/blob/master/src/tests/integration/kcfg/test_cell_map.py#L22).

If you can make a minimal K definition demonstrating the issue, I can turn it into a test too.

jberthold:
> what problem does this solve? It would be good to add a minimal example to the test-cases present here

Adding a `#Ceil(all_of_cterm)` is a problem in the booster where we properly distinguish between terms and predicates (constraints): For instance, `#Ceil(#Equals(.., ..))` will be considered an error, as `#Equals` is already either `#Top` or `#Bottom`. The old backend accepts this as it does not make the distinction between term and predicate. I could not come up with a test case that would actually fail because of that.

I have added a `claim [addition-var]: <k> 3 + N => 7 </k> requires N ==Int 4` which shows an RPC request that would fail in the booster backend, see the diff here: (request with PR change left, request without right having `#Ceil(#And(.., #Equals(.., ..)))`
image
This `#Ceil` obviously does not make it into the response so it is hard to make a regression test for this detail.

ehildenb (approved):

PR #226, opened 2023-02-24
Fix minor bugs in unparsing attributes and add tests

ehildenb:
I tried unparsing a definition read in from `compiled.json`, and noticed taht pyk actually crashed on this. It's because the `Location` attribute is stored as:

```
...,"org.kframework.attributes.Location":[2135,3,2135,20],...
```

And our unparser was assuming that the values in attributes were strings, not more complex values. So this PR:

- Makes it so we can unparse anything that has string represenattion as values in attributes.
- Adds a test that a definition read in from the frontend is unparseable.
- Adds tests for unparsing certain special frontend attributes correctly.


ehildenb:
See discussion: https://runtimeverification.slack.com/archives/C7E701MFG/p1677265793462139

tothtamas28 (approved):

src/pyk/kast/kast.py
```diff
@@ -141,6 +141,20 @@ def update(self, atts: Mapping[str, Any]) -> 'KAtt':
     def remove(self, atts: Iterable[str]) -> 'KAtt':
         return KAtt({k: v for k, v in self.atts.items() if k not in atts})
 
+    def pretty(self) -> str:
```
tothtamas28, line 144: This can be a property.

ehildenb: Done.

PR #239, opened 2023-03-03
Add coarse-grained locking to `KBuild.kompile`

### LINKED ISSUE #237: Race condition in `KBuild.kompile`
The `KBuild.kompile(package: Package, target_name: str)` method encapsulates the call to `kompile` for a specific target of a package:
* it syncs the files of subpackages into `kbuild_dir`
* it checks that the files are up to date
* it calls `kompile`

There is a race condition if the `kbuild_dir` directory is shared among several processes that call `KBuild.kompile` concurrently, which leads to invoking `kompile` (the frontend executable) on a partially-initialized directory.

tothtamas28:
Fixes #237


tothtamas28:
@goodlyrottenapple, build fails due to a Nix issue I can't find the workaround for. Can you take a look?

goodlyrottenapple:
@tothtamas28 The `filelock` package was missing `hatch-vcs` dependency. I've turned on more vebose logging for nix via `nix build --print-build-logs ...` which should make diagnosing nix issues easier.

tothtamas28:
> I've turned on more verbose logging

Good idea, thank you, @goodlyrottenapple.

geo2a (approved):

PR #251, opened 2023-03-07
Add method `add_module` to `KoreClient`

tothtamas28:


src/pyk/kore/rpc.py
```diff
@@ -421,6 +421,10 @@ def simplify(self, pattern: Pattern) -> Pattern:
         result = self._request('simplify', **params)
         return kore_term(result['state'], Pattern)  # type: ignore # https://github.com/python/mypy/issues/4717
 
+    def add_module(self, module: Module) -> None:
+        result = self._request('add-module', name=module.name, module=Definition((module,)).text)
+        assert result == []
+
```
radumereuta, line 427: What is the purpose of adding a module in the rpc?

radumereuta: I guess I'm asking for more comments. This is not that obvious to me.

tothtamas28: `kore-rpc` exposes this endpoint so that clients can add lemmas on-the-fly.

radumereuta (approved):
I would still like it if it had more comments.
I can have a look at the code, and say what it does, but I can't say what the intent of that code is, or how it should be used.

`add_module` is the perfect example of this disconnect. Having this kind of documentation saves time in the long run.

Issue #248, opened 2023-02-28
Unsatisfiable implication is returning substitution and predicate


ehildenb:
It used to be that pyk was ignoring the value of `satisfiable` in the implies endpoint, but then it was fixed: https://github.com/runtimeverification/pyk/pull/216

But this is now causing problems in KEVM, because the implications that were considered successful before (because tehy had a substitution and predicate), are now being reported as unsuccessful.

Here is one such bug report: https://drive.google.com/file/d/16MrofCqXuyakhVwIq3aIciwuPv3CmLG4/view?usp=share_link

I'm pretty sure this implication _does_ hold, so my guess is that it should just be returning `satisfiable: "true"` instead of false. Either that, or I need to understand why this implication doesn't hold.

In general, if `satisfiable` is false, then we probably should not return _both_ a substitution and a predicate. Perhaps only a substitution, and `#Bottom` as the predicate makes sense though.

ehildenb:
Oh thank you guys for the investigation! I will try and make sure that predicates are desugared fully, and see if it's the same problem!

PR #261, opened 2023-03-11
Turn implication check validity back on

ehildenb:
Fixes: https://github.com/runtimeverification/pyk/issues/248

KEVM had an issue where it was sending macro symbols to the SMT checker, which had undesirable effects. That has been fixed here: https://github.com/runtimeverification/evm-semantics/pull/1627

This PR:

- Re-enables the check that if we get an unsatisfiable implication, then the substitution/predicate are empty.
- Adds another check that if we get a satisfiable implication, that the substitution/predicate are not empty, emits debug message if not the case. This can be useful for cases where the backend is able to tell that implication is not holding, but you want to get more information about why as the user (because it returns the unificand if it doesn't).
- Downgrades an info message to a debug message that was not very useful.
- Disables the (now useless/redundant) check that the instantiated consequent is `#Bottom`, which should make implication checker somewhat faster.


src/pyk/kcfg/explore.py
```diff
@@ -140,15 +140,23 @@ def cterm_implies(
...
+        if not result.satisfiable:
+            if result.substitution is not None:
+                _LOGGER.debug(f'Received a non-empty substitution for unsatisfiable implication: {result.substitution}')
+            if result.predicate is not None:
+                _LOGGER.debug(f'Received a non-empty predicate for unsatisfiable implication: {result.predicate}')
             return None
+        if result.substitution is None:
+            raise ValueError('Received empty substutition for satisfiable implication.')
+        if result.predicate is None:
+            raise ValueError('Received empty predicate for satisfiable implication.')
```
tothtamas28, line 159: Is this part of the contract then?

src/pyk/kcfg/explore.py
```diff
@@ -140,15 +140,23 @@ def cterm_implies(
...
         antecedent_kore = self.kprint.kast_to_kore(antecedent.kast, GENERATED_TOP_CELL)
         consequent_kore = self.kprint.kast_to_kore(_consequent, GENERATED_TOP_CELL)
         _, kore_client = self._kore_rpc
         result = kore_client.implies(antecedent_kore, consequent_kore)
-        if result.substitution is None:
+        if not result.satisfiable:
+            if result.substitution is not None:
+                _LOGGER.debug(f'Received a non-empty substitution for unsatisfiable implication: {result.substitution}')
+            if result.predicate is not None:
+                _LOGGER.debug(f'Received a non-empty predicate for unsatisfiable implication: {result.predicate}')
```
tothtamas28, line 154: Why are these debug log entries and not raised exceptions? Is this something that can happen but for some reason undesirable?

tothtamas28 (approved):

src/pyk/kcfg/explore.py
ehildenb: I'm not sure, but I'm enforcing it now, so that we can see if it holds.

src/pyk/kcfg/explore.py
ehildenb: Yeah, when you have `satisfiable: "false"`, the `substitution` can still be useful to determine which components of the configuration did not match (because those substitutions should all be _matching_ substitutions to do implication checking, but any _non-matching_ conjuncts indicate a place where amtching failed).

PR #273, opened 2023-03-17
Improve ml_pred_to_bool for #Equals(KSequence)

virgil-serbanuta:


virgil-serbanuta:
@tothtamas28 Note that this does not handle all KSequence cases, just the (by far) most common one.

tothtamas28 (approved):

PR #280, opened 2023-03-20
Enable proving claims with circularities via RPC

ehildenb:
Currently we have the `KCFGExplore` class which exposes the `all_path_reachabliity_prove` tactic, but this doesn't really fit in that class for a few reasons. For example, when we go to add circularities, somehow the proof state is no longer just the current KCFG, but also the circularities that are in context (which need to be reloaded the next time the proof is resumed, for example).

This PR:

- Adds a new module `kcfg/reachability.py` with a new class `AllPathReachabilityProof`.
- The `AllPathReachabilityProof` holds:
  - The `KCFG` datastructure of the current proof obligation.
  - The list of circularities in context.
- The `AllPathReachabilityProof` also exposes the method tha used to be `KCFGExplore.all_path_reachability_prove`, which is now `advance_proof`. It:
  - Has all the same arguments as before.
  - Takes into account the circularities. If the proof node that we are currently executing has non-zero depth from the initial node, then it makes sure to call the `execute` endpoint from the module which has the circularities loaded. If we are at zero depth from the frontier, we call it in the orginal definition instead.
- Adds a looping proof, where we require that the user annotate the circularity they want directly.


ehildenb:
`add-module` is here: https://github.com/runtimeverification/pyk/blob/c7ea8fabb7132b2703811c8a5465e4d89e9bb7cb/src/pyk/kore/rpc.py#L422

But has not been exposed in `KCFGExplore` class, so will need to be wired up there.

h0nzZik:
@ehildenb Which part of the codebase should be responsible for adding circularities-as-rules to the kore-rpc? I was thinking that `AGProver` should do that when it is initialized, but at that time it does not have access to `KCFGExplore` yet. I therefore suggest to have an instance of `KCFGExplore` as a member of `AGProver`, instead of as a parameter of `AGProver.advance_proof`.

h0nzZik:
Also, @ehildenb what do you think is the best way to detect
>  If the proof node that we are currently executing has non-zero depth from the initial 
? Should I use `KCFG.paths_between`?

PR #282, opened 2023-03-20
Add ability to to hide cells and node hashes in `KCFGShow.show()`

nwatson22:
This PR adds some optional parameters to `KCFGShow.show()` which allow the user to hide the node hashes for the nodes in `nodes`, as well as to hide specified cells, which applies to both the printed nodes, node deltas, and the generated module.

This is intended to be used in K foundry integration to have the option to omit certain information from the `foundry-show` command output that is likely to change for a test case without the behavior of that test case changing. This would be used for certain tests to verify the KCFG of that test does not change modulo node hashes and certain  "volatile" cells.


src/pyk/__main__.py
```diff
@@ -123,7 +123,8 @@ def kore_to_json() -> None:
 def json_to_kore() -> None:
     text = sys.stdin.read()
     kore = Pattern.from_json(text)
-    print(kore.text)
+    kore.write(sys.stdout)
+    sys.stdout.write('\n')
```
tothtamas28, line 127: The diff contains unrelated changes, please clean up commit history.

tothtamas28 (changes requested):
The diff contains some unrelated changes, please clean up commit history.

src/pyk/__main__.py
nwatson22: I'm not sure how that happened. It should be reverted now.

tothtamas28 (approved):

tothtamas28:
Created an issue related to this part of the code base: https://github.com/runtimeverification/pyk/issues/291

PR #290, opened 2023-03-22
Add parser method `multi_or`

tothtamas28:
Closes https://github.com/runtimeverification/pyk/issues/287


nishantjr (approved):
Thanks again Tamas. This works great.

Issue #259, opened 2023-03-10
Crash in `all_path_reachability_prove` on some configurations with abstracted cells


nwatson22:
## Steps to reproduce:

1. Create files:

`kcfg/test.py`

```py
from pyk.kast.inner import KApply
from pyk.kcfg.tui import KCFGViewer
from pyk.kcfg.explore import KCFGExplore
from pyk.kcfg.kcfg import KCFG
from pyk.ktool.kprove import KProve
from pathlib import Path

kprove = KProve(definition_dir="verification")

claim_id = "main"
spec_file = Path("simple-spec.k")
spec_module = spec_file.name.removesuffix('.k').removesuffix('.md').upper()
claim_label = f'{spec_module}.{claim_id}'
claims = kprove.get_claims(spec_file, spec_module_name=spec_module, claim_labels=[claim_label])

cfg = KCFG.from_claim(kprove.definition, claims[0])
cfg_id = f'{claim_label}.kfcg'

with KCFGExplore(kprove, 7777) as kcfg_explore:
    kcfg_explore.all_path_reachability_prove(
        cfg_id,
        cfg,
        cfg_dir=Path("output"),
        terminal_rules=['AVM-EXECUTION.starttx', 'AVM-EXECUTION.endtx', 'AVM-PANIC.panic'],
    )


kcfg = KCFGExplore.read_cfg(cfg_id, Path("output"))


app = KCFGViewer(kcfg=kcfg, kprint=kprove, minimize=False)
app.run()
```

`pyproject.toml`

```toml
[tool.poetry]
name = "kcfg"
version = "0.1.0"
description = ""
authors = []

[tool.poetry.dependencies]
python = "^3.10"
pyk = { git = "https://github.com/runtimeverification/pyk.git", tag="v0.1.173" }

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"

```

`verification.k`

```k
module VERIFICATION
  imports INT

  configuration 
  <k> $PGM:K </k>
  <outerCell>
    <cell1> 0 </cell1>
    <cell2> 0 </cell2>
  </outerCell>
endmodule
```

`simple-spec.k`

```k
module SIMPLE-SPEC
  imports VERIFICATION

  claim [main]: 
  <k> _ => ?_ </k>
  <outerCell>
     _ => ?_
  </outerCell>

endmodule
```

2. install pyk
`poetry -C kcfg install`

3. create kcfg output dir
`mkdir output`

4. kompile
[...]

PR #293, opened 2023-03-23
Rewrite rename_generated_vars to be more general / avoid split_config_from

Scott-Guest:
Fixes: https://github.com/runtimeverification/pyk/issues/259

`rename_generated_vars` currently uses `split_config_from` to extract leaf cells and look for generated variables within them. This can crash, e.g., when variables are not contained in a cell or are contained in a cell with arity > 1 (see #259).

This PR resolves these crashes by rewriting `rename_generated_vars` to just directly traverse the AST.


ehildenb (approved):

ehildenb:
Very nice test-suite, thank you! I'll let @tothtamas28 review and put `automerge` if it looks good to him as well.

src/pyk/kast/manip.py
```diff
@@ -663,20 +665,34 @@ def undo_aliases(definition: KDefinition, kast: KInner) -> KInner:
...
+            cell_stack.pop()
+            return res
+
+        if type(k) is KVariable and _is_gen_var_name(k.name):
+            if not cell_stack:
+                return k
+            cell_name = cell_stack[-1]
+            new_var = abstract_term_safely(k, base_name=cell_name)
+            while new_var.name in vars:
+                new_var = abstract_term_safely(KVariable(new_var.name), base_name=cell_name)
```
tothtamas28, line 692: Under what condition will this loop execute more than once?

tothtamas28 (commented):
Looks good!

tothtamas28 (approved):

src/pyk/kast/manip.py
ehildenb: I would consider moving the logic of the loop directly into the `abstract_term_safely` method, so that you can pass in the set of variables it is not allowed to return directly, and it will handle making sure to give back a variable name that is not the same as any in that set.

If `new_var.name` and `cell_name` are the same, it would execute more than once. I doubt that situation can happen, if the frontend doesn't do anything silly, but being cautious about it seems like a good idea.

src/pyk/kast/manip.py
```diff
@@ -663,20 +665,34 @@ def undo_aliases(definition: KDefinition, kast: KInner) -> KInner:
...
-            var_subst[v] = new_v
-            vs.append(new_v.name)
-    return Subst(var_subst).apply(term)
+    def _is_gen_var_name(name: str) -> bool:
+        return (
+            name.startswith('_Gen')
+            or name.startswith('?_Gen')
+            or name.startswith('_DotVar')
+            or name.startswith('?_DotVar')
+        )
```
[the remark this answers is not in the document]
Scott-Guest: Oh nice, wasn't aware of this. There's a bit of a discussion here on why it only takes a `Tuple`: [https://stackoverflow.com/questions/45117924/how-does-str-startswith-really-work](https://stackoverflow.com/questions/45117924/how-does-str-startswith-really-work)

src/pyk/kast/manip.py
Scott-Guest: Done

PR #294, opened 2023-03-23
Introduce Constrained Substitution, CSubst, class

ehildenb:
This is somewhat a blocker on: https://github.com/runtimeverification/pyk/pull/124

Working with that PR, it became clear that the `Cover` and `Split` nodes have some duplicate data. So here we factor out that data/functionality into the `CSubst` class, which stores a substitution and a constraint.

- Class `CSubst` is added, with functionality:
  - `apply`: apply a `CSubst` to a `CTerm` to get a new `CTerm`.
  - `add_constraint`: return a new `CSubst` with the additional constraint.
- Method `CTerm.match_with_constraint` now returns an `Optional[CSubst]` instead.
- Adds methods `CTerm.{to_dict,from_dict}` and `CSubst.{to_dict,from_dict}`, and uses those directly in the various `KCFG.{to_dict,from_dict}` places.
- Removes a method `path_condition` from `kcfg.py` which had no uses here.
- Stores a `CSubst` directly as the data of a `KCFG.Cover` instead of storing a constraint/substitution directly.
- Adjust `KCFGExplore.cterm_implies` to return a `CSubst` instead of a tuple of `Subst, KInner`.
- Adjusts the tests to the new datastructures.
- Increases the parallelism of tests on CI.


ehildenb:
@Scott-Guest I've marked you as reviewer to give you a chance to look over this code, but your review is not necessary if you don't have time.

Scott-Guest (approved):
Everything looks good to me, seems like a clear improvement!

.github/workflows/test-pr.yml
```diff
@@ -77,7 +77,7 @@ jobs:
           sudo apt-get -y install graphviz ./kframework_amd64_ubuntu_jammy.deb
           kompile --version
       - name: 'Build and run integration tests'
-        run: make cov-integration
+        run: make cov-integration TEST_ARGS='-n8'
```
tothtamas28, line 80:
```suggestion
        run: make cov-integration COVER_ARGS='-n8'
```

Does this make execution faster, i.e. do `normal-ephemeral` instances have enough cores?

src/pyk/cterm.py
```diff
@@ -52,6 +52,18 @@ def _constraint_sort_key(term: KInner) -> Tuple[int, str]:
...
+        return {
+            'config': self.config.to_dict(),
+            'constraints': [c.to_dict() for c in self.constraints],
+        }
+
+    @staticmethod
+    def from_dict(dct: Dict[str, Any]) -> 'CTerm':
+        config = KInner.from_dict(dct['config'])
+        constraints = [KInner.from_dict(c) for c in dct['constraints']]
+        return CTerm(mlAnd([config] + constraints))
```
tothtamas28, line 65: Created an issue: https://github.com/runtimeverification/pyk/issues/295

Then this could look as:

```python3
config = KInner.from_dict(dct['config'])
constraints = (KInner.from_dict(c) for c in dct['constraints'])
return CTerm(config, constraints)
```

src/pyk/cterm.py
```diff
@@ -97,6 +107,45 @@ def add_constraint(self, new_constraint: KInner) -> 'CTerm':
         return CTerm(mlAnd([self.config, new_constraint] + list(self.constraints), GENERATED_TOP_CELL))
 
 
+@dataclass(frozen=True, order=True)
+class CSubst:
+    subst: Subst
+    constraints: Tuple[KInner, ...]
+
+    def __init__(self, subst: Optional[Subst] = None, constraint: Optional[KInner] = None) -> None:
```
tothtamas28, line 115: Consider this more direct approach:

```suggestion
    def __init__(self, subst: Optional[Subst] = None, constraints: Iterable[KInner] = ()) -> None:
```

Then methods like `add_constraint` would be more straightforward and efficient.

src/pyk/cterm.py
```diff
@@ -97,6 +107,45 @@ def add_constraint(self, new_constraint: KInner) -> 'CTerm':
         return CTerm(mlAnd([self.config, new_constraint] + list(self.constraints), GENERATED_TOP_CELL))
 
 
+@dataclass(frozen=True, order=True)
+class CSubst:
+    subst: Subst
+    constraints: Tuple[KInner, ...]
```
tothtamas28, line 113: This could be a `NamedTuple` instead, then for `__iter__`:

```python3
subst, constraints = csubst 
```

Here, `subst: Subst` and `constraints: Tuple[KInner, ...]`.

Created an issue for `CTerm` for symmetry: https://github.com/runtimeverification/pyk/issues/296

tothtamas28 (approved):

.github/workflows/test-pr.yml
ehildenb: Yes, they should have 12 cores each.

src/pyk/cterm.py
ehildenb: :+1:

src/pyk/cterm.py
ehildenb: I've added `CSubst` to the issue too, can be taken care of then.

src/pyk/cterm.py
ehildenb: Done.

ehildenb: Note that I needed to change `_is_spurious_constraint` to throw away `#Top` constraints too now.

PR #298, opened 2023-03-24
Kill `kore-rpc` with `SIGINT` instead of `SIGTERM`

tothtamas28:
Workaround for https://github.com/runtimeverification/haskell-backend/issues/3557


ehildenb (approved):

jberthold:
Thanks for changing this! Besides the issues with remaining temporary directories, we also  could not get profiling data from the back-end when it was killed forcefully. Using `SIGINT` is the recommended way to terminate a backend programs gracefully.

Issue #291, opened 2023-03-23
Refactor `KCFG` printer methods


tothtamas28:
Move methods to module `show`:

```
EdgeLike.pretty
pretty_segments
pretty
to_dot
```

Extract abstract class `KCFGShow`:

```python3
class KCFGShow(ABC):
    @abstractmethod
    show(self, cfg: KCFG) -> str:
        ...

    def __call__(self, cfg: KCFG) -> str:
        return self.show(cfg)
```

Make each printing function where it makes sense a subclass of `KCFGShow` (e.g. `KCFGDotShow`, `KCFGPrettyShow`, ...).

ehildenb:
The individual components of `pretty_segments` should be exposed, so that they can be unit-tested: https://github.com/runtimeverification/pyk/pull/302#discussion_r1148509812

ehildenb:
This is largely done, printing of KCFGs and Proofs has seen major overhauls.

PR #302, opened 2023-03-26
Refactor KCFG printing, remove verified edge attribute

ehildenb:
This PR is pulled out of: https://github.com/runtimeverification/pyk/pull/124
In that PR, the `pretty_segments` routine is adjusted to accomodate the new `KCFG.Split` edge. The refactoring of that routine can be done standalone, by detecting splits in the old representation (which is what this PR does):

This PR:

- Removes the `verified` attribute from `KCFG` datastructure, as it's not longer as meaningful since we're using RPC based prover.
- Refactors the `_print_subgraph` method of `pretty_segments` to be a bit more readable:
  - Factors out `_print_split_edge` for printing out edge information, which now is more informative.
  - Prints out the current node _first_ (instead of at end of function, then prints out successors).
  - Improves the formatting of the printed CFGs a bit.
  - Factors out node sorter to decide which components of the KCFG to print out first (focuses on printing out graphs first, then remaining nodes second).
- Adds tests of KCFG printing cases that were not exercised before, and broke during this refactoring (and were fixed), including:
  - Test of a `frontier` node being printed in graph.
  - Test of a `stuck` node being printed in graph.
  - Test of the user supplying the `node_printer` optional argument.


ehildenb:
Screenshot from 2023-03-25 20-47-33

Here is example output after this change for a KEVM proof.

src/pyk/kcfg/kcfg.py
```diff
@@ -371,77 +361,124 @@ def _print_node(node: KCFG.Node) -> List[str]:
...
 
+        def _print_split_edge(constraint: KInner) -> List[str]:
+            ret_split_lines: List[str] = []
+            constraints = flatten_label('#And', constraint)
+            if len(constraints) > 0:
+                if len(constraints) == 1:
+                    ret_split_lines.extend(f'constraint: {kprint.pretty_print(c)}' for c in constraints)
+                else:
+                    ret_split_lines.append('constraints:')
+                    ret_split_lines.extend(f'    {kprint.pretty_print(c)}' for c in constraints)
```
tothtamas28, line 372:
```suggestion
            if len(constraints) == 1:
                ret_split_lines.append(f'constraint: {kprint.pretty_print(constraints[0])}')
                    
            elif len(constraints) > 1:
                ret_split_lines.append('constraints:')
                ret_split_lines.extend(f'    {kprint.pretty_print(c)}' for c in constraints)
```

src/pyk/kcfg/kcfg.py
```diff
@@ -371,77 +361,124 @@ def _print_node(node: KCFG.Node) -> List[str]:
...
+            return ret_split_lines
+
         def _print_subgraph(indent: str, curr_node: KCFG.Node, prior_on_trace: List[KCFG.Node]) -> None:
-            edges_from = sorted(self.edge_likes(source_id=curr_node.id))
-            if curr_node in processed_nodes:
-                if not edges_from:
-                    return
-                ret_edge_lines = [(indent + '┊')]
+            processed = curr_node in processed_nodes
```
tothtamas28, line 376: Aside for what's printed for each node, is this a generic tree printer? If so, we could extract that logic into a utility and unit test it (as a separate issue of course).

tothtamas28 (approved):

src/pyk/kcfg/kcfg.py
ehildenb: I guess that can be part of: https://github.com/runtimeverification/pyk/issues/291

PR #303, opened 2023-03-26
Enforced KCFG sanity checks

ehildenb:
This PR pulls some of the work out of: https://github.com/runtimeverification/pyk/pull/124

Basically, we'll be restricting more what a "valid" KCFG is. This PR:

- Makes the various classes `KCFG.*` into something closer to dataclasses by removing their `init` methods.
- Instead, the sanity checks/initialization that `__init__` was doing is moved into `KCFG.create_*` methods. This allows us to make malformed edges/covers/nodes in isolation, but cannot build a KCFG with those data.
- Removes unused functionalities from `KCFG.Edge`.
- Enforces that every node in the KCFG has at most one successor. This is only enforced loosley, because we currently represent splits as multiple 0-length edges.


src/pyk/kcfg/kcfg.py
```diff
@@ -654,6 +620,18 @@ def replace_node(self, node_id: str, new_cterm: CTerm) -> str:
 
         return new_node.id
 
+    def _assert_no_successors(self, source_id: str) -> None:
```
tothtamas28, line 623: `_assert` suggest an `AssertionError` (and thus a bug), consider `_check` instead:

```suggestion
    def _check_no_successors(self, source_id: str) -> None:
```

tothtamas28 (approved):

src/pyk/kcfg/kcfg.py

PR #304, opened 2023-03-27
Fix kcfg show spacing when no nodes or node deltas are printed

nwatson22:
This is a small cosmetic change for a downstream formatting change requested [here](https://github.com/runtimeverification/evm-semantics/pull/1663#discussion_r1149578438). It changes the output of foundry show to put spacing in between the kcfg summary and the generated module, as there was no spacing by default when there are no printed nodes or node deltas.


src/pyk/kcfg/show.py
```diff
@@ -82,6 +82,11 @@ def _hide_cells(_k: KInner) -> KInner:
             res_lines.append(self.kprint.pretty_print(config_delta))
             res_lines.append('')
 
+        if any(nodes) == 0 and any(node_deltas) == 0:
```
ehildenb, line 85:
```suggestion
        if not (nodes or node_deltas):
```

ehildenb (approved):

src/pyk/kcfg/show.py
tothtamas28: Both `nodes` and `node_deltas` are `Iterable`-s, and at this point of execution potentially consumed. So I think the best solution would be to set a flag in the loops consuming the iterables, and check them here.

tothtamas28 (approved):

src/pyk/kcfg/show.py
nwatson22: I changed it to use a flag here as suggested.

Issue #310, opened 2023-03-29
Updates to pyk prover architecture


ehildenb:
Blocked on: https://github.com/runtimeverification/pyk/pull/280

Most practical provers that people are using (especially in Ethereum), actually are _bounded_ provers. They unroll each loop a bounded number of times, and then "pass" the proof if it's satisfied in that unroll-bound. By default, K does not allow this, you _must_ do full verification using loop invariants.

But it can be a much better first-touch user experience to offer bounded unrolling of loops instead, because then users don't need to learn as much about loop invariants and such, instead can see if any counterexamples are found in the bounded unrolling.

We'll implement the bounded variant of all path reachability prover, in a few steps.

- [x] Factor out `pyk.prover` module in pyk, and add `pyk/prover/reachability.py` as one of the submodules. https://github.com/runtimeverification/pyk/pull/311
- [x] Class `ProofStatus` in `pyk/prover/prover.py` which contains an enum of: (i) `PASSED`, (ii) `PENDING`, and (iii) `FAILED`.
- [x] Superclass `Prover` which contains abstract methods:
  - `status -> ProofStatus`: status of this proof.
  - `from_dict: Dict -> Prover`: deserializes proof state from stored format.
  - `to_dict: Dict -> Prover`: serializes proof state from stored format.
- [x] Class `ReachabilityProver(Prover)` which contains in addition:
  - `kcfg`: the current proof state.
  - `KCFGExplore`: access to the RPC server.
  - `circularities`: circularities in context for the prover.
  - When serializing/deserializing from disk, this prover should also read/write the circularities.
- [x] Class `AllPathReachabilityProver(Prover)` which contains the method:
  - `prove`: copied from what is current `KCFGExplore.all_path_reachability_prove`.
  - `status` is implemented as: (i) "Passed" if no frontier or stuck nodes, (ii) "Failed" if there are stuck nodes, (iii) "Pending" if there are no stuck nodes, but are frontier nodes.
- [x] Class `BoundedAllPathReachabilityProver(ReachabilityProver)`, which is for bounded model checking.
  - Adds a new heuristic the user supplies to `prove` method, which is the `same_loop_head` heuristic.
  - Adds a new set of node ids called the `loop_capped` set.
  - On initialization, we set the loop unrolling depth.
  - When getting new nodes on the frontier, we first check them for being the `same_loop_head` as prior loops. If it is, then we cap execution there by:
    - Marking that node as stuck.
[...]

PR #311, opened 2023-03-29
Factor out Prover interfaces

h0nzZik:
Part of: https://github.com/runtimeverification/pyk/issues/310

This PR:

- Makes a new module `pyk.proof` which contains the various Proof states. This module defines:
  - The `ProofStatus`, which is an enum of how a proof is doing.
  - The generic `Proof` class, which requires that each `Proof` state be able to report its status and serialize itself to a JSON dictionary.
- Adds a module `pyk.proof.reachability`, which contains:
  - `AllPathReachabilityProof(Proof)`, which:
    - Has the method `advance_proof` which is what used to be `KCFGExplore.all_path_reachability_prove`.
    - Just stores a KCFG as extra state.


src/pyk/kcfg/show.py
```diff
@@ -89,6 +93,11 @@ def _hide_cells(_k: KInner) -> KInner:
             res_lines.append(self.kprint.pretty_print(config_delta))
             res_lines.append('')
 
+        if not (nodes_printed):
+            res_lines.append('')
+        res_lines.append('')
+        res_lines.append('')
+
```
ehildenb, line 100: Github is just being lame here, if you look at the diff locally this is not included. This is from a PR that was just merged: https://github.com/runtimeverification/pyk/pull/304

src/pyk/proof/proof.py
```diff
@@ -0,0 +1,44 @@
...
+_LOGGER: Final = logging.getLogger(__name__)
+
+
+class ProofStatus(Enum):
+    PASSED = 'passed'
+    FAILED = 'failed'
+    PENDING = 'pending'
+
+
+class Proof:
```
tothtamas28, line 19:
```suggestion
class Proof(ABC):
```

src/pyk/proof/proof.py
```diff
@@ -0,0 +1,44 @@
...
+        if actual != expected:
+            raise ValueError(f'Expected "type" value: {expected}, got: {actual}')
+
+    @classmethod
+    @abstractmethod
+    def from_dict(cls: Type[Proof], dct: Dict[str, Any]) -> Proof:
+        proof_type = dct['type']
+        if proof_type in Proof._PROOF_TYPES:
+            return globals()[proof_type].from_dict(dct)
+        raise ValueError(f'Expected "type" value in: {Proof._PROOF_TYPES}, got {proof_type}')
```
tothtamas28, line 35: If I understand correctly, `Proof` is an open class, so hardcoding all subclasses is not feasible. Fortunately, at this point, it does not seem to be necessary either.

Once we reach the level of abstraction where we have an abstract class that depends on an abstract `Proof`, we can come up some mechanism for registering the subclasses for parsing.

```suggestion
    @classmethod
    @abstractmethod
    def from_dict(cls: Type[Proof], dct: Dict[str, Any]) -> Proof:
        ...
```

src/pyk/proof/proof.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @property
+    @abstractmethod
+    def status(self) -> ProofStatus:
+        ...
+
+    @abstractmethod
+    def to_dict(self) -> Dict[str, Any]:
+        ...
```
tothtamas28, line 44:
```suggestion
    @property
    @abstractmethod
    def dict(self) -> Dict[str, Any]:
        ...

    @property
    @abstractmethod
    def status(self) -> ProofStatus:
        ...
```

src/pyk/proof/proof.py
tothtamas28, line 19: I understand that a proof 1) is serializable / deserializable 2) has a status. What are the clients of this interface?

src/pyk/proof/reachability.py
```diff
@@ -0,0 +1,126 @@
...
+
+    from ..cterm import CTerm
+    from ..kast.inner import KInner
+
+T = TypeVar('T', bound='Proof')
+
+_LOGGER: Final = logging.getLogger(__name__)
+
+
+class AllPathReachabilityProof(Proof):
```
tothtamas28, line 22: For this, I suggest composition instead of inheritance. That way, you can avoid:

```python3
prover = AllPathReachabilityProof(kcfg)
```

Also, this class name might too long for the frequency it will be typed out with.

So consider:

```python3
AGProof(Proof):
    kcfg: KCFG
    ...
```

```python3
AGProver:
    kcfg: KCFG

    @property
    def proof(self) -> Proof:   # can be AGProof as well
        return AGProof(self.kcfg)

    ...
```

If you want to take this one step further, you can

```python3
P = TypeVar('P', bound=Proof)

class Prover(Generic[P], ABC):
    @property
    @abstractmethod
    def proof(self) -> P:
        ...

class AGProver(Prover[AGProof]):
    ...
```

but this might be a bit premature at this level of abstraction.

tothtamas28: Sorry for the confusion. I mean this assignment implies prover is a proof.

src/pyk/proof/proof.py
ehildenb: We want to be able to store proof state, in general. So proofs need to be resumable, and directly representable as data. Having this interface enforces that.

src/pyk/proof/reachability.py
```diff
@@ -0,0 +1,143 @@
...
+        simplify_init: bool = True,
+        implication_every_block: bool = True,
+    ) -> KCFG:
+        def _write_proof() -> None:
+            if kproofs_dir:
+                proof_dict = self.proof.dict
+                proof_dict['proofid'] = proofid
+                proof_path = kproofs_dir / f'{hash_str(proofid)}.json'
+                proof_path.write_text(json.dumps(proof_dict))
+                _LOGGER.info(f'Updated AGProof file {proofid}: {proof_path}')
```
tothtamas28, line 75: Consider pulling this to a method.

src/pyk/proof/reachability.py
```diff
@@ -0,0 +1,143 @@
...
+        cfg = KCFG.from_dict(dct['cfg'])
+        return AGProof(cfg)
+
+    @property
+    def dict(self) -> Dict[str, Any]:
+        return {'type': 'AGProof', 'cfg': self.kcfg.to_dict()}
+
+
+class AGProver:
+    proof: AGProof
```
tothtamas28, line 50: In the end, why did you choose this over `kcfg: KCFG`?

src/pyk/proof/reachability.py
```diff
@@ -0,0 +1,143 @@
...
+        kcfg_explore: KCFGExplore,
+        kproofs_dir: Optional[Path] = None,
+        is_terminal: Optional[Callable[[CTerm], bool]] = None,
+        extract_branches: Optional[Callable[[CTerm], Iterable[KInner]]] = None,
+        max_iterations: Optional[int] = None,
+        execute_depth: Optional[int] = None,
+        cut_point_rules: Iterable[str] = (),
+        terminal_rules: Iterable[str] = (),
+        simplify_init: bool = True,
+        implication_every_block: bool = True,
```
tothtamas28, line 67: Note for an upcoming PR:

Some of these parameters could probably be converted into an argument or method of

1) this class, or
2) some other class, say, `AGProver.Strategy`, that you can pass as a (constructor or method) argument.

tothtamas28 (approved):

src/pyk/proof/reachability.py
h0nzZik: 1. Conceptually, a prover manipulates a proof, not a CFG.
2. If a user if the prover wants the proof, he can just grab it, without the need to create a temporary wrapper object.

PR #315, opened 2023-03-30
Refactor the `kllvm` module

tothtamas28:
~Blocked on: K release 5.6.23~

Enables importing `_kllvm` from a given location. This way, the module file can be reused between runs.


tothtamas28:
## A simple test

```
$ poetry run python3 test.py
```

`test.py`

```python3
import logging
from pathlib import Path

from pyk.ktool.kompile import kompile
from pyk.kllvm.compiler import KLLVM_MODULE_FILE_NAME, RUNTIME_MODULE_FILE_NAME, compile_kllvm, compile_runtime
from pyk.kllvm.importer import import_kllvm, import_runtime


logging.basicConfig(level=logging.INFO)


# -----
# Setup
# -----

definition_dir = Path('imp-kompiled')
if not definition_dir.exists():
    kompile('k-files/imp.k', output_dir=definition_dir)

if not (definition_dir / KLLVM_MODULE_FILE_NAME).exists():
    compile_kllvm(definition_dir, verbose=True)

import_kllvm(definition_dir)

from pyk.kllvm.parser import parse_file

if not (definition_dir / RUNTIME_MODULE_FILE_NAME).exists():
    compile_runtime(definition_dir, verbose=True)

runtime = import_runtime(definition_dir)


# ---------
# Interpret
# ---------

for i in range(10000):
    print(i)
    init_pattern = parse_file('init.kore')
    final_pattern = runtime.interpret(init_pattern)
    print(final_pattern)

```

`init.kore`

```
LblinitGeneratedTopCell{}(Lbl'Unds'Map'Unds'{}(Lbl'Stop'Map{}(),Lbl'UndsPipe'-'-GT-Unds'{}(inj{SortKConfigVar{}, SortKItem{}}(\dv{SortKConfigVar{}}("$PGM")),inj{SortPgm{}, SortKItem{}}(Lblint'UndsSClnUnds'{}(Lbl'UndsCommUnds'{}(\dv{SortId{}}("x"),Lbl'Stop'List'LBraQuotUndsCommUndsQuotRBraUnds'Ids{}()),Lbl'UndsEqlsUndsSCln'{}(\dv{SortId{}}("x"),inj{SortInt{}, SortAExp{}}(\dv{SortInt{}}("1"))))))))
```

tothtamas28:
@Baltoli, can you try the script above on your system? I get the following error:

```
...
1019
Lbl'-LT-'generatedTop'-GT-'{}(Lbl'-LT-'T'-GT-'{}(Lbl'-LT-'k'-GT-'{}(dotk{}()),Lbl'-LT-'state'-GT-'{}(Lbl'UndsPipe'-'-GT-Unds'{}(inj{SortId{}, SortKItem{}}(\dv{SortId{}}("x")),inj{SortInt{}, SortKItem{}}(\dv{SortInt{}}("1"))))),Lbl'-LT-'generatedCounter'-GT-'{}(\dv{SortInt{}}("0")))
1020
Cannot read file: tmp.parse.50h46B
```

Strangely, the script fails consistently in iteration 1020.


Also, why is the temp file created? Is `Parser(file_name)` a wrapper around an interface that parses from string?

Baltoli:
> can you try the script above on your system?

yes, will do

> Also, why is the temp file created? Is Parser(file_name) a wrapper around an interface that parses from string?

It's the other way around. The LLVM backend uses flex to generate a lexer, and that requires a file on disk, so to implement "parse a string" [we have to dump it to a file on disk](https://github.com/runtimeverification/llvm-backend/blob/d6386a2bcdafa92b1f100b93f52c94bf5e69bf2c/lib/parser/KOREParser.cpp#L13-L28).

src/pyk/kllvm/importer.py
```diff
@@ -0,0 +1,89 @@
...
+    runtime.interpret = _make_interpreter(term_cls)  # type: ignore
+
+
+def _make_interpreter(term_cls: Type) -> Callable[..., Pattern]:
+    from .parser import parse_text  # TODO eliminate
+
+    def interpret(pattern: Pattern, *, depth: Optional[int] = None) -> Pattern:
+        term = term_cls(pattern)
+        term.step(depth if depth is not None else -1)
+        return parse_text(str(term))
```
tothtamas28, line 65: > It's the other way around. The LLVM backend uses flex to generate a lexer, and that requires a file on disk, so to implement "parse a string" we have to dump it to a file on disk

@Baltoli the missing temporary file issue might actually come from here then. Would that make sense?

tothtamas28: The error comes from `Parser.from_string`.

tothtamas28:
@Baltoli the K update went through, tests are passing now.

Baltoli (approved):
Looks good!

PR #324, opened 2023-04-03
Improve performance of `dequoted`

tothtamas28:
Replaces calls `int(c, 16)` by `dict` lookups. Also improves error handling.


src/pyk/dequote.py
```diff
@@ -25,20 +25,26 @@
...
 
 def dequoted(it: Iterable[str]) -> Iterator[str]:
     acc = 0
     cnt = 0
     state = NORMAL
     for c in it:
         if state == CPOINT:
+            if c not in HEX_TABLE:
+                raise ValueError(f'Expected hex digit, got: {c}')
+
```
ehildenb, line 39: Does this check introduce some unnecessary overhead? It's an extra branch every iteration of the loop.....

Do we really lose that much in terms of error messaging if it's omitted? I guess it's not telling us that it explicitly expects a Hex instead of something else.

ehildenb (approved):

src/pyk/dequote.py
tothtamas28: It introduces some overhead, but it's not significant. I'd rather keep it until it becomes a bottleneck.

> I guess it's not telling us that it explicitly expects a Hex instead of something else.

This would raise a `KeyError` then. I also tried handling this with `try-except`, but it's only a bit faster but in my opinion less clean.

PR #325, opened 2023-04-03
Refactors and cleanups to new proof class

ehildenb:
This makes several refactors to the prover class, and should fix the issue we're having with KEVM: https://github.com/runtimeverification/evm-semantics/issues/1701 (or at least partially). It is paired with this KEVM PR: https://github.com/runtimeverification/evm-semantics/pull/1706

The new structure is not perfect, but it is somewhat an improvement:

- Methods `read_cfg` and `write_cfg` are removed from `KCFGExplore`, to make sure we aren't using them directly anymore.
- `Proof` class now carries with it the proof id, and teh optional `proof_dir`. Method `Proof.write_proof` will serialize the proof to a file on disk if the optional `proof_dir` is populated.
- Static method `read_proof`, and `proof_exists` is added to `AGProof`.
- A small bug in how multi-line constarints were printed in teh KCFG is fixed (screenshots below).


ehildenb:
And after fixing it:

Screenshot from 2023-04-07 00-02-13
Screenshot from 2023-04-07 00-05-35

anvacaru (approved):

Issue #300, opened 2023-03-24
Integrate `kore-exec-covr` into `KCFGExplore`


geo2a:
`KCFGExplore.cterm_execute` should be able to produce reports of rewrite and simplification rules for debugging, similar to the current `kore-exec-covr` tool.

PR #327, opened 2023-04-04
Add explicit Haskell logging options to `KCFGExplore` and `KoreServer`

### LINKED ISSUE #299: Integrate `kore-exec-covr` into `KProve`
The `kore-exec-covr` scraps K's Haskell backend's oneline logs for rewrite and simplification rule applications.

Currently, the tool is implemented as a standalone executable. We need to enable it's functionality in the `KProve.prove`/`KProve.prove_claim`/` KProve.prove_cterm`.

geo2a:
(Partially) Fixes #299 and #300 

This PR makes adds arguments to `KCFGExplore` and `KoreServer` initializers that explicitly pass Haskell backend logging options.

Upon finishing a proof, the Haskell backend log file can be analyzed with `kore-exec-covr`. I think it is preferable to keep this analysis standalone until we have a logging endpoint in `kore-rpc`.

@tothtamas28 I'm reluctant to modify the integration tests to also test the log file parsing functionality since it may be brittle due to the fact that the `domains.md` likely come from the Nix store. We already have some [unit tests](https://github.com/runtimeverification/pyk/blob/master/src/tests/unit/test_kore_exec_covr.py) for log parsing logic. Perhaps it is enough? Please let me know what you think.


src/tests/integration/utils.py
```diff
@@ -110,13 +110,28 @@ def kprove(
...
         return kprove
 
+    @pytest.fixture
+    def haskell_log_entries(self) -> Iterable[str]:
+        return ['DebugAppliedRewriteRules', 'DebugApplyEquation']
+
+    @pytest.fixture
+    def log_axioms_file(self, kprove: KProve) -> Path:
+        return kprove.use_directory / 'kore-exec.debug-log'
+
```
tothtamas28, line 120: Consider eliminating these fixtures, i.e. passing the values directly to `KCFGExplore`. As long as the values are not shared, it'd be more readable that way I think.

tothtamas28 (approved):

tothtamas28:
> We already have some [unit tests](https://github.com/runtimeverification/pyk/blob/master/src/tests/unit/test_kore_exec_covr.py) for log parsing logic. Perhaps it is enough?

I think so, yes.

Issue #306, opened 2023-03-28
Further changes to edge abstractions


ehildenb:
This PR https://github.com/runtimeverification/pyk/pull/124 introduces a new `KCFG.Split` abstraction, which gives us more structure to our KCFG. As pointed out here https://github.com/runtimeverification/pyk/pull/124#discussion_r1150255143, we can push this further.

Changes to make:

- [ ] Rename `KCFG.Edge` to `KCFG.BasicBlock` or `KCFG.Block`, to reflect that it's one basic block of execution (that will reflect the terminology used in the Summarizer paper better too).
- [ ] Rename `KCFG.EdgeLike` to `KCFG.Edge`, as a single-target edge.
- [x] Make it so that storing `stuck` is first-class, rather than storing `expanded` being first-class. `expanded` will be computed instead as "no outgoing `KCFG.Successor`"
- [x] Restrict new `KCFG.Block` and `KCFG.Split` so that they do not go backwards. Not sure this is possible, because the definition `rule a => b     rule b => a` would cause `KCFG.Block` which go backwards.
- [x] Simplify the definition of `is_leaf` to just "has no successors"? Currently we allow it to have covers and still be a leaf, but not sure why we do this. see: https://github.com/runtimeverification/pyk/pull/329/files#r1160812758

Finished:

- [x] Add a new class `KCFG.MultiEdge`, which has the `targets` variable instead of the `target` variable. Make `KCFG.Split` a subclass of `KCFG.MultiEdge`.
- [x] Add a new class `KCFG.NDBranch`, which is a `KCFG.MultiEdge`, and represents a non-deterministic split. The data should be a `targets` of the new target states that are achieved.

ehildenb:
Open to suggestions on teh Edeg names.

ehildenb:
Related: https://github.com/runtimeverification/pyk/issues/378

PR #329, opened 2023-04-04
Add nondeterministic branch nodes.

virgil-serbanuta:
This PR introduces non-deterministic splits into the KCFG.

In particular:

- Class `KCFG.MultiEdge` is factored out with multiple targets.
- Class `KCFG.NDBranch` is factored out, which represents a 1-step non-deterministic branch from source node to target nodes. The various usual facilities around this edge type are implemented.
- `KCFGExplore.extend` uses a very simple heuristic that should be sound for detecting total and mutually exclusive branches, but otherwise falls back to using `NDBranch` to insert the transition (I suspect this will fix: https://github.com/runtimeverification/pyk/issues/398).
- Tests are added of the non-determinstic functionality/graph creation.
- IMP test is updated to use the extract-branches heuristic when testing the `paths_between` functionality.
- `KCFG.prune` is modified to keep initial and target nodes in the KCFG by default, instead of removing them.


virgil-serbanuta:
@ehildenb I'm not sure if this is what you had in mind.

src/pyk/kcfg/kcfg.py
```diff
@@ -138,12 +138,44 @@ def target_ids(self) -> List[str]:
             return sorted(t.id for t, _ in self.targets)
 
         def pretty(self, kprint: KPrint) -> List[str]:
-            return ['Split node: {len(self.targets)}']
+            return [f'Split node: {len(self.targets)}']
+
+    @dataclass(frozen=True)
+    class NDBranch(Successor):
+        source: KCFG.Node
+        targets: Tuple[Tuple[KCFG.Node, Tuple[KInner, ...]], ...]
```
ehildenb, line 146: I don't think it makes sense to attach constraints to the branch edges for NDBranch. Instead should just go straight to each target in one step. Teh target may very well have more constraints attached to it, but having a constraint attached to an edge means "I only take this edge if this constaint applies", but that's not the case here (because the `ensures` clause means the constraint gets applied _after_ the edge applies).

src/pyk/kcfg/kcfg.py
```diff
@@ -138,12 +138,44 @@ def target_ids(self) -> List[str]:
...
+
+        def __init__(
+            self,
+            source: KCFG.Node,
+            targets: Iterable[Tuple[KCFG.Node, Iterable[KInner]]],
+        ):
+            object.__setattr__(self, 'source', source)
+            object.__setattr__(self, 'targets', tuple((node, tuple(constraints)) for node, constraints in targets))
```
ehildenb, line 154: Remove this `__init__` method, and remove the one on `Split` as well. They are `dataclass`, so should get their own derived `__init__`.

src/pyk/kcfg/kcfg.py
```diff
@@ -138,12 +138,44 @@ def target_ids(self) -> List[str]:
...
+
+        def __lt__(self, other: Any) -> bool:
+            if not isinstance(other, KCFG.NDBranch):
+                return NotImplemented
+            return (self.source, self.target_ids) < (other.source, other.target_ids)
```
ehildenb, line 159: Make a superclass of `Split` and `NDBranch` called `MultiEdge`, which contains this function and the `targets` property (of type `Union[Tuple[KCFG.Node, ...], Tuple[Tuple[KCFG.Node, KInner], ...]]`).

src/pyk/kcfg/kcfg.py
```diff
@@ -442,6 +494,26 @@ def _print_subgraph(indent: str, curr_node: KCFG.Node, prior_on_trace: List[KCFG
                 ret_lines.append(('edge_{curr_node.id}_{target.id}', ret_edge_lines))
                 _print_subgraph(indent + '   ', target, prior_on_trace + [curr_node])
 
+            if type(successor) is KCFG.NDBranch:
+                ret_lines.append(('unknown', [f'{indent}┃']))
```
ehildenb, line 498:
```suggestion
                ret_lines.append(('unknown', [f'{indent}┃ (1 step)']))
```

src/pyk/kcfg/kcfg.py
```diff
@@ -842,6 +923,22 @@ def create_split(self, source_id: str, splits: Iterable[Tuple[str, CSubst]]) ->
...
+        if len(ndbranches) <= 1:
+            raise ValueError(
+                f'Cannot create non-deterministic branches node with less than 2 targets: {source_id} -> {ndbranches}'
+            )
+
+        source_id = self._resolve(source_id)
+        ndbranch = KCFG.NDBranch(
+            self.node(source_id), ((self.node(nid), constraints) for nid, constraints in ndbranches)
+        )
+        self._ndbranches[source_id] = ndbranch
```
ehildenb, line 940: We should just get the list of `str` as input here, not constraints.

If constraints enable taking one or another of the branches, it should be a `Split`, not a `NDBranch`.

ehildenb:
@virgil-serbanuta please add a new definition `k-files/non-det.k` and a proof to that file, and run that proof in the test-suite `src/tests/integration/kcfg/test_non_det.py`. You can copy the contents of the file `src/tests/integration/kcfg/test_imp.py` to that new file to get the needed test harness, and adjust it to point at the new files.

This test should exercise the non-determinism, and in particular, the `AGProver.advance_proof(...)` should identify when it's a non-determinstic branching node vs something where the branches can be extracted, and build a NDBranch node instead of a Split node when needed there.

ehildenb:
After some thinking, I think we should have:

- Class `MultiEdge(Successor)` with fields `source: KCFG.Node`, `targets: Iterable[KCFG.Node]`, and property `target_ids: Iterable[str]`.
- `Split` now has another property called `splits: Dict[str, CSubst]`, which returns the `CSubst` associated with the id of a given target node.

That way, when pretty-printing this multi-edge (or when computing "reachable", or "paths between"), we can use largely the same logic for both type of `MultiEdge`, while maintaining that a `Split` stores some extra data.

Also, we should drop any extra data on the multi-edge `NDBranch`, as it's not needed.

virgil-serbanuta:
I implemented your first comment. Tomorrow I'll look into implementing your second comment.

virgil-serbanuta:
I think that git completely messed my PR, I'm not sure exactly how it managed to do that.

src/pyk/kcfg/kcfg.py
virgil-serbanuta: Fixed.

src/pyk/kcfg/kcfg.py
virgil-serbanuta: I removed only the one on `NDBranch`, after fixing your other comments, the one on `Split` actually does something useful (ensures data consistency).

src/pyk/kcfg/kcfg.py
virgil-serbanuta: Is it better now?

src/pyk/kcfg/kcfg.py
virgil-serbanuta: Fixed, although I didn't use your suggestion because the github PR seems broken and I wasn't sure if this would have broken more things (e.g. my local client when I get the latest PR version from github).

virgil-serbanuta:
> @virgil-serbanuta please add a new definition `k-files/non-det.k` and a proof to that file, and run that proof in the test-suite `src/tests/integration/kcfg/test_non_det.py`. You can copy the contents of the file `src/tests/integration/kcfg/test_imp.py` to that new file to get the needed test harness, and adjust it to point at the new files.
> 
> This test should exercise the non-determinism, and in particular, the `AGProver.advance_proof(...)` should identify when it's a non-determinstic branching node vs something where the branches can be extracted, and build a NDBranch node instead of a Split node when needed there.

Done

src/pyk/kcfg/kcfg.py
```diff
@@ -110,40 +111,67 @@ def pretty(self, kprint: KPrint, minimize: bool = True) -> Iterable[str]:
...
-            object.__setattr__(self, 'source', source)
-            object.__setattr__(self, 'targets', tuple(targets))
-
-        def __lt__(self, other: Any) -> bool:
-            if not isinstance(other, KCFG.Split):
-                return NotImplemented
-            return (self.source, self.target_ids) < (other.source, other.target_ids)
+            targets = list(targets)
+            super().__init__(source, tuple(n for n, _ in targets))
+            object.__setattr__(self, 'splits', {n.id: s for n, s in targets})
```
ehildenb, line 142: So, as far as I know, `@dataclass` modifier lets you not declare the constructor `__init__`, but you need to structure you class in a certain way:

```
@dataclass(frozen=True)
class Split(MultiEdge):
    source: KCFG.Node
    targets: Tuple[KCFG.Node, ...]
    splits: Dict[str, CSubst]
```

Then you get that `Split(source, target, splits)` constructs the `Split` class as you have.

Then you can provide convenience functions which are more compact in arguments in the `KCFG.create_split` and `KCFG.split_on_constraints` methods (which users will use), while maintaining that this is a pure dataclass, which I guess has aesthetic and computational advantages.

src/pyk/kcfg/kcfg.py
```diff
@@ -110,40 +111,67 @@ def pretty(self, kprint: KPrint, minimize: bool = True) -> Iterable[str]:
...
-            return sorted(t.id for t, _ in self.targets)
+        def pretty(self, kprint: KPrint) -> List[str]:
+            return [f'Split node: {len(self.targets)}']
+
+        def with_single_target(self, target: KCFG.Node) -> KCFG.Split:
+            return KCFG.Split(self.source, [(target, self.splits[target.id])])
+
+    @dataclass(frozen=True)
+    class NDBranch(MultiEdge):
+        def to_dict(self) -> Dict[str, Any]:
```
ehildenb, line 158:
```suggestion
    @dataclass(frozen=True)
    class NDBranch(MultiEdge):
        source: KCFG.Node
        targets: Tuple[KCFG.Node, ...]

        def to_dict(self) -> Dict[str, Any]:
```

Actually not sure if this is needed, but I've seen redeclaration of the fields on other dataclasses. Maybe @tothtamas28 can comment?

src/pyk/kcfg/kcfg.py
```diff
@@ -422,22 +458,37 @@ def _print_subgraph(indent: str, curr_node: KCFG.Node, prior_on_trace: List[KCFG
...
+            if isinstance(successor, KCFG.MultiEdge):
                 ret_lines.append(('unknown', [f'{indent}┃']))
-
-                for target, csubst in successor.targets[:-1]:
-                    ret_edge_lines = _print_split_edge(csubst)
-                    ret_edge_lines = [indent + '┣━━┓ ' + ret_edge_lines[0]] + add_indent(
-                        indent + '┃  ┃ ', ret_edge_lines[1:]
-                    )
+                if type(successor) is KCFG.NDBranch:
+                    ret_lines.append(('unknown', [f'{indent}┃ (1 step)']))
```
ehildenb, line 464:
```suggestion
                multiedge_label = '1 step' if type(successor) is KCFG.NDBranch else 'branch'
                ret_lines.append(('unknown', [f'{indent}┃ ({multiedge_label})']))
```

Maybe this would look nice?

src/pyk/kcfg/kcfg.py
```diff
@@ -908,9 +982,13 @@ def is_split(self, node_id: str) -> bool:
...
         return node_id in self._splits
 
+    def is_ndbranch(self, node_id: str) -> bool:
+        node_id = self._resolve(node_id)
+        return node_id in self._ndbranches
+
     def is_leaf(self, node_id: str) -> bool:
         node_id = self._resolve(node_id)
-        return node_id not in self._edges and node_id not in self._splits
+        return node_id not in self._edges and node_id not in self._splits and node_id not in self._ndbranches
```
ehildenb, line 991: Could this be implemented as `return len(self.successors(node_id)) > 0`?

Probably no need to call `self._resolve` either then?

src/pyk/kcfg/kcfg.py
```diff
@@ -1036,11 +1111,34 @@ def reachable_nodes(self, source_id: str, *, reverse: bool = False, traverse_cov
...
+        used_variables: Dict[str, List[KVariable]] = var_occurrences(ct.config)
+        has_constraint = False
+        for constraint in ct.constraints:
+            if not constraint in ct.constraints:
+                if has_unused_questionmark_variables(constraint, used_variable_names=used_variables):
+                    continue
+                has_constraint = True
+        if not has_constraint:
+            return True
+    return False
```
ehildenb, line 1144: Let's move this function to where it's used, as `AGProver._is_nd_branch`, not here.

src/pyk/kcfg/kcfg.py
ehildenb: The data classes should just be holding data, it's the helpers that create them in `KCFG` that ensure consistency of the overall `KCFG` that is created.

src/pyk/kcfg/kcfg.py
ehildenb: Yeah, GitHub is consistently messing up the display of diffs.

src/pyk/proof/reachability.py
```diff
@@ -137,15 +138,39 @@ def advance_proof(
...
         return self.proof.kcfg
+
+    @staticmethod
+    def is_nd_branch(next_cterms: Iterable[CTerm]) -> bool:
+        def has_unused_questionmark_variables(term: KInner, *, used_variable_names: dict[str, Any]) -> bool:
+            return (
+                len([var for var in var_occurrences(term) if var not in used_variable_names and var.startswith('?')])
+                > 0  #
+            )
+
```
ehildenb, line 165: I don't think we use question marks at all for the RPC prover do we? We manually put things inside existentials if that's needed.

Question marks are a hint to the frontend about existentially binding things, be we strip out all those decorators when using the RPC server. Would be curious if you found a case where it's not happening.

ehildenb: I think in general, to detect nd branches, we should do the normal branch extraction routine below (see if we can identify a unique constraint per branch term). Then if we have all teh unique constraints, and we `orBool` them together, it should be `true`, so we could compute `{ true #Equals c1 orBool c2 orBool ... orBool cN }`, and call `simplify` on it. If it does not come back `#Top`, then assume it's a non-deterministic branch instead.

src/tests/unit/test_kcfg.py
```diff
@@ -463,7 +500,16 @@ def _short_hash(i: int) -> str:
         f'┣━━┓ constraint: #Equals ( x , 19 )\n'
         f'┃  ┃ subst: V13 <- V19\n'
         f'┃  │\n'
-        f'┃  └─ \033[1m{_short_hash(19)} (frontier, leaf)\033[0m\n'
+        f'┃  ├─ {_short_hash(19)} (expanded)\n'
+        f'┃  ┃\n'
+        f'┃  ┃ (1 step)\n'
+        f'┃  ┣━━┓\n'
+        f'┃  ┃  │\n'
+        f'┃  ┃  └─ \033[1m{_short_hash(23)} (frontier, leaf)\033[0m\n'
```
ehildenb, line 508: Hmmm, I'm not sure I like this printing actually. Would it make more sense to have `(nd step)`, and have it printed on _each_ branch, not just before the branches? Or to have `(nd step)` and have an extra blank line after it? Could do the same for the `(branch)` too.

What do you think?

src/pyk/kcfg/kcfg.py
ehildenb: Ah, it looks like you're correct.... I wonder why that choice about `leaf` was made. I've mentioned it here: https://github.com/runtimeverification/pyk/issues/306

src/pyk/kcfg/kcfg.py
ehildenb: So my thinking is that we're trying to amke it so that constructing edges/splits/etc... directly is allowed in invalid flexible ways, but going through the `KCFG.create_...` methods enforces the overall KCFG sanity checks. That way we can do hacky things if needed with the datastructures directly, but still should have safeguards against making malformed KCFGs.

It looks good now, feel free to resolve!

src/pyk/kcfg/kcfg.py
tothtamas28: `MultiEdge` is a dataclass itself, so it is not needed. But it is also abstract, so I think it shouldn't be a dataclass.

src/pyk/kcfg/kcfg.py
```diff
@@ -854,9 +909,23 @@ def create_split(self, source_id: str, splits: Iterable[tuple[str, CSubst]]) ->
...
+
+        ndbranches = list(ndbranches)
+
+        if len(ndbranches) <= 1:
+            raise ValueError(
+                f'Cannot create non-deterministic branches node with less than 2 targets: {source_id} -> {ndbranches}'
+            )
+
+        source_id = self._resolve(source_id)
+        ndbranch = KCFG.NDBranch(self.node(source_id), tuple(self.node(nid) for nid in ndbranches))
```
ehildenb, line 926:
```suggestion
        ndbranch = KCFG.NDBranch(self.node(source_id), tuple(self.node(nid) for nid in ndbranches))
```

ehildenb:
There are several more improvements/refactors I want to make to the KCFG at this point, to simplify the overall algorithms, but they will have to wait for a future PR as this is already growing quite large. In particular:

- The `show` functionality should be factored out to `KCFGShow`.
- Anywhere that we are having separate logic for processing different types of successors likely can be made more compact.

ehildenb:
https://github.com/runtimeverification/pyk/pull/403 has taken care of refactoring `KCFGShow`

ehildenb:
I've confirmed that this new `test_non_det.py`, if run with master version of pyk, has the following behavior:

- Crashes if you do `KCFG.from_dict(kcfg.to_dict())` of the generated KCFG, which is the observed crash here: https://github.com/runtimeverification/pyk/issues/398.
- Crashes with index out of bounds error if that crash is skipped by not serializing to JSON.

This PR gets us past both crashes.

ehildenb:
I don't want to merge this until https://github.com/runtimeverification/evm-semantics/pull/1784 is merged, because it has the potential to break many things in KEVM and I want to make sure that KEVM is working with the version before this one ahead of merging this.

ehildenb:
I've confirmed that we get basically the same output on KEVM using this branch as on current KEVM master.

nwatson22 (approved):

PR #335, opened 2023-04-06
Bump minimal Python version to 3.10

### LINKED ISSUE #135: Bump Python version to `^3.10`
This issue is for discussing PROs and CONs / tracking TODOs.

* [What’s New In Python 3.9](https://docs.python.org/3/whatsnew/3.9.html)
* [What’s New In Python 3.10](https://docs.python.org/3/whatsnew/3.10.html)

---

TODOs after version bump:

- [ ] Enable `flake8-bugbear` warning `B905`: `zip() without an explicit strict= parameter set.`
- [ ] Change `@lru_cache(maxsize=None)` to `@cache`
- [ ] ...

tothtamas28:
Fixes #135

* Update Python version to `^3.10`
* Run `pyupgrade`


src/pyk/cli_utils.py
```diff
@@ -97,15 +98,15 @@ def is_relative_to(_self: Path, other: Path) -> bool:
...
     *,
     check: bool = True,
-    input: Optional[str] = None,
+    input: str | None = None,
     pipe_stdout: bool = True,
     pipe_stderr: bool = False,
-    cwd: Optional[Union[str, Path]] = None,
-    env: Optional[Mapping[str, str]] = None,
-    logger: Optional[Logger] = None,
+    cwd: str | Path | None = None,
```
ehildenb, line 107: And in this case I think I prefer `Optional[str | Path]`

Just a nitpick. Does our styling enforce one way or another?

src/pyk/coverage.py
```diff
@@ -48,7 +48,7 @@ def translate_coverage(
     dst_all_rules: Iterable[str],
     dst_definition: KDefinition,
     src_rules_list: Iterable[str],
-) -> List[str]:
+) -> list[str]:
```
ehildenb, line 51: Why this difference? I think of `list` as a value and `List` as a type. Any benefit to going with the lowercase one?

ehildenb (approved):

src/pyk/coverage.py
tothtamas28: `list` is a type, (i.e. `type(list) == type`). `List` was necessary in Python < 3.9 becasue in those versions `list` didn't support subscripting, hence generics. ([PEP 585](https://peps.python.org/pep-0585/))

src/pyk/cli_utils.py
tothtamas28: This has been automatically changed by `pyupgrade --py310-plus`. We use the tool on CI, so it's enforced.

src/pyk/cli_utils.py
```diff
@@ -97,15 +98,15 @@ def is_relative_to(_self: Path, other: Path) -> bool:
 
 
 def run_process(
-    args: Union[str, Iterable[str]],
+    args: str | Iterable[str],
     *,
     check: bool = True,
-    input: Optional[str] = None,
+    input: str | None = None,
```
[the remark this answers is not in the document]
tothtamas28: This is also an automatic `pyupgrade --py310-plus` change.

PR #344, opened 2023-04-12
Add missing imports for `pyk.kllvm.load`

tothtamas28:
The bug didn't manifest as test discovery imports all the test modules, and one of them already imported the module.


Scott-Guest (approved):
LGTM, fixes the issue on my machine

PR #347, opened 2023-04-12
AGBMCProver: implement basic bounded-model-checker prover

ehildenb:
Part of: https://github.com/runtimeverification/pyk/issues/310
~Blocked on: https://github.com/runtimeverification/pyk/pull/362~

Most Ethereum verification tools implement Bounded Model Checking, which is where they unroll each loop a fixed number of times. This improves their stability to code changes because they don't need to provide loop invariants.

This PR introduces bounded model checking as a proof technique to the pyk library:

- Class `AGBMCProof` is introduced, which is an `AGProof`, but it stores also a list of states that are bounded due to reaching loop-unroll bound.
- Class `AGBMCProver` is introduced, which is an `AGProver`, but every iteration of calling the `AGProver` it also checks if any of the new nodes have reached the depth bound. It accepts the as well a `same_loop` predicate which allows identifying loop heads for BMC purposes.
- For IMP, the `_is_terminal` check is extended to also consider `<k> SOME_VAR </k>` as terminal, os that the backend doesn't start inventing program by narrowing on that case.
- A test of the BMC prover is added which is not true in the unbounded case, but is true with bounded loop unrolling.
- A test of the BMC prover is added which fails in the code following the unrolled loop.


src/pyk/proof/reachability.py
```diff
@@ -68,6 +68,72 @@ def summary(self) -> Iterable[str]:
...
+        return AGBMCProof(id, cfg, bmc_depth, bounded_states=bounded_states, proof_dir=proof_dir)
+
+    @property
+    def dict(self) -> dict[str, Any]:
+        return {
+            'type': 'AGBMCProof',
+            'id': self.id,
+            'cfg': self.kcfg.to_dict(),
+            'bmc_depth': self.bmc_depth,
+            'bounded_states': list(self._bounded_states),
```
ehildenb, line 120:
```suggestion
            'bounded_states': self._bounded_states,
```

src/pyk/proof/reachability.py
```diff
@@ -141,44 +207,111 @@ def advance_proof(
...
+                        for nd in self.proof.kcfg.reachable_nodes(f.id, reverse=True, traverse_covers=True)
+                        if nd.id != f.id and self._same_loop(nd.cterm, f.cterm)
+                    ]
+                    if len(prior_loops) >= self.proof.bmc_depth:
+                        self.proof.kcfg.add_expanded(f.id)
+                        self.proof.bound_state(f.id)
+            if max_iterations is not None and max_iterations <= iterations:
+                _LOGGER.warning(f'Reached iteration bound {self.proof.id}: {max_iterations}')
+                break
+            iterations += 1
```
ehildenb, line 303:
```suggestion
            if max_iterations is not None and max_iterations <= iterations:
                _LOGGER.warning(f'Reached iteration bound {self.proof.id}: {max_iterations}')
                break
            iterations += 1

            for f in self.proof.kcfg.frontier:
                if f.id not in self._checked_nodes:
                    self._checked_nodes.append(f.id)
                    prior_loops = [
                        nd.id
                        for nd in self.proof.kcfg.reachable_nodes(f.id, reverse=True, traverse_covers=True)
                        if nd.id != f.id and self._same_loop(nd.cterm, f.cterm)
                    ]
                    if len(prior_loops) >= self.proof.bmc_depth:
                        self.proof.kcfg.add_expanded(f.id)
                        self.proof.bound_state(f.id)
```

ehildenb:
Here is an attachment which contains the KCFGs  for the new BMC proofs. It would be good to confirm that your intuition of what BMC is supposed to produce is indeed what is produced here. They are in the same order there as they are listed in the `test_imp.py` file.

[imp-bmc-proofs.txt](https://github.com/runtimeverification/pyk/files/11228439/imp-bmc-proofs.txt)

src/pyk/proof/reachability.py
```diff
@@ -68,6 +68,72 @@ def summary(self) -> Iterable[str]:
         ]
 
 
+class AGBMCProof(AGProof):
```
tothtamas28, line 71:
```suggestion
class BMCProof(Proof):
```

* Consider `BMCProof` as name for conciseness.
* Having `AGProof` as a superclass might not be justified in this case, consider inheriting from `Proof` instead.

src/pyk/proof/reachability.py
```diff
@@ -141,44 +207,112 @@ def advance_proof(
...
                     f'Found {len(branches)} branches for node {self.proof.id}: {shorten_hashes(curr_node.id)}: {[kcfg_explore.kprint.pretty_print(bc) for bc in branches]}'
                 )
                 self.proof.kcfg.split_on_constraints(curr_node.id, branches)
+                continue
+
+        self.proof.write_proof()
+        return self.proof.kcfg
+
+
+class AGBMCProver(AGProver):
```
tothtamas28, line 263: If `AGBMCProof` is no longer an `AGProof`, then  you'll probably need this change as well:

```suggestion
class AGBMCProver(Prover):
```

(Also, consider changing the name to `BMCProver`.)

src/pyk/proof/reachability.py
```diff
@@ -141,44 +207,112 @@ def advance_proof(
...
+                    prior_loops = [
+                        nd.id
+                        for nd in self.proof.kcfg.reachable_nodes(f.id, reverse=True, traverse_covers=True)
+                        if nd.id != f.id and self._same_loop(nd.cterm, f.cterm)
+                    ]
+                    if len(prior_loops) >= self.proof.bmc_depth:
+                        self.proof.kcfg.add_expanded(f.id)
+                        self.proof.bound_state(f.id)
+
+            super().advance_proof(
```
tothtamas28, line 306: If `AGBMCProver` is no longer an `AGProver`, then this will no longer work. The solution is to either extract this exploration strategy into a helper class or function, or if you want to have it in this class hierarchy, then to extract a common abstract base class.

src/pyk/proof/reachability.py
ehildenb: We'll be renaming these classes soon anyway, so I'll leave it to that: https://github.com/runtimeverification/pyk/issues/355

src/pyk/proof/reachability.py
ehildenb: Here is the proposed new design: https://github.com/runtimeverification/pyk/issues/361

src/pyk/proof/reachability.py
ehildenb: Here is the proposed new design, to be implemented after this PR is merged (so @geo2a can make progress on setting up BMC for KEVM).

src/pyk/proof/reachability.py
```diff
@@ -137,3 +204,63 @@ def advance_proof(
...
+
+            for f in self.proof.kcfg.frontier:
+                if f.id not in self._checked_nodes:
+                    self._checked_nodes.append(f.id)
+                    prior_loops = [
+                        nd.id
+                        for nd in self.proof.kcfg.reachable_nodes(f.id, reverse=True, traverse_covers=True)
+                        if nd.id != f.id and self._same_loop(nd.cterm, f.cterm)
+                    ]
+                    if len(prior_loops) >= self.proof.bmc_depth:
```
ehildenb, line 252: This should be:

`>`, so that with `--bmc-depth 1`, for example, it's asserting things about the path where it was unrolled once, instead of throwing it away.

src/tests/integration/kcfg/test_imp.py
```diff
@@ -211,6 +211,84 @@
...
+    tuple[str, str, str, str, int | None, int | None, int, Iterable[str], Iterable[str], ProofStatus]
+] = (
+    (
+        'bmc-loop-concrete-1',
+        'k-files/imp-simple-spec.k',
+        'IMP-SIMPLE-SPEC',
+        'bmc-loop-concrete',
+        20,
+        20,
+        1,
```
ehildenb, line 224: All these depth bounds should be adjusted down by 1.

PR #351, opened 2023-04-13
Fix for bug report rpc commands generation

### LINKED ISSUE #349: `--bug-report` does not save rpc calls correctly
When running pyk/kevm-pyk with the bug report option, the current way rpc calls are saved leads to lost calls. This is because during a run, every time pyk closes and re-opens a new rpc-server, the request id's start from 1. as a result the new calls don't seem to be added to the tar. This is evident in the `commands/` directory which has:

`001.sh`:
```
#!/usr/bin/env bash
set -euxo pipefail
cat rpc/001_request.json | nc -Nv localhost 64118 > rpc/001_actual.json
```
and later `008.sh`:
```
#!/usr/bin/env bash
set -euxo pipefail
cat rpc/001_request.json | nc -Nv localhost 64126 > rpc/001_actual.json
```
 notice how the port is different but the files are the same, despite them being calls to a different server

goodlyrottenapple:
Fixes #349

This PR adds a unique id of the current rpc server process to the bug report to distinguish different instances of the rpc server across the whole run.


src/pyk/kore/rpc.py
```diff
@@ -110,7 +110,8 @@ def request(self, method: str, **params: Any) -> dict[str, Any]:
         _LOGGER.info(f'Sending request to {server_addr}: {old_id} - {method}')
         req = json.dumps(payload)
         if self._bug_report:
-            bug_report_request = f'rpc/{old_id:03}_request.json'
+            id_with_port = f'{self._port}_{old_id:03}'
+            bug_report_request = f'rpc/{id_with_port}_request.json'
```
tothtamas28, line 114: What do you think?

```suggestion
            bug_report_request = f'rpc_{bug_report_id}/{old_id:03}_request.json'
```

where `bug_report_id` is based on `id(self)`, `self._port`, object creation timestamp, or some combination of these (maybe hashed).

tothtamas28: Maybe it should be the responsibility of `BugReport` to create a unique folder (with an optional prefix) for clients and return it's name.

goodlyrottenapple: I went with `id(self)` which works correctly and is pretty simple and readable.

tothtamas28 (approved):

PR #352, opened 2023-04-13
Include summarization capabilities in Proof class

ehildenb:
This enriches the Proof class heirarchy to allow the subclasses to specify compact way to summarize the proof state:

- Method `proof_exists` is promoted from `AGProof` to `Proof`.
- Abstract method `Proof.summary` is added.
- Implementation of `Proof.summary` is provided to the `AGProof` class.
- The check for `is_terminal` in `AGProver` is factored out into its own method like `_check_subsumed`.
- An implementation of `is_terminal` is provided for the IMP prover integration tests.
- The IMP prover integration tests also allow specifying failing proofs, by passing in teh resulting proof status as a parameter to the test.


src/pyk/proof/reachability.py
```diff
@@ -63,6 +59,24 @@ def from_dict(cls: type[AGProof], dct: Mapping[str, Any], proof_dir: Path | None
...
+            f'    stuck: {len(self.kcfg.stuck)}',
+        ]
+
+    def pretty(
+        self,
+        kprint: KPrint,
+        minimize: bool = True,
+        node_printer: Callable[[CTerm], Iterable[str]] | None = None,
+        omit_node_hash: bool = False,
+        **kwargs: Any,
```
ehildenb, line 76:
```suggestion
```

nwatson22 (commented):
Looks good. I was trying to compare behavior to foundry-list in evm-semantics, but it turned out that was broken due to a change in the cfg format. But this should be able to be used there in place of the CfgStat class.

nwatson22 (approved):

PR #356, opened 2023-04-13
Fix crashes in KCFG viewer

ehildenb:
This PR fixes 4 types of crashes/problems that the KCFG viewer was having:

- When the same node shows up multiple times in the KCFG view, the pretty_segments routine appends suffixes to susequent ones. This is to make sure that each segment gets a unique id for TUI purposes. But this causes lookup errors when trying to display the given node in the more detailed view. So now we always make sure to strip out these suffixes when looking up the node/edge to display in the viewer.
- There was an f-string error, where we were not actually doing substitution for the edge identifier on splits.
- The KCFG viewer did not handle `KCFG.Split` nodes correctly, this adds that functionality.
- When a given click results in a node that doesn't exist (like an `unknown` segment from the KCFG), instead of raising an AssertionError, we should just continue.


ehildenb:
This is how we display split nodes for now: 
image

nwatson22 (approved):
I confirmed I'm not getting any more click crashes with this branch. And the new display for splits looks good as well. This was an annoying bug, so thanks for getting rid of it.

PR #366, opened 2023-04-17
Add KCFG function for computing path constraints

nwatson22:
This PR adds a new function `path_constraints(str) -> KInner` to the `KCFG` class, which takes a node ID and computes the constraints added at split edges along the shortest path to that node. If there are any cover edges along the path, it will apply the cover's substitution to the collected constraint to make sure the constraints are in terms of the original variables.


src/pyk/kcfg/kcfg.py
```diff
@@ -961,6 +961,20 @@ def prune(self, node_id: str) -> None:
...
+        paths = self.paths_between(self.get_unique_init().id, final_node_id)
+        if len(paths) == 0:
+            raise ValueError(f'No path found to specified node: {final_node_id}')
+        if len(paths) > 1:
+            raise ValueError(f'Multiple paths found to specified node: {final_node_id}')
+        path = paths[0]
+        for edge in path:
+            if type(edge) is KCFG.Split:
+                for _, csubst in edge.targets:
+                    constraints += csubst.constraints
```
ehildenb, line 975: We need to be collecting the substitutions and covers as well.

So if we have:

```
ND1 ---[ Cover(constraint=#Top, subst=alpha) ]---> ND2 ---[ Split(constraint=B) ]---> ND3
```

We need to return `alpha(B)` as the path-condition. That is, covers get applied to the path condition that comes _after_ them.

If we have something like this:

```
ND1 ---[Split(constraint=B, subst=alpha)]---> ND2
```

We need to return `alpha.ml_pred #And B`

src/tests/integration/kcfg/test_imp.py
```diff
@@ -210,6 +210,21 @@
...
+] = (
+    (
+        'imp-simple-fail-branch',
+        'k-files/imp-simple-spec.k',
+        'IMP-SIMPLE-SPEC',
+        'fail-branch',
+        None,
+        None,
+        ['IMP-VERIFICATION.halt'],
+        ['IMP-VERIFICATION.halt'],
```
ehildenb, line 224:
```suggestion
        ['IMP-VERIFICATION.halt'],
        [],
```

src/tests/integration/kcfg/test_imp.py
```diff
@@ -368,3 +383,44 @@ def test_all_path_reachability_prove(
...
+    def test_collect_path_constraints(
+        self,
+        kprove: KProve,
+        kcfg_explore: KCFGExplore,
+        test_id: str,
+        spec_file: str,
+        spec_module: str,
+        claim_id: str,
+        max_iterations: int,
+        max_depth: int,
```
ehildenb, line 402: Here you have `max_iterations` and `max_depth` as `int`, but in your test you have them as `None`, so the type-checking doesn't amke sense. Just hardcode the correct max-depth/max-iterations for every test I think.

src/pyk/kcfg/kcfg.py
```diff
@@ -961,6 +961,20 @@ def prune(self, node_id: str) -> None:
         for node in nodes:
             self.remove_node(node.id)
 
+    def path_constraints(self, final_node_id: str) -> list[KInner]:
+        constraints: list[KInner] = []
+        paths = self.paths_between(self.get_unique_init().id, final_node_id)
+        if len(paths) == 0:
+            raise ValueError(f'No path found to specified node: {final_node_id}')
+        if len(paths) > 1:
+            raise ValueError(f'Multiple paths found to specified node: {final_node_id}')
```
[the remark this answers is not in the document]
nwatson22: I added the new method, and now just use that to find the constraints for the shortest path
 when there is more than one path.

src/pyk/kcfg/kcfg.py
nwatson22: I added logic to deal with this. As we discussed, it's not tested right now, so could you make sure it's doing the right thing?

src/tests/integration/kcfg/test_imp.py
nwatson22: I removed the halt as a cut rule. Curiously, having it as a terminal rule started working again, so I'm guessing the issue I was having with the test extending execution beyond `#FINISH` may have been caused by a bug.

src/tests/integration/kcfg/test_imp.py
nwatson22: The typing I used here was following the same pattern used for other tests in the same file, but yes, it doesn't make sense. I would think the type checker would catch that. I updated it on the other tests in that file to be consistent.

src/pyk/kcfg/kcfg.py
```diff
@@ -968,6 +968,31 @@ def prune(self, node_id: str) -> None:
...
+        path = self.shortest_path_between(self.get_unique_init().id, final_node_id)
+        if path is None:
+            raise ValueError(f'No path found to specified node: {final_node_id}')
+        for edge in path:
+            if type(edge) is KCFG.Split:
+                for _, csubst in edge.targets:
+                    constraints.append(apply_substs(csubst.constraint))
+            if type(edge) is KCFG.Cover:
+                cover_substs.append(edge.csubst)
+        return constraints
```
ehildenb, line 994:
```suggestion
        curr_constraint = mlTop()
        curr_subst = Subst({})
        for edge in reversed(path):
            if type(edge) is KCFG.Split:
                assert len(edge.targets) == 1
                _, csubst = edge.targets[0]
                curr_constraint = mlAnd([curr_subst(csubst.constraint), curr_constraint])
                curr_subst = csubst.subst(curr_subst)
            if type(edge) is KCFG.Cover:
                curr_constraint = edge.csubst(curr_constraint)
        return curr_constraint
```

Perhaps this is simpler, and doesn't require the weird stateful `apply_substs` method?

nwatson22: It's simpler and looking at this now I realize my version needs to be applying the substitutions in  the other order. I'm not sure what the line
```
                curr_subst = csubst.subst(curr_subst)
```
is doing, though. The way I understand, the substitution gets applied when we hit it going backwards, and so there would be no need to chain it onto the current substitution. The version I've pushed just now is:
```
        curr_constraint: KInner = mlTop()
        for edge in reversed(path):
            if type(edge) is KCFG.Split:
                assert len(edge.targets) == 1
                _, csubst = edge.targets[0]
                curr_constraint = mlAnd([csubst.constraint, curr_constraint])
            if type(edge) is KCFG.Cover:
                curr_constraint = edge.csubst.apply(CTerm.from_kast(curr_constraint)).kast
        return curr_constraint

File: src/pyk/kcfg/kcfg.py [hunk 6]
```diff
@@ -968,6 +969,24 @@ def prune(self, node_id: str) -> None:
...
+    def path_constraints(self, final_node_id: str) -> KInner:
+        path = self.shortest_path_between(self.get_unique_init().id, final_node_id)
+        if path is None:
+            raise ValueError(f'No path found to specified node: {final_node_id}')
+        curr_constraint: KInner = mlTop()
+        for edge in reversed(path):
+            if type(edge) is KCFG.Split:
+                assert len(edge.targets) == 1
+                _, csubst = edge.targets[0]
+                curr_constraint = mlAnd([csubst.constraint, curr_constraint])
```
ehildenb, line 985:
```suggestion
                _, csubst = edge.targets[0]
                curr_constraint = mlAnd([csubst.subst.ml_pred, csubst.constraint, curr_constraint])
```

The substitution here represents constarints on variables tat we should include in the path condition.

src/pyk/kcfg/kcfg.py
```diff
@@ -968,6 +969,24 @@ def prune(self, node_id: str) -> None:
...
+            if type(edge) is KCFG.Cover:
+                curr_constraint = edge.csubst.apply(CTerm.from_kast(curr_constraint)).kast
```
ehildenb, line 987:
```suggestion
                curr_constraint = mlAnd([edge.csubst.constraint, edge.csubst.subst.apply(curr_constraint)])
```

Turning `curr_constraint` into a `CTerm` will probably _always_ fail, because it does not have a configuration.

Instead, we should project out the `subst` and just apply that to the constraint directly as a `KInner`.

We could also add the `edge.csubst.constraint` with the above `mlAnd` added too. That should be _ok_ to do, because that constarint should already exist on the path leading up to this point for us to add it like that. We may just not want to include those constraints eventually, but we can figure that out later.

src/pyk/kcfg/kcfg.py
nwatson22: I made the change suggested. The example I have been working with doesn't have anything in the subst field. Are substs sometimes generated on split edges?

src/pyk/kcfg/kcfg.py
nwatson22: I changed this.

src/pyk/kcfg/kcfg.py
ehildenb: They can be, but not for IMP in this case. But in general yes.

src/pyk/kcfg/kcfg.py
```diff
@@ -968,6 +969,24 @@ def prune(self, node_id: str) -> None:
...
+            raise ValueError(f'No path found to specified node: {final_node_id}')
+        curr_constraint: KInner = mlTop()
+        for edge in reversed(path):
+            if type(edge) is KCFG.Split:
+                assert len(edge.targets) == 1
+                _, csubst = edge.targets[0]
+                curr_constraint = mlAnd([csubst.subst.ml_pred, csubst.constraint, curr_constraint])
+            if type(edge) is KCFG.Cover:
+                curr_constraint = mlAnd([edge.csubst.constraint, edge.csubst.subst.apply(curr_constraint)])
+        return curr_constraint
```
ehildenb, line 988:
```suggestion
        return mlAnd(flatten_label('#And', curr_constraint))
```

This will get rid of the annoying `#Top #And ...` at the beginning of the constraint in your example.

src/tests/integration/kcfg/test_imp.py
```diff
@@ -258,6 +263,22 @@
...
+    (
+        'imp-simple-fail-branch-unlimited-iterations',
+        'k-files/imp-simple-spec.k',
+        'IMP-SIMPLE-SPEC',
+        'fail-branch',
+        None,
+        1,
+        ['IMP-VERIFICATION.halt'],
+        [],
+        '#Top #And { false #Equals _S:Int <=Int 123 }',
```
ehildenb, line 278:
```suggestion
        '{ false #Equals _S:Int <=Int 123 }',
```

ehildenb (approved):

src/pyk/kcfg/kcfg.py
nwatson22: It looks like `flatten_label` doesn't remove the `#Top` term.

nwatson22: I  changed it to use `ml_pred_to_bool` and then `simplify_bool` to simplify the output. I'm not sure if we have a preference of format for the output, but this might make it more readable.

src/pyk/kcfg/kcfg.py
```diff
@@ -985,7 +986,7 @@ def path_constraints(self, final_node_id: str) -> KInner:
                 curr_constraint = mlAnd([csubst.subst.ml_pred, csubst.constraint, curr_constraint])
             if type(edge) is KCFG.Cover:
                 curr_constraint = mlAnd([edge.csubst.constraint, edge.csubst.subst.apply(curr_constraint)])
-        return mlAnd(flatten_label('#And', curr_constraint))
+        return simplify_bool(ml_pred_to_bool(mlAnd(flatten_label('#And', curr_constraint))))
```
ehildenb, line 989: We almost certainly don't want to make this change, it could be unsound.

Issue #12, opened 2022-03-24
[ksimplify] - Tool for simplifying matching logic patterns


ehildenb:
Blocked on: https://github.com/runtimeverification/haskell-backend/issues/3298

- `ksimplify [<kast-arg>*] <pattern-file>`: Parse in `<pattern-file>` (containing potential symbolic variables), simplify it, and print the result. All the same sort/parsing options handled by `kast` are handled here.

One approach to implementing these features is document here: https://docs.google.com/drawings/d/1XIjyPCjB6GBEDCw1S_0ik-kvQPrTctqYINOnVI1Ww_I/edit

- [x] Haskell backend implements `kore-simplify`: https://github.com/runtimeverification/haskell-backend/issues/3011
- [ ] Frontend implements new `ksimplify` command, which has basically the same interface as `kast`, but invokes `kore-simplify` between parsing and unparsing.

radumereuta:
Ana: We should remove `kore-simplify`, pyk should implement a wrapper over the simplify endpoint provided by the `kore-rpc` server. The summarizer should provide the backend with simplification tests to help us reimplement the simplifier component.

ehildenb:
Reopen if needed still. We mostly use the RPC server directly anymore.

PR #372, opened 2023-04-19
Support for functional claims via EqualityProof and EqualityProver

ehildenb:
This sets up a specialized prover for equality/functional proofs, which can be discharged much faster than reachability proofs. The motivation for this comes from the fact that we are currently using the reachability prover for KEVM equality proofs (via the "runLemma => doneLemma" style proofs), and it's quite a bit slower than it should be (many calls to RPC server instead of just one call to simplify).

- A simple test of an equality proof for IMP is added.
- The class `EqualityProof` is added, which stores as data all the components of the equality, as well as the simplified state of the equality. In addition, a convenience method `EqualityProof.from_claim` is provided for extracting `EqualityProof` from a `KClaim`.
- The class `EqualityProver` is added which holds an `EqualityProof` and is able to "advance" it by calling the simplifier on the underlying equality. If it makes progress, the simplified equality is stored as part of the proof state.


src/pyk/proof/equality.py
```diff
@@ -0,0 +1,158 @@
...
+
+_LOGGER: Final = logging.getLogger(__name__)
+
+
+class EqualityProof(Proof):
+    lhs_body: KInner
+    rhs_body: KInner
+    sort: KSort
+    lhs_constraints: Iterable[KInner]
+    rhs_constraints: Iterable[KInner]
```
tothtamas28, line 32:
```suggestion
    lhs_constraints: tuple[KInner, ...]
    rhs_constraints: tuple[KInner, ...]
```

src/tests/integration/kcfg/test_imp.py
```diff
@@ -495,3 +504,28 @@ def test_all_path_bmc_reachability_prove(
         )
 
         assert proof.status == proof_status
+
+    @pytest.mark.parametrize(
+        'test_id,spec_file,spec_module,claim_id,proof_status',
+        FUNC_PROVE_TEST_DATA,
+        ids=[test_id for test_id, *_ in FUNC_PROVE_TEST_DATA],
+    )
+    def test_functional_prove(
```
tothtamas28, line 513: Consider moving this to `tests/integration/proof/test_equality.py`.

tothtamas28 (approved):
Great feature! :rocket:

k-files/imp-simple-spec.k
```diff
@@ -1,5 +1,12 @@
 requires "imp-verification.k"
 
+module IMP-FUNCTIONAL-SPEC
+    imports IMP-VERIFICATION
+
+    claim [concrete-addition]: 3 +Int 4 => 7
+
+endmodule
```
ana-pantilie, line 8: What does the frontend transform this into?

ehildenb: I don't think it sends it to kore, but I'm not sure. Here, we are using `KProve.get_claims`, which basically calls `kprove --dry-run --emit-json ...`. That is a KAST file, not a Kore file, which we read in directly, and then we use the RPC server to discharge it as follows:

A rule:

```
rule LHS => RHS requires REQ ensures ENS
```

Gets sent to the Haskell backend `simplify` endpoint as:

```
{ LHS #And REQ #Equals RHS #And ENS }
```

ehildenb: So very little of the logic is happening in frontend, it's happening in Python, the frontend is just parsing things for us.

src/pyk/proof/equality.py
ehildenb: Done.

ehildenb:
This doesn't make #12 obsolete, because we still want to provide a CLI tool which users can call saying `pyk simplify ...` where they jjust provide a K term, and it produces the simplified variant of the term.

PR #377, opened 2023-04-20
Fix bug in how extract_branches heuristic is applied in AGProver

ehildenb:
Currently, when we have an `extract_branches` heuristic, we'll try it before trying anything else. This can lead to the heuristic being tried twice in a row, which leads to a double split, where all leaf-nodes are considered proven (either they're infeasible, with contradictory constraints, or they are duplicate, so the KCFG things it's looped backwards). This bug was discovered independently by @lucasmt and @nwatson22 .

This PR fixes it a little bit:

- Add a test on IMP which should be failing, but is passing if you supply an `extract_branches` heuristic.
- Change how the `extract_branches` heuristic is applied, to not consider nodes that are the results of splits.


ehildenb:
I've confirmed that this seems to give more reasonable KCFGs on failing proofs in KEVM.

tothtamas28 (approved):

nwatson22 (approved):

PR #379, opened 2023-04-20
Add support for the --enable-llvm-debug  flag in kompile

anvacaru:
part of: https://github.com/runtimeverification/evm-semantics/issues/1670


src/pyk/ktool/kompile.py
```diff
@@ -109,6 +111,7 @@ def kompile(
         ccopts=ccopts,
         no_llvm_kompile=no_llvm_kompile,
         concrete_rules=concrete_rules,
+        enable_llvm_debug=enable_llvm_debug,
```
tothtamas28, line 114: Consider ordering arguments in the order of parameters in the function signature.

src/pyk/ktool/kompile.py
```diff
@@ -56,6 +56,7 @@ def kompile(
     ccopts: Iterable[str] = (),
     no_llvm_kompile: bool = False,
     enable_search: bool = False,
+    enable_llvm_debug: bool = False,
```
tothtamas28, line 59: In all related function definitions and function calls, consider ordering arguments in the order of parameters established here.

tothtamas28 (approved):

src/pyk/ktool/kompile.py
ehildenb: @anvacaru this was not addressed. Either respond to the comment, or address the issue.

Also, I believe that you missed some spots where you need to thread the argument through to. In this PR, for example (https://github.com/runtimeverification/pyk/pull/375/files), I had to thread it through to 8 places. In your PR, I only see it theraded through to 6 places.

anvacaru: My bad, I'll open a new PR to address this.

Issue #355, opened 2023-04-13
Should `AGProver/AGProof` actually be `AFProver/AFProof`?


ehildenb:
@h0nzZik we picked `AGProof` from the CTL terminology: https://en.wikipedia.org/wiki/Computation_tree_logic

But `AG` means "all globally", which means that once the property holds, it continues to hold into the future (as far as I can tell). But that's not what our prover is checking, it's just checking that it holds at a specific point in the future, and doesn't care about states after that. I think what we have implemented is closer to `AF`, which is "all finally", meaning on every finite path of execution, there is some state where the target property holds.

Proof classes intorduced here: https://github.com/runtimeverification/pyk/pull/311

h0nzZik:
You are right that AG is incorrect for what we do, we do AF. Imo AG means that the property holds on all paths in all states.

h0nzZik:
But actually, even CTL's AF has different semantics than all-path reachability, since it does not enforce the property on non-terminating executions.

ehildenb:
Maybe instead we should just call it `APRProver` and `APRProof`

h0nzZik:
APRProver sounds good to me. With regards the CTL*, that logic even does not know the concept of termination - only nonterminating runs are considered there. So even the weak future may be imprecise, or at least confusing.

h0nzZik:
Shall I create a PR with the renaming?

ehildenb:
Yes, please do.

h0nzZik:
Closed by https://github.com/runtimeverification/pyk/pull/382.

PR #382, opened 2023-04-20
rename AGProof to APRProof

h0nzZik:
(and AGProver to APRProver)
as per the issue https://github.com/runtimeverification/pyk/issues/355


src/pyk/proof/reachability.py
```diff
@@ -22,21 +22,28 @@
 _LOGGER: Final = logging.getLogger(__name__)
 
 
-class AGProof(Proof):
+class APRProof(Proof):
+    """APRProof and APRProver implement all-path reachability logic,
+    as introduced by A. Stefanescu and others in their 'All-Path Reachability Logic' paper:
```
tothtamas28, line 27:
```suggestion
    as introduced by A. Stefanescu and others in their paper 'All-Path Reachability Logic':
```

tothtamas28 (approved):

ehildenb:
We also should rename `AGBMCProof` to `APRBMCProof` instead.

PR #384, opened 2023-04-20
measure branching factor of KCFG

h0nzZik:
@ehildenb  Does `KCFG.branching_factor` look like it measures what we want to measure? In the current state it does not handle merging nodes specially, meaning that it would return `4` on the following double-diamond-like KCFG:
```
a -> b1
a -> b2
b1 -> c
b2 -> c
c -> d1
c -> d2
d1 -> e
d2 -> e
```
If you like it this way, I will also add the check to all other tests (`test_cell_map.py`).

TODO: add more tests


ehildenb:
I think we can make it even simpler, and we don't need to include it in the `kcfg` class (so that it's not an official measure, just a measure for testing purposes on the use of circularities).

So for a simpler measure, we can just measure: "how many nodes point into the target node + how mnay frontier nodes + how many stuck nodes". It's not really branching factor as much as "leaf number".

```
target_id = kcfg.get_unique_target().id
target_subsumed_nodes = len(kcfg.edges(target_id=target_id)) + len(kcfg.covers(target_id=target_id)) + len(kcfg.splits(target_id=target_id))
frontier_nodes = len(kcfg.frontier())
stuck_nodes = len(kcfg.stuck())
return target_subsumed_nodes + frontier_nodes + stuck_nodes
```

The reason I suggest this branching factor (and moving it into the test file directly rather than having it be an official measure of the KCFG), is that it's easy to compute and will allow you to test if the circularities are being use effectively to reduce branching factor.

ehildenb:
Remember here that the point is to test that a proof using a circularity about a branch does not have tthe branch in he overall proof. So maybe you can add the test of that in this PR as well.

h0nzZik:
That subproof-reusing functionality is not implemented yet, so I will create a test that asserts the higher number of leaves (2) instead of one and add a comment to it.

ehildenb (approved):

PR #389, opened 2023-04-21
Add tests for exercising the LLVM backend with Unicode input 

### LINKED ISSUE #385: Add test for throwing Kore with unicode to the LLVM backend
Currently, we are testing throwing unicode characters to:

- `kast`: frontend CLI tool, and
- `kore-rpc`: (tests skipped) for execution.

We should also test the LLVM backend.

tothtamas28:
Fixes #385 

Adds tests for using `krun` and the LLVM bindings.


src/tests/integration/test_string.py
```diff
@@ -128,33 +171,52 @@ def test_cli_rule_to_kast(definition_dir: Path, text: str) -> None:
...
+
+    # When
+    actual = krun.run_kore_term(kore)
+
+    # Then
+    assert actual == expected
 
+
+@pytest.mark.parametrize('text', TEST_DATA, ids=TEST_DATA)
+def test_bindings(runtime: ModuleType, text: str) -> None:
```
tothtamas28, line 188: Note that this test is not disabled, so things seem to break only when going through a textual KORE interface. (Or if the string is manipulated during rewrite.)

src/tests/integration/test_string.py
```diff
@@ -128,33 +171,52 @@ def test_cli_rule_to_kast(definition_dir: Path, text: str) -> None:
...
-                App("Lbl'-LT-'k'-GT-'", (), (kseq,)),
-                App("Lbl'-LT-'generatedCounter'-GT-'", (), (int_dv(0),)),
-                App("Lbl'-LT-'s'-GT-'", (), (s,)),
-            ),
-        )
+def test_krun(haskell_dir: Path, text: str) -> None:
+    # Given
+    kore = kore_config(text, '')
+    expected = kore_config(None, text)
+    krun = KRun(haskell_dir)
```
tothtamas28, line 178: If you want to see this break:

```suggestion
def test_krun(definition_dir: Path, text: str) -> None:
    # Given
    kore = kore_config(text, '')
    expected = kore_config(None, text)
    krun = KRun(definition_dir)
```

src/tests/integration/test_string.py
Scott-Guest: This behavior is expected as far as I can tell.

The Unicode issues with the LLVM backend that I'm aware of are
1. `\xhh` is treated as a raw byte `0xhh`, but Pyk, the Java frontend, and the Haskell backend treat it as `U+00hh`
2.  Unicode characters are UTF-8 encoded to be able to fit into a `char[]`, causing problems when
    1. manipulating strings (because the operations are not Unicode-aware)
    2. mixing Unicode characters with `\xhh` escapes

We never run into (1) or (2i) here because the escape sequences are only relevant during textual parsing / unparsing, but `llvm_to_kore` and `kore_to_llvm` just directly convert AST to AST (with pybind UTF-8 encoding and decoding at the boundary). We never run into 2ii because we don't manipulate the string.

src/tests/integration/test_string.py
Scott-Guest: Can we do something like
```
@pytest.mark.parametrize('text', TEST_DATA, ids=TEST_DATA)
def test_krun(definition_dir: Path, llvm_dir: Path, text: str) -> None:
    if definition_dir == llvm_dir:
        # https://github.com/runtimeverification/k/issues/3344
        pytest.skip()
```
to document that this is failing test which is intentionally skipped?

src/tests/integration/test_string.py
```diff
@@ -6,39 +6,82 @@
...
-KOMPILE_MAIN_FILE = 'k-files/string-rewrite.k'
-KOMPILE_MAIN_MODULE = 'STRING-REWRITE'
-
 TEST_DATA: Final = (
     'hello',
     '\n',
+    '\x01',
     '武天老師',
     '🙂',
 )
```
Scott-Guest, line 37: Let's add a few more test cases to ensure we hit [all of the UTF-8 encoding ranges](https://en.wikipedia.org/wiki/UTF-8#Encoding): £ (U+00A3) and 𐍈 (U+10348)

src/tests/integration/test_string.py
```diff
@@ -128,33 +171,52 @@ def test_cli_rule_to_kast(definition_dir: Path, text: str) -> None:
...
         pytest.skip()
 
     # Given
-    init = config(string_dv(text), string_dv(''))
+    kore = kore_config(text, '')
+    expected = kore_config(None, text)
 
     # When
-    with KoreServer(definition_dir, KOMPILE_MAIN_MODULE) as server:
+    with KoreServer(haskell_dir, KOMPILE_MAIN_MODULE) as server:
```
Scott-Guest, line 217: Might just be my lack of familiarity with `KoreServer`, but why exclude the `llvm_dir` here?

Scott-Guest (approved):
Left a few comments, but LGTM overall. Thank you for doing this!

src/tests/integration/test_string.py
tothtamas28: Done.

src/tests/integration/test_string.py
tothtamas28: `KoreServer` is a wrapper for the Haskell backend's [`kore-rpc`](https://github.com/runtimeverification/haskell-backend/blob/master/docs/2022-07-18-JSON-RPC-Server-API.md) tool.

PR #394, opened 2023-04-25
Node refutation and subproofs

geo2a:
https://github.com/runtimeverification/pyk/pull/396 Work towards [evm-semantics#1753](https://github.com/runtimeverification/evm-semantics/issues/1753)

Blocked on: #419 

Changes:

**Subproofs**

* `Proof` now holds a list of proof ids that are required to pass for the parent proof to pass. If a proof has subproofs, it must have a `proof_dir` set, and the subproofs are looked up there.
* factor-out `read_proof` to be a standalone function, rather than a `Proof` static method. This allows looking up the proof type in the JSON and returning the proof of the appropriate type
* Modifies the `*Proof.status` property to consider subproofs
* adds unit tests for the above logic

Subproofs are intended to serve as "proof obligations", i.e. auxiliary proofs that are factored-out of a larger ongoing proof.

All subproofs are "direct dependencies" of their parent proof, i.e. the parent proof cannot have *explicit* chains of subproofs. These chains could be implicit, i.e. the subproofs can have their own subproofs.

Related work is the `ProofSchedule` (PR #396) --- it is intended to make transitive dependencies explicit by organizing them into a DAG.

**Node refutation**

Having subproofs allows us to add a possibility to refute nodes in the `APRProver`. Refuting a node means (see `APRProver.refute_node`):

* marking it as expanded in the KCFG of the proof
* generating a `RefutationProof` that states the following: the conjunction of the node's conditions along the path AND the the condition of the first node down the split is unsatisfiable
* Add the generated proof as a subproof of the parent `APRProof`.
* If all stuck nodes of a proof haver a passing refutation, the parent proof passes.

*Note*: The node refutations proofs are proofs of the unsatisfiability of the node's path constraint. They have different meaning than `EqualityProof`'s, and are constructed in a slightly different way. I've decided to separate `RefutationProof` into its own class that does not inherit `EqualityProof` to make this distinction clear.


ehildenb:
In the KEVM case, we have to build actually a `Proof` DAG, because we have some contracts that call other contracts (eg. contract A calls contract B), and we want ot do contract B's proofs before contract A's so we can use them to speed up the calls into contract B.

geo2a:
@ehildenb this is ready for review. @h0nzZik could you have a look as well? I think it this functionality is somewhat related to the circularities PR (#395) and is somewhat logic-heavy. I'd appreciate both of you having a good look at this.

src/pyk/proof/reachability.py
```diff
@@ -45,35 +68,121 @@ def read_proof(id: str, proof_dir: Path) -> APRProof:
...
+        closest_branch = branches_on_path[0]
+        assert type(closest_branch) is KCFG.Split or type(closest_branch) is KCFG.NDBranch
+        if type(closest_branch) is KCFG.NDBranch:
+            _LOGGER.error(
+                f'Cannot refute node {shorten_hash(node.id)} following a non-determenistic branch: not yet implemented'
+            )
+            return None
+
+        assert type(closest_branch) is KCFG.Split
+        refuted_branch_root = closest_branch.targets[0]
```
h0nzZik, line 153: I do not understand why we are choosing `targets[0]`. What is special about the `0`-th target?

src/pyk/proof/reachability.py
```diff
@@ -45,35 +68,121 @@ def read_proof(id: str, proof_dir: Path) -> APRProof:
...
             f'    frontier: {len(self.kcfg.frontier)}',
             f'    stuck: {len(self.kcfg.stuck)}',
+            f'    refuted: {len(self.node_refutations.keys())}',
+            'Subproofs:' if len(self.subproof_ids) else '',
+        ]
+        for summary in subproofs_summaries:
+            yield from summary
+
+    def construct_node_refutation(self, node: KCFG.Node) -> EqualityProof | None:
+        """Construct an EqualityProof stating that the node's path condition is unsatisfiable"""
```
h0nzZik, line 133: Do in understand correctly that this method does not construct an actual proof, but only a proof obligation? Maybe we  should rename all the `*Proof` classes into `*Obligation`.

src/pyk/proof/reachability.py
```diff
@@ -214,6 +345,35 @@ def advance_proof(
...
+        refutation = self.proof.construct_node_refutation(node)
+        if refutation is None:
+            _LOGGER.error(f'Failed to refute node {shorten_hash(node.id)}')
+            return None
+        if extra_constraint is not None:
+            _LOGGER.info(f'Adding the provided extra cosntraint {extra_constraint} to refutation {refutation.id}')
+            refutation.add_constraint(extra_constraint)
+        refutation.write_proof()
+
+        # mark the node-to-refute as expanded to prevent further exploration
```
h0nzZik, line 359: We are marking it as expanded before the refutation happens, and it stays expanded even if the refutation does not succeed? Is that intentional?

src/pyk/proof/equality.py
```diff
@@ -145,14 +200,55 @@ class EqualityProver:
...
         self.proof = proof
 
-    def advance_proof(self, kcfg_explore: KCFGExplore) -> None:
-        if not self.proof.status == ProofStatus.PENDING:
+    def advance_proof(self, kcfg_explore: KCFGExplore, bind_consequent_variables: bool = True) -> None:
+        if self.proof.satisfiable is not None:
             return
 
-        kore = kcfg_explore.kprint.kast_to_kore(self.proof.equality)
+        dummy_config = kcfg_explore.kprint.definition.empty_config(sort=GENERATED_TOP_CELL)
```
h0nzZik, line 207: It might be also an option to use a list of free/used variables (`ListItem(X) ListItem(Y) ...`) instead of an empty configuration. Also, is there a reason why we assign to `dummy_config` so far before its first use?

src/pyk/proof/equality.py
```diff
@@ -16,67 +18,82 @@
...
     from ..kast.outer import KClaim, KDefinition
     from ..kcfg import KCFGExplore
+    from ..ktool.kprint import KPrint
 
     T = TypeVar('T', bound='Proof')
 
 _LOGGER: Final = logging.getLogger(__name__)
 
 
+class ProofInhabitation(Enum):
```
h0nzZik, line 31: What does it mean for a proof to be inhabited? In type theory, there is this notion of types being inhabited (and proofs being the inhabitants of the type) - is there some parallel?

geo2a: This turned out to be confusing, yes. What I meant is the satisfiability of the `EqualityProof`'s underlying conditional implication, as returned by the `check_implication` endpoint of `KoreClient`.

I've pushed changes that eliminate this class and refer to satisfiability instead.

src/pyk/proof/reachability.py
geo2a: Yes, this method constructs the obligation.

> Maybe we should rename all the `*Proof` classes into `*Obligation`

That would be the right thing to do, I guess. The current `*Proof` classes become actual proofs once we their status is `PASSED` and there's a KCFG on disk to check.

We can make that name change in a different PR.

src/pyk/proof/reachability.py
```diff
@@ -45,35 +68,121 @@ def read_proof(id: str, proof_dir: Path) -> APRProof:
...
+            yield from summary
+
+    def construct_node_refutation(self, node: KCFG.Node) -> EqualityProof | None:
+        """Construct an EqualityProof stating that the node's path condition is unsatisfiable"""
+        if not node in self.kcfg.nodes:
+            raise ValueError(f'No such node {node.id}')
+
+        # construct the path from the KCFG root to the node to refute
+        path = single(self.kcfg.paths_between(source_id=self.kcfg.get_unique_init().id, target_id=node.id))
+        # traverse the path back from the node-to-refute and filter-out split nodes and non-deterministic branches
```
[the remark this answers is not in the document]
geo2a: That is a spot-on, thanks! Will clarify.

src/pyk/proof/reachability.py
geo2a: Yes, this is intentional. The idea is to allow the user a "quick-and-dirty" way to bypass "uninteresting" branches in a proof. The intended workflow is:
* the user identifies a branch find in a `APRProof` that they think should be infeasible;
* the user calls `refute_node` on any node of this branch, which marks it expanded to prevent further exploration;
* The user not needs to discharge the generated `EqualityProof` obligation. They may continue working on the interesting branches of the proof.
* Once the user is convinced that the whole proof passes and the `EqualityProof` obligations of the branches they pruned actually hold, they need to add lemmas that would make ad-hoc pruning unnecessary.

Does this workflow makes sense to you? I'm hoping to be able to test it on KEVM soon.

src/pyk/proof/equality.py
geo2a: I've moved `dummy_config` closer to its use-site, thanks!

Regarding, the list of variables, it did not work. Let's stick to the empty config for now.  I'll make another attempt in a follow-up PR.

k-files/refute-node-spec.k
```diff
@@ -0,0 +1,7 @@
+requires "refute-node.k"
+
+module REFUTE-NODE-SPEC
+    imports REFUTE-NODE
+
+    claim [split-int-fail]: d(N) => e(N +Int 1)
```
ehildenb, line 6:
```suggestion
    claim [split-int-fail]: <k> d(N) => e(N +Int 1) ... </k>
```

src/pyk/proof/equality.py
```diff
@@ -87,32 +97,42 @@ def read_proof(id: str, proof_dir: Path) -> EqualityProof:
...
+            if self.rhs_body == FALSE:
+                return ProofStatus.FAILED
+            else:
+                return ProofStatus.PASSED
         else:
-            return ProofStatus.FAILED
+            if self.rhs_body == FALSE:
+                return ProofStatus.PASSED
+            else:
+                return ProofStatus.FAILED
```
ehildenb, line 117: I don't quite understand this logic? Why do we care about the RHS of the proof? We just want to show that the LHS implies the RHS in the matching-logic sense right?

src/pyk/proof/equality.py
```diff
@@ -124,18 +144,34 @@ def dict(self) -> dict[str, Any]:
...
+            [ f'LHS: {kprint.pretty_print(self.lhs_body)}',
+              f'RHS: {kprint.pretty_print(self.rhs_body)}',
+              f'Implication: {kprint.pretty_print(mlImplies(self.implication[0], self.implication[1]))}',]
+            + ([f'Simplified constraints: {kprint.pretty_print(self.simplified_constraints)}']
+               if self.simplified_constraints is not None else [])
+            + ([f'Implication satisfiable: {self.satisfiable}']
+               if self.satisfiable is not None else [])
+            + ([f'Implication predicate: {self.predicate}']
+               if self.predicate is not None else [])
+        )
```
ehildenb, line 167: I don't think we need special formatting here do we?

```
ret_lines = [
    f'LHS: ...',
    f'RHS: ...',
    f'Implication: ...',
]
if self.simplified_constraints is not None:
    ret_lines.append(f'Simplified constraints: ...')
...    
return ret_lines
```

Better to not have to turn the style checker off.

src/pyk/proof/equality.py
```diff
@@ -145,14 +181,55 @@ class EqualityProver:
     def __init__(self, proof: EqualityProof) -> None:
         self.proof = proof
 
-    def advance_proof(self, kcfg_explore: KCFGExplore) -> None:
-        if not self.proof.status == ProofStatus.PENDING:
+    def advance_proof(self, kcfg_explore: KCFGExplore, bind_consequent_variables: bool = True) -> None:
```
ehildenb, line 184: I don't think we need the `bind_consequent_variables` flag here, it's never needed that it is `False`, so let's just assume that it's `True` here.

Also, likely we can just change `KCFGExplore.cterm_implies` in the follownig ways and avoid the duplication:

- Always assume that we want to bind consequent variables (remove that option there).
- Return a tuple where the first item in the tuple is the `satisfiable` result, instead of just the `CSubst`.

ehildenb: We'll have to build the `dummy_config` here is this method directly, then pass in those ones to the `cterm_implies` endpoint.

ehildenb: And we should put a `#TODO` where we are building the dummy config, and make an issue about it on the haskell-backend board: "Why do we need to supply a config to the implication endpoint?"

src/pyk/proof/proof.py
```diff
@@ -44,6 +47,38 @@ def proof_exists(id: str, proof_dir: Path) -> bool:
...
+        from .utils import read_proof
+
+        if self.proof_dir is None and len(self.subproof_ids) > 0:
+            raise ValueError(f'Cannot read subproofs {self.subproof_ids} of proof {self.id} with no proof_dir')
+        if len(self.subproof_ids) == 0:
+            yield from ()
+        else:
+            assert self.proof_dir
+            for proof_id in self.subproof_ids:
+                yield read_proof(proof_id, self.proof_dir)
```
ehildenb, line 69: We should have this be:

```
@property
def subproofs(self) -> Iterable[Proof]:
   # Only read from disk if the subproof has changed somehow (inspect the hash of the subproof file)
   # Store the subproofs in a `_subproofs: dict[str, Proof]` field of this class.
```

To minimize Disk IO when manipulating subproofs.

src/pyk/proof/equality.py
geo2a: Makes sense, fixed.

src/pyk/proof/equality.py
geo2a: The interpretation of the implication check is something I can't quite nail down.

Supposing, we use the modified `KCFGExplore.cterm_implies` (see #416), and we receive either `None`, or the pair `(satisfiability, csubst)`. How do we determine if the proof passed? The questions we're asking by attempting an `EqualityProof` is whether `LHS == RHS` under the given constraints. How do we interpret the result of `cterm_implies` to answer this question?

ehildenb:
@geo2a let's get just the changes to `EqualityProof` pulled out into its own PR, that can probably be merged pretty quickly while we go back and forth on how to do subproofs.

geo2a:
@ehildenb I've factored-out the `EqualityProof` changes into #419

geo2a:
@ehildenb I think I'm happy with this PR now: the node refutations received a special `RefutationProof` class that encapsulates the unsatisfiability of node's path condition. I think it was a mistake to try and put both refutations and equality into the same class.

src/pyk/proof/equality.py
```diff
@@ -226,3 +295,46 @@ def advance_proof(self, kcfg_explore: KCFGExplore) -> None:
...
+            dummy_config = kcfg_explore.kprint.definition.empty_config(sort=GENERATED_TOP_CELL)
+            result = kcfg_explore.cterm_implies(
+                antecedent=CTerm(config=dummy_config, constraints=[]),
+                consequent=CTerm(config=dummy_config, constraints=[consequent_kast]),
+            )
+            if result is None:
+                _LOGGER.warning('cterm_implies returned None, the implication is unsatisfiable')
+            else:
+                self.proof.set_csubst(result)
+        self.proof.write_proof()
```
ehildenb, line 340: Hmmmm, seems like a decent amount of duplication of logic here. Is there any way to make this inherit from `EqualityProver` and use `super().advance_proof(...)`? Also, can you summarize for me why we are using a different class entirely than the `EqualityProof`? Is it just more specific because we know specifically that the consequent will always be `FALSE`?

src/pyk/proof/proof.py
```diff
@@ -25,25 +26,105 @@ class ProofStatus(Enum):
...
+            raise ValueError(f'Cannot check if proof {self.id} with no proof_dir is up-to-date')
+        proof_path = self.proof_dir / f'{hash_str(id)}.json'
+        if proof_path.exists() and proof_path.is_file():
+            match check_method:
+                case 'checksum':
+                    return self.checksum == hash_file(proof_path)
+                case _:  # timestamp
+                    return self._last_modified == proof_path.stat().st_mtime_ns
+        else:
+            return False
```
ehildenb, line 84: Let's just default to comparing checksums here. No need to cause ambiguity, and timestamps can be tricky.

src/pyk/proof/equality.py
geo2a: After trying to integrate this work into KEVM, I'm thinking more that `check_implication` is not what we want.

What we what is a much simple thing: "given a conjunction, tell me if its satisfiable or not". That is all what we need to know to refute nodes, i.e. we do not need the `#Implies` connective.

In the first place, we've tried using `check_implication` because we wanted satifiability information, but it actually requires us to deal with a lot of unnecessary technicalities, like building the actual implication.

Today, I done some experiments building the implications I'd think should equisatisfiable with the initial conjunction (the path condition we're checking), but either my understanding is flawed, or the backend is not doing what I want. For example:

```
GENERATED_TOP_CELL:GeneratedTopCell #Implies
#Exists Y:Int . #Exists X:Int .
     ( { true #Equals 0 <Int X:Int }
#And ( { true #Equals 0 <Int Y:Int }
#And ( { true #Equals X:Int <=Int 100 }
#And ( { true #Equals Y:Int <=Int 100 }
#And { true #Equals Y:Int <=Int X:Int } ) ) ) ) )
```

is unsatisfiable, while I'd expect it to be. Now, if I remove the constraints in `X` and `Y` from the consequent:

```
( GENERATED_TOP_CELL:GeneratedTopCell #Implies
#Exists Y:Int . #Exists X:Int . { true #Equals Y:Int <=Int X:Int } )
```

The implication becomes satisfiable.

Additionally, in both cases `chech_implication` returns

```
predicate: Top(sort=SortApp(name='SortGeneratedTopCell', sorts=()))
substitution: Top(sort=SortApp(name='SortGeneratedTopCell', sorts=()))
```

which, as I understand, corresponds to the ["stuck" case](https://github.com/runtimeverification/haskell-backend/blob/7f51e8c9038f7fa20e90f7886503683456866d2a/docs/2022-07-18-JSON-RPC-Server-API.md?plain=1#L355) described in Kore RPC spec.

The backend has the necessary code to check satisfiability of first-order logic formulas. Perhaps we could consider exposing it via RPC?

src/pyk/proof/proof.py
nwatson22: The check method now defaults to checksums.

src/pyk/kcfg/kcfg.py
```diff
@@ -177,7 +176,7 @@ def with_single_target(self, target: KCFG.Node) -> KCFG.NDBranch:
     _lock: RLock
 
     def __init__(self) -> None:
-        self._node_id = count(1)
+        self._node_id = 1
```
ehildenb, line 179: Why are we getting rid of the counter usage?

src/pyk/proof/proof.py
```diff
@@ -59,9 +140,23 @@ def dict(self) -> dict[str, Any]:
...
+        raise ValueError(f'Could not load Proof from file {id}: {proof_path}')
+
     @property
     def summary(self) -> Iterable[str]:
-        return [
-            f'Proof: {self.id}',
-            f'    status: {self.status}',
-        ]
+        subproofs_summaries = [subproof.summary for subproof in self.subproofs]
+        return chain([f'Proof: {self.id}', f'    status: {self.status}'], *subproofs_summaries)
```
ehildenb, line 162: Maybe we should indent the subproof summaries 2 spaces or something?

src/pyk/proof/equality.py
```diff
@@ -173,6 +162,89 @@ def summary(self) -> Iterable[str]:
...
+            simplified_constraints=simplified_constraints,
+            proof_dir=proof_dir,
+        )
+
+    @property
+    def summary(self) -> Iterable[str]:
+        return [
+            f'RefutationProof: {self.id}',
+            f'  status: {self.status}',
+        ]
```
ehildenb, line 236:
```suggestion
    @property
    def summary(self) -> Iterable[str]:
        return [
            f'RefutationProof: {self.id}',
            f'    status: {self.status}',
        ]
```

Are we using 2 spaces or 4 on other proof summaries?

src/pyk/kcfg/kcfg.py
ehildenb: Let's pull this counter change out into it's own bugfix PR, so that @tothtamas28 can review it separately and quickly.

src/pyk/proof/proof.py
```diff
@@ -59,9 +140,23 @@ def dict(self) -> dict[str, Any]:
...
+        proof_path = proof_dir / f'{hash_str(id)}.json'
+        if Proof.proof_exists(id, proof_dir):
+            proof_dict = json.loads(proof_path.read_text())
+            proof_type = proof_dict['type']
+            _LOGGER.info(f'Reading {proof_type} from file {id}: {proof_path}')
+            if proof_type in Proof._PROOF_TYPES:
+                return locals()[proof_type].from_dict(proof_dict, proof_dir)
+
+        raise ValueError(f'Could not load Proof from file {id}: {proof_path}')
+
```
ehildenb, line 158: Let's pull out `Proof.is_uptodate`, `Proof.checksum` (which we can call `Proof.digest` for consistency), and `Proof.read_proof` out into its own PR.

src/pyk/kcfg/kcfg.py
nwatson22: Now pulled into #484.

src/pyk/proof/proof.py
nwatson22: Pulled into #485.

src/pyk/proof/equality.py
nwatson22: It's now consistently 4 spaces for all `summary`s.

src/pyk/proof/proof.py
nwatson22: It's now consistently 4 spaces for all `summary`s.

src/pyk/proof/reachability.py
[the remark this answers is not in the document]
nwatson22: Changed it to "traverse the path back from the node-to-refute and choose only split nodes and non-deterministic branches".

ehildenb:
To get the kore terms, you can do `make test-integration TEST_ARGS='-k MY_TEST_SELECTOR --log-file log --log-file-level debug'`

ehildenb:
Create issue with weird behavior of the `implies` endpoint against the haskell-backend repo, and print a big warning when building the `EqualityProof` that the user should expect unsound behavior.

src/pyk/proof/equality.py
```diff
@@ -169,9 +158,92 @@ def pretty(self, kprint: KPrint) -> Iterable[str]:
...
+
+    def set_csubst(self, csubst: CSubst) -> None:
+        self.csubst = csubst
+
+    @property
+    def status(self) -> ProofStatus:
+        if self.simplified_constraints is None:
+            return ProofStatus.PENDING
+        elif self.csubst is not None:
+            return ProofStatus.FAILED
```
ehildenb, line 195: Why is it failing if `self.csubst is not None`? Shouldn't it be failing if `self.csubst is None`, the opposite?

src/pyk/proof/reachability.py
```diff
@@ -146,13 +196,77 @@ def add_terminal(self, nid: NodeIdLike) -> None:
...
+        except ValueError:
+            _LOGGER.error(f'Node {node.id} is not reachable from the initial node.')
+            return None
+        # traverse the path back from the node-to-refute and choose only split nodes and non-deterministic branches
+        branches_on_path = list(filter(lambda x: type(x) is KCFG.Split or type(x) is KCFG.NDBranch, reversed(path)))
+        if len(branches_on_path) == 0:
+            _LOGGER.error(f'Cannot refute node {node.id} in linear KCFG')
+            return None
+        closest_branch = branches_on_path[0]
+        assert type(closest_branch) is KCFG.Split or type(closest_branch) is KCFG.NDBranch
```
ehildenb, line 234: Is this assertion needed? We're asserting that it's `KCFG.Split` later, and already checking if it's `KCFG.NDBranch` later too.

src/pyk/proof/equality.py
```diff
@@ -218,3 +290,49 @@ def advance_proof(self, kcfg_explore: KCFGExplore) -> None:
...
+                antecedent=CTerm(config=dummy_config, constraints=[]),
+                consequent=CTerm(config=dummy_config, constraints=[consequent_kast]),
+            )
+            if result is None:
+                _LOGGER.warning('cterm_implies returned None, the implication is unsatisfiable')
+                _LOGGER.warning(
+                    'Refutation is possibly falsely indicated as correct. See https://github.com/runtimeverification/haskell-backend/issues/3605.'
+                )
+            else:
+                self.proof.set_csubst(result)
```
ehildenb, line 337: Look at how `EqualityProver` has been refactored here: https://github.com/runtimeverification/pyk/pull/469/files#diff-933465707ef1f3131367afb6fec30121ee670bdedc5ac1f97f942f79832b7e2aR181, and make sure we're trying to match that logic as closely as possible. This PR was introduced before the changes there were merged.

ehildenb: In particular, it re-orders the way we do the computation, and doesn't store an intermediate for `proof_failed_trivially`, just inspects the simplified antecedent/consequent.

src/pyk/proof/equality.py
```diff
@@ -218,3 +290,49 @@ def advance_proof(self, kcfg_explore: KCFGExplore) -> None:
...
+        _LOGGER.info(f'Attempting RefutationProof {self.proof.id}')
+
+        proof_failed_trivially = False
+        consequent_simplified_kore, _ = kore_client.simplify(kcfg_explore.kprint.kast_to_kore(consequent_kast))
+        consequent_simplified_kast = kcfg_explore.kprint.kore_to_kast(consequent_simplified_kore)
+        _LOGGER.info(f'Simplified consequent: {kcfg_explore.kprint.pretty_print(consequent_simplified_kast)}')
+        self.proof.set_simplified_constraints(consequent_simplified_kast)
+        if is_top(consequent_simplified_kast):
+            _LOGGER.warning(
+                'Consequent of implication (proof equality) simplifies to #Top. The constraitns are satisfiable, the implication will not be checked.'
```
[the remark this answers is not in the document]
nwatson22: Fixed. It was not also in EqualityProver.

src/pyk/proof/equality.py
ehildenb: Also, we should be emitting the warning about incorrect refutation _every_ time an `EqualityProof` is constructed, or if a `RefutationProof` is constructed. In the constructors of `EqualityProof` and `RefutationProof`, we should have the warning say something like "Building a *Proof that has known soundness issues: See ...".

src/pyk/proof/proof.py
```diff
@@ -25,14 +26,25 @@ class ProofStatus(Enum):
...
         self.id = id
         self.proof_dir = proof_dir
+        self.subproof_ids = subproof_ids if subproof_ids is not None else []
+        self._subproofs = {}
+        if self.proof_dir is None and len(self.subproof_ids) > 0:
+            raise ValueError(f'Cannot read subproofs {self.subproof_ids} of proof {self.id} with no proof_dir')
+        if len(self.subproof_ids) > 0:
+            assert self.proof_dir
+            for proof_id in self.subproof_ids:
+                self.fetch_subproof(proof_id, force_reread=True)
```
ehildenb, line 48: Let's pull out the subproof work into its own PR, with a few changes:

Implement a default `dict` method for `Proof` class, instead of leaving it abstract. `dict` should build the dictionary which contains things like `{'id': self.id, 'subproofs': self._subproof_ids}`. Then inheriting classes can do (for example, for `APRProof`):

```
class Proof:
    @property
        return {'id': self.id, 'subproofs': self._subproof_ids}

class APRProof:
    @property
        dct = super().dict
        dct['type'] = str(self.type)
        dct['init'] = self.init
```

src/pyk/proof/equality.py
nwatson22: This is refactored to be more consistent with the other one and the warning is moved to the constructor of both EqualityProof and RefutationProof.

src/pyk/proof/proof.py
nwatson22: Pulled out into #500, now merged.

src/pyk/proof/equality.py
```diff
@@ -218,3 +300,42 @@ def advance_proof(self, kcfg_explore: KCFGExplore) -> None:
...
+        _LOGGER.info(f'Attempting RefutationProof {self.proof.id}')
+
+        if self.proof.status is not ProofStatus.PENDING:
+            _LOGGER.info(f'RefutationProof finished {self.proof.id}: {self.proof.status}')
+            return
+
+        antecedent_kast = mlTop()
+        consequent_kast = self.proof.constraint
+
+        antecedent_simplified_kast, _ = kcfg_explore.kast_simplify(antecedent_kast)
```
ehildenb, line 321: I guess we can skip this call, since `antecedent_kast` is just `mlTop()`, and it will avoid a round-trip to the backend.

src/pyk/proof/equality.py
```diff
@@ -218,3 +300,42 @@ def advance_proof(self, kcfg_explore: KCFGExplore) -> None:
...
+        consequent_simplified_kast, _ = kcfg_explore.kast_simplify(consequent_kast)
+        self.proof.set_simplified_constraints(consequent_simplified_kast)
+        _LOGGER.info(f'Simplified antecedent: {kcfg_explore.kprint.pretty_print(antecedent_simplified_kast)}')
+        _LOGGER.info(f'Simplified consequent: {kcfg_explore.kprint.pretty_print(consequent_simplified_kast)}')
```
ehildenb, line 325: We probably also don't need to log the simplified antecedent, and instead of calling the other one the `consequent`, we can just call it the `constraints` perhaps?

src/pyk/proof/proof.py
```diff
@@ -26,19 +26,20 @@ class ProofStatus(Enum):
...
         self._subproofs = {}
-        if self.proof_dir is None and len(list(subproof_ids)) > 0:
-            raise ValueError(f'Cannot read subproofs {subproof_ids} of proof {self.id} with no proof_dir')
-        if len(list(subproof_ids)) > 0:
+        if self.proof_dir is None and len(self.subproof_ids) > 0:
+            raise ValueError(f'Cannot read subproofs {self.subproof_ids} of proof {self.id} with no proof_dir')
+        if subproof_ids is not None and len(list(subproof_ids)) > 0:
+            assert self.proof_dir
             for proof_id in subproof_ids:
                 self.fetch_subproof(proof_id, force_reread=True)
```
ehildenb, line 44: Let's revert these changes, I don't think they're necessary, and are probably holdover from before #500 was pulled out.

src/pyk/proof/reachability.py
```diff
@@ -395,6 +498,42 @@ def advance_proof(
...
+            _LOGGER.error(f'Failed to refute node {node.id}')
+            return None
+        if extra_constraint is not None:
+            _LOGGER.info(f'Adding the provided extra cosntraint {extra_constraint} to refutation {refutation.id}')
+            refutation.add_constraint(extra_constraint)
+        refutation.write_proof()
+
+        # mark the node-to-refute as expanded to prevent further exploration
+        #          self.proof.kcfg.add_expanded(node.id)
+        self.proof.add_terminal(node.id)
```
ehildenb, line 514: Rather than tracking the refuted nodes as `terminal`, we should just have a separate set (or even better, a dictionary), and if the node is in that dictionary, then it's refuted. That way we can track `terminal` and `refuted` separately.

src/pyk/proof/reachability.py
```diff
@@ -50,8 +53,9 @@ def __init__(
         init: NodeIdLike,
         target: NodeIdLike,
         logs: dict[int, tuple[LogEntry, ...]],
-        terminal_nodes: Iterable[NodeIdLike] | None = None,
         proof_dir: Path | None = None,
+        node_refutations: dict[int, str] | None = None,
```
ehildenb, line 57: Could just have this be `dict[NodeIdLike, RefutationProof]` instead? Then to check if a node is refuted, we check if that node is in the `node_refutations` dirtionary? It would require having the keys actually be resolved, so we can do direct comparison.

src/pyk/proof/equality.py
```diff
@@ -218,3 +300,42 @@ def advance_proof(self, kcfg_explore: KCFGExplore) -> None:
...
         self.proof.write_proof()
+
+
+class RefutationProver:
+    proof: RefutationProof
+
+    def __init__(self, proof: RefutationProof) -> None:
+        self.proof = proof
+
+    def advance_proof(self, kcfg_explore: KCFGExplore) -> None:
```
ehildenb, line 311: Right now, it seems we're building `#Top #Implies PATH_CONSTRAINT`, and then checking that it goes to `#Bottom`. But instead we could just simplify `PATH_CONSTRAINT` directly (no call to implication checker) via the `cterm_simplify` endpoint. Any reason not to do that instead?

src/pyk/proof/reachability.py
```diff
@@ -101,7 +139,9 @@ def read_proof(id: str, proof_dir: Path) -> APRProof:
 
     @property
     def status(self) -> ProofStatus:
-        if len(self.failing) > 0 or self.subproofs_status == ProofStatus.FAILED:
+        all_stuck_refuted = self.stuck_nodes_refuted
+
+        if (len(self.failing) > 0 and not all_stuck_refuted) or self.subproofs_status == ProofStatus.FAILED:
             return ProofStatus.FAILED
         elif len(self.pending) > 0 or self.subproofs_status == ProofStatus.PENDING:
             return ProofStatus.PENDING
```
ehildenb, line 147: The `APRProof.status` definition should not change. Instead, have `APRProof.is_failing` take into account if `APRProof.is_refuted`.

src/pyk/proof/reachability.py
```diff
@@ -73,6 +91,26 @@ def pending(self) -> list[KCFG.Node]:
...
+    #              for node_id, proof_id in self.node_refutations.items():
+    #                  yield node_id, cast('RefutationProof', self.fetch_subproof(proof_id))
+
+    @property
+    def stuck_nodes_refuted(self) -> bool:
+        stuck_nodes = self.kcfg.stuck + self.terminal
+        return all(self.is_refuted(node.id) for node in stuck_nodes)
+
+    #          refutations = dict(self.read_refutations())
+    #          return all(n.id in self.node_refutations.keys() and refutations[n.id].csubst is None for n in stuck_nodes)
```
ehildenb, line 109: We won't need this if we just adjust `APRProof.is_failing` and `APRProof.is_pending` to use the new `APRProof.is_refuted` directly. Then `APRProof.status` doesn't change, and so we don't need this predicate anymore.

src/pyk/proof/reachability.py
```diff
@@ -158,20 +200,32 @@ def dict(self) -> dict[str, Any]:
...
         logs = {k: [l.to_dict() for l in ls] for k, ls in self.logs.items()}
         dct['logs'] = logs
         return dct
 
+    def add_refuted(self, nid: NodeIdLike, proof: RefutationProof) -> None:
+        self.node_refutations[self.kcfg._resolve(nid)] = proof
+
+    def remove_refuted(self, nid: NodeIdLike) -> None:
+        del self.node_refutations[self.kcfg._resolve(nid)]
+
```
ehildenb, line 214: I think `node_refutations` should be an internal variable perhaps, `_node_refutations`.

We perhaps don't need `add_refuted` and `remove_refuted`, just in `APRProver` add method `APRProver.refute_node(node_id)`, and `APRProve.unrefute_node(node_id)`.

src/pyk/proof/reachability.py
```diff
@@ -395,6 +436,90 @@ def advance_proof(
...
 
+    def refute_node(self, kcfg_explore: KCFGExplore, node: KCFG.Node, extra_constraint: KInner | None = None) -> None:
+        _LOGGER.info(f'Attempting to refute node {node.id}')
+        refutation = self.construct_node_refutation(node)
+        if refutation is None:
+            _LOGGER.error(f'Failed to refute node {node.id}')
+            return None
+        if extra_constraint is not None:
+            _LOGGER.info(f'Adding the provided extra cosntraint {extra_constraint} to refutation {refutation.id}')
+            refutation.add_constraint(extra_constraint)
```
ehildenb, line 447: What's the idea of how `add_constraint` will be used?

src/pyk/proof/reachability.py
```diff
@@ -395,6 +436,90 @@ def advance_proof(
...
+        refutation.write_proof()
+
+        self.proof.node_refutations[node.id] = refutation
+
+        self.proof.write_proof()
+
+        eq_prover = RefutationProver(refutation)
+        eq_prover.advance_proof(kcfg_explore)
```
ehildenb, line 455: I think we don't want to immediately try and discharge the refutation proof, because it will likely fail (otherwise that branch wouldn't be present in the original APRProof).

Is it necessary to do so?

src/pyk/proof/reachability.py
nwatson22: I don't know. @geo2a, what was the original purpose of this?

src/pyk/proof/reachability.py
```diff
@@ -395,6 +436,90 @@ def advance_proof(
...
+        if not node in self.proof.kcfg.nodes:
+            raise ValueError(f'No such node {node.id}')
+
+        # construct the path from the KCFG root to the node to refute
+        try:
+            path = single(self.proof.kcfg.paths_between(source_id=self.proof.init, target_id=node.id))
+        except ValueError:
+            _LOGGER.error(f'Node {node.id} is not reachable from the initial node.')
+            return None
+        # traverse the path back from the node-to-refute and choose only split nodes and non-deterministic branches
```
ehildenb, line 479:
```suggestion
        path = single(self.proof.kcfg.paths_between(source_id=self.proof.init, target_id=node.id))
```

ehildenb: I think all the error handling is not needed. These conditions should not fail, and if they do, the stack trace of the call to `paths_between` and `single` should be informative enough.

src/pyk/proof/reachability.py
ehildenb: I think let's just remove it for now.

src/pyk/proof/reachability.py
```diff
@@ -395,6 +436,80 @@ def advance_proof(
...
+        last_constraint = mlEquals(TRUE, ml_pred_to_bool(csubst.constraints[0]), arg_sort=BOOL)
+
+        refutation_id = self.proof.get_refutation_id(node.id)
+        _LOGGER.info(f'Adding refutation proof {refutation_id} as subproof of {self.proof.id}')
+        refutation = RefutationProof(
+            id=refutation_id,
+            sort=BOOL,
+            constraints=[*pre_split_constraints, last_constraint],
+            proof_dir=self.proof.proof_dir,
+        )
```
ehildenb, line 508: Hmmmm, looking at this, I think we should basically be constructing:

```
rule last_constraint => false requires pre_split_constraints
```

Which would be:

```
{ true #Equals pre_split_constraints } #Implies { false #Equals last_constraint }
```

Instead, we're constructing:

```
#Top #Implies (last_consrtaint #And pre_split_constraints)
```

Have we tried the first approach? It seems cleaner and more direct to me, and stores the pre-split constraints separately in the `RefutationProof` from the split constraint itself. In the current way, we can't seem to be able to tell from the `RefutationProof` which was the last constraint in the path, so we're losing that information.

src/pyk/proof/reachability.py
nwatson22: I removed this logic. Now the user now has to call `advance_proof` on the RefutationProof to actually attempt the refutation.

src/pyk/proof/reachability.py
nwatson22: It's removed. It was sort of used in testing, where these extra constraints were used to make the refutation fail or pass depending on what was being passed in.

src/pyk/proof/reachability.py
nwatson22: I changed it so the `RefutationProof` now tracks its `pre_split_constraints` and `last_constraint` separately and gives the `pre_split_constraints` as the antecedent and the negated `last_constraint` as the consequent. Is this what you meant?

src/pyk/proof/equality.py
```diff
@@ -169,10 +161,112 @@ def pretty(self, kprint: KPrint) -> Iterable[str]:
...
+            'id': self.id,
+            'sort': self.sort.to_dict(),
+            'pre_constraints': [c.to_dict() for c in self.pre_constraints],
+            'last_constraint': self.last_constraint.to_dict(),
+        }
+        if self.simplified_constraints is not None:
+            dct['simplified_constraints'] = self.simplified_constraints.to_dict()
+        if self.csubst is not None:
+            dct['csubst'] = self.csubst.to_dict()
+        return dct
```
ehildenb, line 230:
```suggestion
    @property
    def dict(self) -> dict[str, Any]:
        dct = super().dict
        dct['type'] = 'RefutationProof'
        dct['sort'] = self.sort.to_dict()
        dct['pre_constraints'] = [c.to_dict() for c in self.pre_constraints]
        dct['last_constraint'] = self.last_constraint.to_dict()
        if self.simplified_constraints is not None:
            dct['simplified_constraints'] = self.simplified_constraints.to_dict()
        if self.csubst is not None:
            dct['csubst'] = self.csubst.to_dict()
        return dct
```

src/pyk/proof/equality.py
```diff
@@ -218,3 +312,41 @@ def advance_proof(self, kcfg_explore: KCFGExplore) -> None:
...
+        consequent_kast = mlEqualsFalse(self.proof.last_constraint)
+
+        antecedent_simplified_kast, _ = kcfg_explore.kast_simplify(antecedent_kast)
+        consequent_simplified_kast, _ = kcfg_explore.kast_simplify(consequent_kast)
+        self.proof.set_simplified_constraints(consequent_simplified_kast)
+        _LOGGER.info(f'Simplified constraints: {kcfg_explore.kprint.pretty_print(consequent_simplified_kast)}')
+
+        if is_top(consequent_simplified_kast):
+            _LOGGER.warning(f'Consequent of implication (proof equality) simplifies to #Top {self.proof.id}')
+            self.proof.set_csubst(CSubst(Subst({}), ()))
```
ehildenb, line 340:
```suggestion
        if is_top(consequent_simplified_kast):
            _LOGGER.warning(f'Consequent of implication (proof equality) simplifies to #Top {self.proof.id}')
            self.proof.set_csubst(CSubst(Subst({}), ()))
        if is_bottom(antecedent_simplified_kast):
            _LOGGER.warning(f'Antecedent of implication (proof equality) simplifies to #Top {self.proof.id}')
            self.proof.set_csubst(CSubst(Subst({}), ()))
```

Now, it looks quite a bit exactly the same as `EqualityProver`? Perhaps we should be inheriting from that class now?

ehildenb: We can worry about that later.

src/pyk/proof/equality.py
```diff
@@ -169,10 +161,110 @@ def pretty(self, kprint: KPrint) -> Iterable[str]:
...
+        _LOGGER.warning(
+            'Building a RefutationProof that has known soundness issues: See https://github.com/runtimeverification/haskell-backend/issues/3605.'
+        )
+
+    #      @property
+    #      def constraint(self) -> KInner:
+    #          return mlAnd(self.constraints)
+
+    #      def add_constraint(self, new_constraint: KInner) -> None:
+    #          self.constraints = (*self.constraints, new_constraint)
```
nwatson22, line 200:
```suggestion
```

src/pyk/proof/reachability.py
```diff
@@ -116,6 +142,7 @@ def from_dict(cls: type[APRProof], dct: Mapping[str, Any], proof_dir: Path | Non
         target_node = dct['target']
         id = dct['id']
         subproof_ids = dct['subproof_ids'] if 'subproof_ids' in dct else []
+        node_refutations = keys_to_int(dct['node_refutations']) if 'node_refutations' in dct else {}
```
ehildenb, line 145: Maybe we should be using `self.kcfg._resolve(...)` on each key read in, to make sure  we're storing them in resolved form so the predicate `is_refuted` will work properly?

We can also then avoid having/using th emethod `keys_to_int`.

src/pyk/proof/reachability.py
```diff
@@ -158,12 +186,17 @@ def dict(self) -> dict[str, Any]:
         dct['init'] = self.init
         dct['target'] = self.target
         dct['terminal_nodes'] = self._terminal_nodes
+        _node_refutations = {node_id: proof.id for (node_id, proof) in self.node_refutations.items()}
+        dct['node_refutations'] = _node_refutations
```
ehildenb, line 190: Any reason to have this on two lines?

src/pyk/proof/reachability.py
```diff
@@ -240,8 +278,9 @@ def from_dict(cls: type[APRBMCProof], dct: Mapping[str, Any], proof_dir: Path |
         target = dct['target']
         bounded_nodes = dct['bounded_nodes']
         bmc_depth = dct['bmc_depth']
-        id = dct['id']
         subproof_ids = dct['subproof_ids'] if 'subproof_ids' in dct else []
+        node_refutations = keys_to_int(dct['node_refutations']) if 'node_refutations' in dct else {}
+        id = dct['id']
```
ehildenb, line 283: Same here, maybe we can use `cfg._resolve(...)` instead of `keys_to_int`?

src/pyk/proof/reachability.py
nwatson22:
```suggestion
        dct['node_refutations'] = {node_id: proof.id for (node_id, proof) in self.node_refutations.items()}
```

src/pyk/proof/reachability.py
```diff
@@ -283,6 +323,7 @@ def summary(self) -> Iterable[str]:
             f'APRBMCProof(depth={self.bmc_depth}): {self.id}',
             f'    status: {self.status}',
             f'    nodes: {len(self.kcfg.nodes)}',
+            f'    refuted: {len(self.node_refutations.keys())}',
```
ehildenb, line 326: Let's put `refuted` after `terminal`. Basically the earlier items in the list are more generic, the later ones are more specific.

src/pyk/proof/reachability.py
nwatson22: Done.

src/pyk/proof/reachability.py
nwatson22: Done.

src/pyk/proof/reachability.py
nwatson22: Done.

src/pyk/proof/reachability.py
ehildenb: Yeah, looks better.

src/tests/integration/kcfg/test_refute_node.py
```diff
@@ -0,0 +1,191 @@
...
+    #      ),
+    #      (
+    #          'refute-node-success-range-N',
+    #          mlEqualsTrue(gtInt(KVariable('N'), intToken(10))),
+    #          (
+    #              mlEqualsTrue(leInt(KVariable('N'), intToken(0))),
+    #              mlEqualsTrue(gtInt(KVariable('N'), intToken(10))),
+    #          ),
+    #          ProofStatus.PASSED,
+    #      ),
```
ehildenb, line 51: Why are these commented out? Are they failing? We should use `pytest.skip()` for them then.

nwatson22: Removed. They were commented since we removed `extra_constraints` for now and that was the only thing varying between the tests.

PR #395, opened 2023-04-26
Support Circularities in `APRProver`

h0nzZik:
This pull request enables support for using dependencies during a proof search done by `APRProver`, as well as support for circularities. The class `APRProof` now takes a list of dependencies of a claim (other than the claim itself), plus a flag saying whether the claim is a circularity - that is, whether it depends on itself. The class `APRProver` then performs all symbolic execution in a newly-added module (with name suffix `-DEPENDS-MODULE`) containing the dependencies of the claim excluding the claim itself; if the claim is a circularity, then all symbolic execution after one step is performed in another module (with name suffix `-CIRCULARITIES-MODULE`) containing all the dependencies AND the claim itself.


src/pyk/kcfg/explore.py
```diff
@@ -313,3 +330,20 @@ def extend(
 
         else:
             raise ValueError('Unhandled case.')
+
+    def add_circularities_module(
+        self, old_module_name: str, new_module_name: str, circularities: Iterable[KClaim]
+    ) -> None:
+        max_priority: int = 1  # Maybe we should compute this dynamically?
```
ehildenb, line 337: Yes, I would pass it in as an argument to this method, and then let the call site compute it by inspecting th edefinition.

src/pyk/kcfg/explore.py
```diff
@@ -313,3 +330,20 @@ def extend(
...
+        kast_rules = [
+            KRule(body=c.body, requires=c.requires, ensures=c.ensures, att=KAtt({'priority': max_priority}))
+            for c in circularities
+        ]
```
ehildenb, line 341: Could we do instead:

```suggestion
        kast_rules = [c.let(att=c.att.update({'priority': max_priority})) for c in circularities]
```

h0nzZik: We need also to change the type of the circularity from `KClaim` to `KRule`, but otherwise, we can preserve the attributes as you suggested.

ehildenb:
@h0nzZik please add a description/summary of changes to the main PR body.

src/pyk/kcfg/explore.py
```diff
@@ -334,3 +351,19 @@ def extend(
...
+        kore_axioms: list[Sentence] = [krule_to_kore(self.kprint.kompiled_kore, r) for r in kast_rules]
+        _, kore_client = self._kore_rpc
+        sentences: list[Sentence] = [Import(module_name=old_module_name, attrs=())]
+        sentences = sentences + kore_axioms
+        m = Module(name=new_module_name, sentences=sentences)
+        _LOGGER.info(f'Adding module {m.text}')
+        kore_client.add_module(m)
+        return None
```
ehildenb, line 369:
```suggestion
```

src/pyk/kcfg/explore.py
```diff
@@ -334,3 +351,19 @@ def extend(
...
+        kast_rules = [
+            KRule(body=c.body, requires=c.requires, ensures=c.ensures, att=c.att.update({'priority': priority}))
+            for c in circularities
+        ]
+        kore_axioms: list[Sentence] = [krule_to_kore(self.kprint.kompiled_kore, r) for r in kast_rules]
+        _, kore_client = self._kore_rpc
+        sentences: list[Sentence] = [Import(module_name=old_module_name, attrs=())]
+        sentences = sentences + kore_axioms
+        m = Module(name=new_module_name, sentences=sentences)
+        _LOGGER.info(f'Adding module {m.text}')
```
ehildenb, line 367:
```suggestion
        _LOGGER.info(f'Adding circularities module {self.id}: {new_module_name}')
```

src/pyk/konvert.py
```diff
@@ -93,13 +93,18 @@ def kast_to_kore(
...
 
-    kore_lhs = _kast_to_kore(krule_lhs.kast)
-    kore_rhs = _kast_to_kore(krule_rhs.kast)
+    kore_lhs = kompiled_kore.add_injections(
+        _kast_to_kore(krule_lhs.kast), sort=SortApp(name='SortGeneratedTopCell', sorts=())
+    )
+    kore_rhs = kompiled_kore.add_injections(
+        _kast_to_kore(krule_rhs.kast), sort=SortApp(name='SortGeneratedTopCell', sorts=())
+    )
+    #
```
ehildenb, line 107:
```suggestion
```

src/tests/integration/kcfg/test_imp.py
```diff
@@ -580,15 +628,18 @@ def test_all_path_reachability_prove(
             kprove.get_claims(Path(spec_file), spec_module_name=spec_module, claim_labels=[f'{spec_module}.{claim_id}'])
         )
 
+        circularities = kprove.get_circularities(Path(spec_file), spec_module_name=spec_module)
+        _LOGGER.info(f'We have {len(circularities)} circularities')
         kcfg = KCFG.from_claim(kprove.definition, claim)
-        proof = APRProof(f'{spec_module}.{claim_id}', kcfg)
+        proof = APRProof(f'{spec_module}.{claim_id}', kcfg, circularities=circularities, uuid=claim.att['UNIQUE_ID'])
```
ehildenb, line 634: I don't think we need `uuid` here, `claim_id` should already likely be the UUID in the case that teh claim doesn't have a name, otherwise it will be the claim name.

src/pyk/proof/reachability.py
```diff
@@ -146,18 +164,42 @@ def summary(self) -> Iterable[str]:
...
+        kcfg_explore: KCFGExplore,
+        main_module_name: str | None = None,
         is_terminal: Callable[[CTerm], bool] | None = None,
         extract_branches: Callable[[CTerm], Iterable[KInner]] | None = None,
     ) -> None:
         self.proof = proof
+        self.kcfg_explore = kcfg_explore
         self._is_terminal = is_terminal
         self._extract_branches = extract_branches
+        self.main_module_name = main_module_name
```
ehildenb, line 187: Could we get `main_module_name` by inspecting `self.kcfg_explore.kprint.definition`?

src/pyk/proof/reachability.py
```diff
@@ -168,9 +210,16 @@ def _check_terminal(self, curr_node: KCFG.Node) -> bool:
                 return True
         return False
 
+    def nonzero_depth(self, node: KCFG.Node) -> bool:
+        init = self.proof.kcfg.init[0]
+        ps = self.proof.kcfg.paths_between(init.id, node.id)
+        for p in ps:
+            if len([n for n in p if type(n) is KCFG.Edge]) >= 2:
+                return True
```
ehildenb, line 218:
```suggestion

        def _path_length(_path: Iterable[KCFG.Successor]) -> int:
            _path = list(_path)
            if len(_path) == 0:
                return 0
            if type(_path[0]) is KCFG.Split or type(_path[0]) is KCFG.Cover:
                return _path_length(_path[1:])
            elif type(_path[0]) is KCFG.NDBranch:
                return 1 + _path_length(path[1:])
            elif type(_path[0]) is KCFG.Edge:
                return _path[0].depth + _path_length(path[1:])
            raise ValueError(f'Cannot handle Successor type: {type(path[0])}')

        path_lengths = map(_path_length, self.proof.kcfg.paths_between(init.id, node.id))
        return sorted(path_lengths)[0] > 0
```

Factor out a function like this to calculate Path length (could live in `kcfg.py` file), and use that directly here.

k-files/imp-simple-spec.k
```diff
@@ -80,4 +80,15 @@ module IMP-SIMPLE-SPEC
...
+
+
+    claim [sum-loop]: <k> while (0 <= $n) { $s = $s + $n ; $n = $n + -1 ; } ~> #FINISH => . ... </k>
+                      <state> $s |-> (S => S +Int ((N *Int (N +Int 1)) /Int 2)) $n |-> (N => -1) </state>
+                      requires 0 <=Int N
+      [circularity]
+
+    claim [sum-N]: <k> int $s, $n, .Ids ; $n = N:Int ; while (0 <= $n) { $s = $s + $n ; $n = $n + -1 ; } ~> #FINISH => . ... </k>
+                      <state> .Map => $s |-> ((N *Int (N +Int 1)) /Int 2) $n |-> -1 </state>
+                    requires 0 <=Int N
```
ehildenb, line 92:
```suggestion
    claim [sum-N]: <k> int $s, $n, .Ids ; $n = N:Int ; while (0 <= $n) { $s = $s + $n ; $n = $n + -1 ; } ~> #FINISH => . ... </k>
                      <state> .Map => $s |-> ((N *Int (N +Int 1)) /Int 2) $n |-> -1 </state>
                    requires 0 <=Int N
            [depends(sum-loop)]
```

ehildenb: Let's make it so that you have to explicitely declare this dependency. Then the modules we make are `CIRCULARITIES-MODULE` and `DEPENDS-MODULE` (or perhaps better names, like `MAIN_MODULE_NAME-CIRCULARITIES` and `MAIN_MODULE_NAME-DEPENDS`), and have the `CIRCULARITIES-MODULE` import the `DEPENDS-MODULE`.

That also means we change the logic slightly when doing proofs. If a proof is labeled as `depends(some-other-proof)` then the other proof is included immediately from step-0. If a proof is labeled as `circularity`, it's included in its own discharge after taking at least 1 step of execution.

h0nzZik:
Should I also include some test which fails to prove a circularity after taking a few steps? To ensure that we really make progress before using the claim itself.

src/tests/integration/kcfg/test_imp.py
h0nzZik: `KSetence.label` is what I need

h0nzZik:
Hi @ehildenb , the backend complains when asked to add a rule which does not have any side condition on the left side: https://github.com/runtimeverification/haskell-backend/blob/5fdb0a4f7f758acde6d498ff10b9a41929be19b0/kore/src/Kore/Rewrite/Rule.hs#L195-L244 - and in PyK, we filter out trivially true side conditions in the CTerm class. Is this something I should address in PyK, or should the backend be adapted? For now I added `requires 0 <=Int 1` to the `if-almost-same` claim, and `use-if-almost-same` can reuse `if-almost-same` just fine this way.

ehildenb:
Can we just add `true` as the side-condition to ones that don't have one when building a claim?

src/pyk/proof/reachability.py
```diff
@@ -31,10 +32,21 @@ class APRProof(Proof):
...
 
-    def __init__(self, id: str, kcfg: KCFG, proof_dir: Path | None = None):
+    def __init__(
+        self,
+        id: str,
+        kcfg: KCFG,
+        proof_dir: Path | None = None,
+        dependencies: Iterable[KClaim] = (),
+        claim_id: str | None = None,
+    ):
```
ehildenb, line 45: Why do we need `id` and `claim_id` both? `id` should already be the `claim_id` in cases where its' coming from a `KClaim`, and otherwise will be something user-specified anyway.

Also, I think `dependencies` should likely not be a list of `KClaim`, but a list of subproofs (a la https://github.com/runtimeverification/pyk/pull/394).

h0nzZik: But subproofs are not in the master yet? Maybe we can change `dependencies` to subproofs later?

src/pyk/proof/reachability.py
```diff
@@ -33,12 +34,21 @@ class APRProof(Proof):
...
 
-    def __init__(self, id: str, kcfg: KCFG, logs: dict[str, tuple[LogEntry, ...]], proof_dir: Path | None = None):
+    def __init__(
+        self,
+        id: str,
+        kcfg: KCFG,
+        logs: dict[str, tuple[LogEntry, ...]],
+        proof_dir: Path | None = None,
+        dependencies: Iterable[KClaim] = (),
+    ):
```
ehildenb, line 47:
```suggestion
    def __init__(
        self,
        id: str,
        kcfg: KCFG,
        logs: dict[str, tuple[LogEntry, ...]],
        proof_dir: Path | None = None,
        dependencies: Iterable[APRProof] = (),
    ):
```

So we are storing subproofs directly.

src/pyk/proof/reachability.py
```diff
@@ -163,18 +179,38 @@ def summary(self) -> Iterable[str]:
...
+        self.kcfg_explore.add_dependencies_module(
+            self.main_module_name,
+            self.some_dependencies_module_name,
+            [c for c in proof.dependencies if c.label != self.proof.id],
+            priority=1,
+        )
+        self.all_dependencies_module_name = self.main_module_name + '-CIRCULARITIES-MODULE'
+        self.kcfg_explore.add_dependencies_module(
+            self.main_module_name, self.all_dependencies_module_name, proof.dependencies, priority=1
+        )
```
ehildenb, line 213:
```suggestion
        self.some_dependencies_module_name = self.main_module_name + '-DEPENDS-MODULE'
        self.kcfg_explore.add_dependencies_module(
            self.main_module_name,
            self.some_dependencies_module_name,
            [build_claim(c) for c in proof.dependencies if c.label != self.proof.id],
            priority=1,
        )
        self.all_dependencies_module_name = self.main_module_name + '-CIRCULARITIES-MODULE'
        self.kcfg_explore.add_dependencies_module(
            self.main_module_name, self.all_dependencies_module_name, proof.dependencies, priority=1
        )
```

Now since we're storing `APRProof`, we need to build the claim right at the moment we are sending it as a module to the backend.

src/tests/integration/k-files/imp-simple-spec.k
```diff
@@ -71,13 +71,30 @@ module IMP-SIMPLE-SPEC
                          <state> $n |-> (0 => 1) $s |-> _S:Int </state>
     
     claim [if-almost-same]:
-        <k> if (_B:Bool) {$s = $n + $s; } else {$s = $s + $n; } ~> #FINISH => . ...</k>
+        <k> if (_B:Bool) {$s = $n + $s; } else {$s = $s + $n; } => . ...</k>
         <state> ($s |-> S:Int $n |-> N:Int) => ($s |-> (S +Int N) $n |-> N) </state>
```
ehildenb, line 75: Let's factor this `if-almost-same` into it's own module `IMP-SIMPLE-SPEC-DEPENDENCIES`, which imports all teh same modules as this one.

Then import that module into this module, and make sure that using `depends(IMP-SIMPLE-SPEC-DEPENDENCIES.if-almost-same)` works (with the fully qualified name).

Also, make a new proof that uses two dependency proofs, so add another proof to _this_ module of a different `if` statement (maybe a multiplication that is commuted instead), then a super-proof `use-fi-almost-same-twice` which uses both the `add` proof from the imported module, and the `mul` proof from this module (without the module name qualifier).

src/pyk/proof/reachability.py
```diff
@@ -62,17 +78,23 @@ def status(self) -> ProofStatus:
...
     def dict(self) -> dict[str, Any]:
         logs = {k: [l.to_dict() for l in ls] for k, ls in self.logs.items()}
-        return {'type': 'APRProof', 'id': self.id, 'cfg': self.kcfg.to_dict(), 'logs': logs}
+        return {
+            'type': 'APRProof',
+            'id': self.id,
+            'cfg': self.kcfg.to_dict(),
+            'dependencies': [c.dict for c in self.dependencies],
+            'logs': logs,
+        }
```
ehildenb, line 97: We need to add `circularity` field in the `dict` and `from_dict`.

ehildenb: Similarly, `dict` and `from_dict` of `APRBMCProof` class need to be modified to handle the new fields.

ehildenb:
Also please update the main PR comment to reflect the current/new architecture.

src/pyk/proof/reachability.py
```diff
@@ -62,17 +78,25 @@ def status(self) -> ProofStatus:
...
 
     @property
     def dict(self) -> dict[str, Any]:
         logs = {k: [l.to_dict() for l in ls] for k, ls in self.logs.items()}
-        return {'type': 'APRProof', 'id': self.id, 'cfg': self.kcfg.to_dict(), 'logs': logs}
+        return {
+            'type': 'APRProof',
+            'id': self.id,
+            'cfg': self.kcfg.to_dict(),
+            'dependencies': [c.dict for c in self.dependencies],
```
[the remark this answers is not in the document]
h0nzZik: Do you have any idea how to test this? We do not have any tests for read_proof/write_proof, do we?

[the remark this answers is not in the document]
ehildenb: No, we don't need to test it. We just want to avoid a bunch of storage that will be redundant.

Baltoli:
@Scott-Guest take a look at the new attributes being used here.

src/pyk/proof/reachability.py
```diff
@@ -62,17 +83,28 @@ def status(self) -> ProofStatus:
     def from_dict(cls: type[APRProof], dct: Mapping[str, Any], proof_dir: Path | None = None) -> APRProof:
         cfg = KCFG.from_dict(dct['cfg'])
         id = dct['id']
+        if len(dct['dependencies']) > 0:
+            if proof_dir is None:
+                raise ValueError('The serialized proof has dependencies but no proof_dir was specified')
+            dependencies = [APRProof.read_proof(id, proof_dir=proof_dir) for id in dct['dependencies']]
+        circularity = dct['circularity']
```
ehildenb, line 90:
```suggestion
        dependencies = [APRProof.read_proof(id, proof_dir=proof_dir) for id in dct['dependencies']]
        circularity = dct['circularity']
```

I guess we'll get a good stack trace if this fails anyway.

src/pyk/proof/reachability.py
```diff
@@ -126,13 +160,24 @@ def status(self) -> ProofStatus:
     def from_dict(cls: type[APRBMCProof], dct: Mapping[str, Any], proof_dir: Path | None = None) -> APRBMCProof:
         cfg = KCFG.from_dict(dct['cfg'])
         id = dct['id']
+        dependencies = [APRProof.from_dict(c) for c in dct['dependencies']]
+        circularity = dct['circularity']
         bounded_states = dct['bounded_states']
```
ehildenb, line 165:
```suggestion
        dependencies = [APRProof.read_proof(id) for id in dct['dependencies']]
        circularity = dct['circularity']
        bounded_states = dct['bounded_states']
```

src/pyk/proof/reachability.py
h0nzZik: @ehildenb the type checker won't allow this, since `proof_dir` can be `None` in the caller, but cannot be `None` in `APRProof.read_proof`.

src/pyk/proof/reachability.py
h0nzZik: This should be similar to the previous piece of code.

h0nzZik:
From a discussion with Everett, it followed that we want to promote the `admitted` flag (as well as the `admit` method) to the base class, and also to show it in the proof summary.

src/pyk/ktool/kprove.py
```diff
@@ -327,23 +327,46 @@ def get_claim_modules(
...
+        spec_file: Path,
+        spec_module_name: str | None = None,
+        include_dirs: Iterable[Path] = (),
+        md_selector: str | None = None,
+    ) -> list[KClaim]:
+        all_claims = self.get_all_claims(
+            spec_file=spec_file, spec_module_name=spec_module_name, include_dirs=include_dirs, md_selector=md_selector
+        )
+        return [cl for l, cl in all_claims.items() if cl.is_circularity]
+
```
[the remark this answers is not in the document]
ehildenb: And then with this being removed, `get_all_claims` is then _only_ being used by `get_claims`. So perhaps we should revert it back to how it was before, since that functionality is not needed anymore.

src/pyk/proof/equality.py
```diff
@@ -108,12 +111,15 @@ def status(self) -> ProofStatus:
             return ProofStatus.PASSED
 
     @classmethod
-    def from_dict(cls: type[EqualityProof], dct: Mapping[str, Any], proof_dir: Path | None = None) -> EqualityProof:
+    def from_dict(
+        cls: type[EqualityProof], dct: Mapping[str, Any], proof_dir: Path | None = None, admitted: bool = False
+    ) -> EqualityProof:
```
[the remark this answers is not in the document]
ehildenb: We also need to include `admitted` field in the `dict` property below.

src/pyk/proof/proof.py
```diff
@@ -129,11 +140,14 @@ def dict(self) -> dict[str, Any]:
...
             'subproof_ids': self.subproof_ids,
+            'admitted': self.admitted,
         }
 
     @classmethod
     @abstractmethod
-    def from_dict(cls: type[Proof], dct: Mapping[str, Any]) -> Proof:
+    def from_dict(
+        cls: type[Proof], dct: Mapping[str, Any], proof_dir: Path | None = None, admitted: bool = False
+    ) -> Proof:
```
ehildenb, line 150: I don't think `admitted` should be part of the `from_dict` signature.

src/pyk/proof/reachability.py
```diff
@@ -109,32 +117,40 @@ def status(self) -> ProofStatus:
             return ProofStatus.PASSED
 
     @classmethod
-    def from_dict(cls: type[APRProof], dct: Mapping[str, Any], proof_dir: Path | None = None) -> APRProof:
+    def from_dict(
+        cls: type[APRProof], dct: Mapping[str, Any], proof_dir: Path | None = None, admitted: bool = False
+    ) -> APRProof:
```
ehildenb, line 122: `admitted` should be read from the `dct`, not passed as argument.

src/pyk/proof/reachability.py
```diff
@@ -234,18 +264,23 @@ def is_pending(self, node_id: NodeIdLike) -> bool:
         )
 
     @classmethod
-    def from_dict(cls: type[APRBMCProof], dct: Mapping[str, Any], proof_dir: Path | None = None) -> APRBMCProof:
+    def from_dict(
+        cls: type[APRBMCProof], dct: Mapping[str, Any], proof_dir: Path | None = None, admitted: bool = False
+    ) -> APRBMCProof:
```
ehildenb, line 269: Admitted should be read from the dict, not passed as argument.

src/pyk/proof/reachability.py
```diff
@@ -296,21 +334,66 @@ def summary(self) -> Iterable[str]:
...
+            fr: CTerm = pf.kcfg.node(pf.init).cterm
+            to: CTerm = pf.kcfg.node(pf.target).cterm
+            fr_config_sorted = self.kcfg_explore.kprint.definition.sort_vars(fr.config, sort=KSort('GeneratedTopCell'))
+            to_config_sorted = self.kcfg_explore.kprint.definition.sort_vars(to.config, sort=KSort('GeneratedTopCell'))
+            kc = KClaim(
+                body=KRewrite(fr_config_sorted, to_config_sorted),
+                requires=ml_pred_to_bool(mlAnd(fr.constraints)),
+                ensures=ml_pred_to_bool(mlAnd(to.constraints)),
+            )
+            return kc
```
ehildenb, line 371: This should be promoted to a method `APRProof.as_claim(kprint: KPrint)`. Also note that maybe you can reuse `cterm.build_claim`, though that one pushes down the rewrites in the configuration, which maybe is not desirable, and it also handles renaming variables to put the appropriate `?` or `_` in places where they are needed.

src/pyk/proof/reachability.py
```diff
@@ -296,21 +345,54 @@ def summary(self) -> Iterable[str]:
...
         self._abstract_node = abstract_node
+        self.main_module_name = self.kcfg_explore.kprint.definition.main_module_name
+
+        subproofs: list[Proof] = (
+            [Proof.read_proof(i, proof_dir=proof.proof_dir) for i in proof.subproof_ids]
+            if proof.proof_dir is not None
+            else []
+        )
+
+        apr_subproofs: list[APRProof] = [pf for pf in subproofs if type(pf) is APRProof]
```
ehildenb, line 378:
```suggestion
        apr_subproofs: list[APRProof] = [pf for pf in subproofs if isinstance(pf, APRProof)]
```

This will also include `APRBMCProof` types as well.

src/pyk/proof/reachability.py
```diff
@@ -296,21 +345,54 @@ def summary(self) -> Iterable[str]:
...
 class APRProver:
     proof: APRProof
+    kcfg_explore: KCFGExplore
     _is_terminal: Callable[[CTerm], bool] | None
     _extract_branches: Callable[[CTerm], Iterable[KInner]] | None
     _abstract_node: Callable[[CTerm], CTerm] | None
 
+    main_module_name: str
+    some_dependencies_module_name: str
+    all_dependencies_module_name: str
```
ehildenb, line 355:
```suggestion
    main_module_name: str
    dependencies_module_name: str
    circularities_module_name: str
```

I think this naming is more clear, but up to your.

ehildenb (approved):

PR #396, opened 2023-04-26
Introduce ProofSchedule data structure

geo2a:
Work towards [evm-semantics#1753](https://github.com/runtimeverification/evm-semantics/issues/1753)

Introduce the `ProofSchedule` data structure that abstracts a directed acyclic graph (DAG) over several proofs that depend on each other, akin to a "proof build system" or a "proof Excel spreadsheet".

### Initialization

A `ProofSchedule` is constructed from a `proof_dir` holding Pyk JSON-serialized proofs and a *list of edges* describing the structure of DAG. See `ProofSchedule.__init__`.

### Execution

The `ProofSchedule.fire_proof(proof_id: str, kcfg_explore: KCFGExplore)` triggers execution of a desired proof in the graph, possible executing its dependencies. If a dependency is `PASSED`, it will not be re-proven.

## Implementation

Internally, the DAG is stored in a `igraph.Graph` object. The `igraph` library provides extensive API for graph manipulation which we would not want to implement ourselves. The `ProofSchedule` already uses topological sorting and subgraph induction implemented by `igraph`.

## Test suite / Examples

The examples are based on almost trivial (only need one simplification) Equality proofs on top of IMP.

See `tests/integration/test_proof_schedule.py#test_fire_two_proofs` for the most interesting example that exercises a forked DAG case, i.e. one root proof on which two branches depend. The first `fire_prove` method call solves the proofs 0, 1 and 2:
schedule-one-branch

And the second solves 3 and 4:
schedule-two-branches


ehildenb:
This is outdated given the updates to `Proof` which now has a notion of subproofs.

PR #397, opened 2023-04-26
Rework the `ktool.kompile` module

tothtamas28:
* Use `**kwargs` in `kompile`
* Factor out type declarations and logic into class hierarchy `Kompile`
* Remove `llvm_kompile` and `haskell_kompile` (if a strict interface is needed classes `LLVMKompile` and `HaskellKompile` can be used)
* Use the new classes to simplify integration test utilities

The goal is to make adding new arguments easier. For each new flag, the following should be added to exactly one of  `KompileArgs`, `LLVMKompile`, `HaskellKompile`:

* A new attribute declaration
* Logic to `__init__` to validate and set the attribute
* Logic to `args` to convert the attribute value to command line arguments


src/pyk/ktool/kompile.py
```diff
@@ -1,327 +1,307 @@
...
+        base_args = KompileArgs(**common_args)
+        match backend:
+            case KompileBackend.HASKELL:
+                haskell_args = {key: value for key, value in dct.items() if key in HASKELL_ARGS}
+                return HaskellKompile(base_args, **haskell_args)
+            case KompileBackend.LLVM:
+                llvm_args = {key: value for key, value in dct.items() if key in LLVM_ARGS}
+                return LLVMKompile(base_args, **llvm_args)
+            case _:
+                raise ValueError(f'Unsupported backend: {backend.value}')
```
ehildenb, line 65: Can we check that if the backend is set to `llvm`, that no haskell args are used, for example?

```suggestion
    def from_dict(dct: Mapping[str, Any]) -> Kompile:
        backend = KompileBackend(dct.get('backend') or 'llvm')
        common_args = {key: value for key, value in dct.items() if key in COMMON_ARGS}
        base_args = KompileArgs(**common_args)
        llvm_args = {key: value for key, value in dct.items() if key in LLVM_ARGS}
        haskell_args = {key: value for key, value in dct.items() if key in HASKELL_ARGS}
        match backend:
            case KompileBackend.HASKELL:
                assert len(llvm_args) == 0
                return HaskellKompile(base_args, **haskell_args)
            case KompileBackend.LLVM:
                assert len(haskell_args) == 0
                return LLVMKompile(base_args, **llvm_args)
            case _:
                raise ValueError(f'Unsupported backend: {backend.value}')
```

For example (though may not be as good as a version that actually prints an error message with the offending args listed).

ehildenb (approved):

src/pyk/ktool/kompile.py
tothtamas28: Done.

PR #402, opened 2023-04-27
Move `debug` from `KompileArgs` to `Kompile.__call__`

tothtamas28:
This argument has no effect on the kompilation result.


ehildenb:
I guess I'm missing the distinction between this and an argument like `--backend haskell`, because this does end up changing the artifacts that are written to disk as part of compilation (namely, the cached Java classes describing the kompilation have the read_only flag set). But approving anyway! Seems fine.

ehildenb (approved):

tothtamas28:
> this does end up changing the artifacts that are written to disk as part of compilation

Then it indeed shouldn't be moved to `__call__`, reverting the change.

PR #404, opened 2023-04-28
Refactors to KCFG edge representations

ehildenb:
This PR is pulled out of: https://github.com/runtimeverification/pyk/pull/329

The goal (and observation) here is that there is a lot of common functionality between both normal edges and multi-edges. In particular, both of them can define `targets` variable, which can simplify routines which are doing traversals of the graph.

- Abstract method `KCFG.Successor.targets` is added, and implemented for all successor types.
- Method `KCFG.Split.target_ids` is promoted to `KCFG.Successor.target_ids`, so all classes have access to it.
- Method `KCFG.successors` now takes arguments indicating which types of successors should be included in the list (default is all of them).
- Method `KCFG.predecessors` is added, which is the dual of `KCFG.successors` and takes the same arguments.
- Method `KCFG.reachable_nodes` uses the new `KCFG.{successors/predecessors}` to good effect to simplify quite a bit.
- Method `KCFG.Split.with_single_target` is added, which returns a new `Split` with only the one target in place (used in `paths_between`).
- `KCFG.Split` now names its argument `_targets` instead, so that it can match the signature of `@property targets`, but still exposes `@property splits` as a dictionary.
- `KCFG.prune` now returns the list of pruned nodes, makes sure to traverse covers as well.
- Makes `_print_split_edge` handle projecting out the components of the `KCFG.Split` node for printing directly, so that the same interface should work for printing `KCFG.NDBranch` with minimal modification.


src/tests/unit/test_kcfg.py
```diff
@@ -56,7 +56,7 @@ def split(i: int, js: Iterable[int]) -> KCFG.Split:
         csubst = term(i).match_with_constraint(term(j))
         assert csubst is not None
         split_substs.append(csubst)
-    return KCFG.Split(node(i), zip(map(node, js), split_substs, strict=True))
+    return KCFG.Split(node(i), tuple(zip((node(j) for j in js), split_substs, strict=True)))
```
tothtamas28, line 59: Consider implementing `Split.__init__(source: ...,  targets: Iterable[...])`, then

```suggestion
    return KCFG.Split(node(i), zip((node(j) for j in js), split_substs, strict=True))
```

src/pyk/kcfg/kcfg.py
```diff
@@ -49,10 +59,23 @@ def __lt__(self, other: Any) -> bool:
...
+
     class EdgeLike(Successor):
         source: KCFG.Node
         target: KCFG.Node
 
+        @property
+        def targets(self) -> tuple[KCFG.Node, ...]:
+            return (self.target,)
+
     @dataclass(frozen=True)
```
tothtamas28, line 79: Let's mark these dataclasses `@final`, it helps the reader understand the `Successor` class hierarchy.

tothtamas28 (approved):

src/pyk/kcfg/kcfg.py
```diff
@@ -29,7 +28,18 @@
...
+        Union[
+            'KCFG.Node',
+            'KCFG.Successor',
+            'KCFG.EdgeLike',
+            'KCFG.Edge',
+            'KCFG.Cover',
+            'KCFG.Split',
+        ]
+    ]
+):
```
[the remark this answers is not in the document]
ehildenb: Done.

src/tests/unit/test_kcfg.py
ehildenb: Done.

src/pyk/kcfg/kcfg.py
ehildenb: Done.

Issue #393, opened 2023-04-25
Debug data in the KCFG viewer


goodlyrottenapple:
The booster backend includes a lot of granular tracing of everything happening during rewrite, including failed unfication/matches and simplification/function evaluation of terms and side-conditions. Because the development of the simplifier is getting ricky, we would like to integrate all this debugging data into the KCFG viewer. This will be useful not only to the haskell team but also the end users, who could use the debug mode to trace why e.g. a certain simplification rule isn't being applied, etc.
This proposal supersedes https://github.com/runtimeverification/haskell-backend/issues/3561 in that the booster should support the above plus a much finer-grained tracing info.

## RPC protocol changes

When one of the following flags: `log_successful_rewrites`, `log_failed_rewrites`, `log_successful_rsimplifications` or `log_failed_simplifications` are sent with the execute/simplify request, the backends should return a `log` (or similar) object as part of the response. The response should in fact be an array of rewrite/simplify traces, with the following format:

### rewrite success

if a semantic rule succeeds, yields

```yaml
tag: rewrite_success
rewritten_term: <optional CTerm>
rule_id: <unique_id>
```

the `rewritten_term` is optional as it may not be present when executing with the old backend (unclear how difficult it would be to implement)

### rewrite failure

if a rule fails to apply, the backend will emit

```yaml
tag: rewrite_failure
reason : <string>
rule_id: <unique_id>
```

### simplification
The booster distinguishes between concrete and symbolic simplification and we want to be able to tell which one was applied and how often, therefore we can have either of the two traces

```yaml
tag: llvm_simplification
original_term : <KInner>
result : <SimplificationResult>
```

```yaml
tag: simplification
original_term : <optional KInner>
result : <SimplificationResult>
```

where `original_term` may be optional when coming from the old backend and the `SimplificationResult` object can be one of the following:

```yaml
tag: success
rewritten_term : <optional KInner>
rule_id: <unique_id>
```

where the `rewritten_term` is optional for the old backend and `rule` is optional because when calling the LLVM backend, we have no way of obtaining this information (the LLVM backend can in fact apply many rules in sequence)

```yaml
tag: failure
reason: <string>
rule_id: <unique_id>
```

This schema should cover the use-cases of the original issue as well as provide a detailed view of all the intermediate rewrite steps and simplifications.

## Pyk data representation changes

[...]

PR #405, opened 2023-04-28
Add logs for kore-rpc calls and KCFG nodes

goodlyrottenapple:
This PR implements the proposed changes in https://github.com/runtimeverification/pyk/issues/393, namely:

* Add `LogEntry` class to `kore.rpc` which holds the logging data returned from the rpc server execute endpoint
* Add `logs` attribute to the `ExecuteResult` as well as `ImpliesResult` and the simplify rpc endpoint (the last two require https://github.com/runtimeverification/haskell-backend/pull/3580 to be merged)
* Store the logs inside the KCFG `Node` for analysis via the kcfg-viewer (will be implemented in a followup PR)


goodlyrottenapple:
This is just a draft atm as I've not tested this yet. Just wanted to see if my approach is correct.

ehildenb:
I think we should also attach the same logs to `simplify` endpoint and `check-implication` endpoint, so that we can get the equations applied while trying to do those routines.

It can be the same thing where we make best effort to have the old backend fill in those data, but prioritize the new backend filling in those data when those endpoints come available in the new backend.

ehildenb:
Rather than attaching the logs to the KCFG, I would rather have the `KCFGExplore` class track the logs external to the `KCFG` datastructure. So the calles `KCFGExplore.cterm_*` ... return a `tuple[result, LOGS]`. Then teh user can decide what to do with those log entries, and store them in a more appropriate datastructure.

ehildenb:
For this PR, let's just have the various `KCFGExplore.cterm_*` methods take an extra `log: bool = False` option, and return the logs as part of their return type (in a `tuple[ORIGINAL_RETURN_TYPE, Iterable[Log] | None]`). Then we can get this merged, and @geo2a is working on overhauling teh KCFG viewers to be more specific to the proof types: https://github.com/runtimeverification/pyk/issues/413, where we can incorporate a "debug" view to show the log output.

ehildenb (approved):

PR #411, opened 2023-05-04
ktool/kprint: undo aliases when pretty-printing

ehildenb:
Fixes: https://github.com/runtimeverification/evm-semantics/issues/1801

A simple change to undo aliases at unparsing time before pretty-printing, which can be turned off by the user if needed. Initial experiments with KEVM were showing that it took 5-10s to undo the aliases using a naive rewriting technique, os a new one was introduced.

- Several functions in `kast.manip` were removed that had no uses: `disjunct_constrained_terms`, `remove_disjuncts`, `constraint_subsume`, and `match_with_constraint`.
- `TestDefn` is renamed to `TestImpDefn`, and is made parametric in `test_pretty_print`.
- `TestAliasDefn` is added, with `k-flies/aliases.k`, and several (previously failing) tests are added. One test is added which we'd like even better output by undoing aliases with AC symbols.
- A simple `indexed_rewrite` method is added to `kast.manip` which can very quickly apply lots of rewrites over a tree, by:
  - Separating the rewrites into different buckets based on their top symbol on the LHS.
  - Traverse the entire term a single time applying all the relevant rewrites at each given point instead of traversing separately for each rewrite.
- `undo_aliases` is fixed to reverse the aliases, and use the fast `indexed_rewrite` to apply them.
- `KPrint.pretty_print` now undoes aliases by default, but allows the user to turn that off.


deps/k_release
```diff
@@ -1 +1 @@
-5.6.77
+5.6.85
```
ehildenb, line 1: Not sure why this is showing up here, this change was just merged: https://github.com/runtimeverification/pyk/pull/410

ehildenb:
Here is this working on KEVM (very fast printing): 
image

The `pow*` are aliases defined in KEVM, and `maxBlockNum`.

src/pyk/ktool/kprint.py
```diff
@@ -303,7 +303,9 @@ def _add_sort_injection(self, pat: Pattern, isort: KSort, osort: KSort) -> Patte
             )
         return App('inj', [SortApp('Sort' + isort.name), SortApp('Sort' + osort.name)], [pat])
 
-    def pretty_print(self, kast: KAst) -> str:
+    def pretty_print(self, kast: KAst, unalias: bool = True) -> str:
```
tothtamas28, line 306:
```suggestion
    def pretty_print(self, kast: KAst, *, unalias: bool = True) -> str:
```

src/pyk/kast/manip.py
```diff
@@ -639,45 +620,57 @@ def apply_existential_substitutions(constrained_term: KInner) -> KInner:
...
-
-
-def match_with_constraint(constrained_term_1: KInner, constrained_term_2: KInner) -> Subst | None:
-    state1, constraint1 = split_config_and_constraints(constrained_term_1)
-    state2, constraint2 = split_config_and_constraints(constrained_term_2)
-    subst = state1.match(state2)
-    if subst is not None and constraint_subsume(Subst(subst)(constraint1), constraint2):
-        return subst
-    return None
+def indexed_rewrite(rewrites: Iterable[KRewrite], kast: KInner) -> KInner:
```
tothtamas28, line 623:
```suggestion
def indexed_rewrite(kast: KInner, rewrites: Iterable[KRewrite]) -> KInner:
```

tothtamas28 (approved):

src/pyk/ktool/kprint.py

PR #415, opened 2023-05-09
Add `implication_failure_reason` to KCFGExplore

nwatson22:
Adds the method `implication_failure_reason` to KCFGExplore. This method was originally written by @ehildenb as part of the K summarizer. It makes an implication check between two CTerms and attempts to show why the implication fails, if it fails (the cells that didn't match or the leftover implication between constraints).
Adds a test [rc/tests/integration/kcfg/test_impl_failure_reason.py](https://github.com/runtimeverification/pyk/compare/noah/implication-failure-reason?expand=1#diff-f0f5b7f4cadcc3620c868a84af0daf7754f2016b4f66ea2fcbcb0fd2182481a9) to exercise this method.


src/tests/integration/kcfg/test_impl_failure_reason.py
```diff
@@ -0,0 +1,72 @@
...
+class TestImpProof(KCFGExploreTest):
+    KOMPILE_MAIN_FILE = 'k-files/imp.k'
+
+    def test_implication_failure_reason(
+        self,
+        kcfg_explore: KCFGExplore,
+        kprove: KProve,
+    ) -> None:
+        lhs_config = KApply('<T>', [KApply('<k>', [KToken('1', 'Int')]), KApply('<state>', [KToken('.Map', 'Map')])])
+        rhs_config = KApply('<T>', [KApply('<k>', [KToken('2', 'Int')]), KApply('<state>', [KToken('.Map', 'Map')])])
```
ehildenb, line 25: Put this test into `kcfg/test_imp.py`, and factor out a testing harness that is parametric using data like `IMPLICATION_FAILURE_TEST_DATA`, and a parametrized test `test_implification_failure`. There are other parametric tests in that file to use as examples of how to do this.

src/pyk/kcfg/explore.py
```diff
@@ -178,6 +191,85 @@ def cterm_implies(
...
+        concrete_config, *concrete_constraints = concrete
+        abstract_config, *abstract_constraints = abstract
+        config_match = abstract_config.match(concrete_config)
+        if config_match is None:
+            _, concrete_subst = split_config_from(concrete_config)
+            cell_names = concrete_subst.keys()
+            failing_cells = []
+            for cell in cell_names:
+                concrete_cell = get_cell(concrete_config, cell)
+                abstract_cell = get_cell(abstract_config, cell)
```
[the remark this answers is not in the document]
nwatson22: I switched to using the suggested methods.

src/tests/integration/kcfg/test_impl_failure_reason.py
nwatson22: I moved the test into `test_imp.py` and it is properly parameterized now.

src/pyk/kcfg/explore.py
```diff
@@ -178,6 +189,83 @@ def cterm_implies(
...
+            if impl != mlTop(k.GENERATED_TOP_CELL):
+                fail_str = self.kprint.pretty_print(impl)
+                negative_cell_constraints = list(filter(_is_negative_cell_subst, concrete_constraints))
+                if len(negative_cell_constraints) > 0:
+                    fail_str = (
+                        f'{fail_str}\n\nNegated cell substitutions found (consider using _ => ?_):\n'
+                        + '\n'.join([self.kprint.pretty_print(cc) for cc in negative_cell_constraints])
+                    )
+                return (False, f'Implication check failed, the following is the remaining implication:\n{fail_str}')
+        return (True, '')
```
ehildenb, line 267:
```suggestion
            abstract_constraints = [
                config_match.subst.apply(abstract_constraint) for abstract_constraint in abstract.constraints
            ]
            abstract_constraints = list(
                filter(
                    lambda x: not CTerm._is_spurious_constraint(x),
                    [config_match.subst.apply(abstract_constraint) for abstract_constraint in abstract.constraints],
                )
            )
            impl = CTerm._ml_impl(concrete_constraints, abstract_constraints)
            if impl != mlTop(k.GENERATED_TOP_CELL):
                fail_str = self.kprint.pretty_print(impl)
                negative_cell_constraints = list(filter(_is_negative_cell_subst, concrete_constraints))
                if len(negative_cell_constraints) > 0:
                    fail_str = (
                        f'{fail_str}\n\nNegated cell substitutions found (consider using _ => ?_):\n'
                        + '\n'.join([self.kprint.pretty_print(cc) for cc in negative_cell_constraints])
                    )
                return (False, f'Implication check failed, the following is the remaining implication:\n{fail_str}')
        return (True, '')
```

src/tests/integration/kcfg/test_imp.py
```diff
@@ -71,6 +71,38 @@
     ),
 )
 
+IMPLICATION_FAILURE_TEST_DATA: Final = (
+    (
+        'different-cell',
+        ('int $n ; $n = 0 ;', '.Map'),
+        ('int $n ; $n = 1 ;', '.Map'),
+        'Structural matching failed, the following cells failed individually (abstract => concrete):\nK_CELL: ( int $n , .Ids ; $n = 0 ; => int $n , .Ids ; $n = 1 ; )',
```
ehildenb, line 79:
```suggestion
        (
            'Structural matching failed, the following cells failed individually (abstract => concrete):\n'
            'K_CELL: ( int $n , .Ids ; $n = 0 ; => int $n , .Ids ; $n = 1 ; )'
        ),
```

src/tests/integration/kcfg/test_imp.py
```diff
@@ -71,6 +71,38 @@
...
+            'int $n ; $n = 0 ;',
+            '1 |-> A:Int 2 |-> B:Int',
+            mlAnd(
+                [
+                    mlEqualsTrue(KApply('_<Int_', [KVariable('A', 'Int'), KToken('1', 'Int')])),
+                    mlEqualsTrue(KApply('_<Int_', [KVariable('B', 'Int'), KToken('2', 'Int')])),
+                ]
+            ),
+        ),
+        'Implication check failed, the following is the remaining implication:\n{ true #Equals A:Int <Int 1 }\n#And { true #Equals B:Int <Int 1 } #Implies { true #Equals B:Int <Int 2 }',
```
ehildenb, line 103:
```suggestion
        (
            'Implication check failed, the following is the remaining implication:\n'
            '{ true #Equals A:Int <Int 1 }\n'
            '#And { true #Equals B:Int <Int 1 } #Implies { true #Equals B:Int <Int 2 }'
        ),
```

src/tests/integration/kcfg/test_imp.py
```diff
@@ -711,3 +743,25 @@ def test_functional_prove(
...
+        consequent: tuple[str, str] | tuple[str, str, KInner],
+        expected: str,
+    ) -> None:
+        antecedent_term = self.config(kcfg_explore.kprint, *antecedent)
+        consequent_term = self.config(kcfg_explore.kprint, *consequent)
+
+        (b, s) = kcfg_explore.implication_failure_reason(antecedent_term, consequent_term)
+
+        assert b == False
+        assert s == expected
```
ehildenb, line 767:
```suggestion
        failed, actual = kcfg_explore.implication_failure_reason(antecedent_term, consequent_term)

        assert failed == False
        assert actual == expected
```

src/pyk/kcfg/explore.py
```diff
@@ -178,6 +189,83 @@ def cterm_implies(
...
+                negative_constraint = constraint_bool.args[0]
+                if type(negative_constraint) is KApply and negative_constraint.label.name == '_andBool_':
+                    constraints = flatten_label('_andBool_', negative_constraint)
+                    cell_constraints = list(filter(_is_cell_subst, constraints))
+                    if len(cell_constraints) > 0:
+                        return True
+            return False
+
+        concrete_config, *concrete_constraints = concrete
+        abstract_config, *abstract_constraints = abstract
```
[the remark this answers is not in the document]
nwatson22: I removed these. It looked like the `abstract_config` was iteratively having substitutions applied to itself which is why I didn't remove them before, but I guess it actually being used after that, so I removed that part too.

src/pyk/kcfg/explore.py
nwatson22: Applied these changes.

src/pyk/kcfg/explore.py
```diff
@@ -178,6 +189,76 @@ def cterm_implies(
...
+                False,
+                f'Structural matching failed, the following cells failed individually (abstract => concrete):\n{failing_cells_str}',
+            )
+        else:
+            abstract_constraints = list(
+                filter(
+                    lambda x: not CTerm._is_spurious_constraint(x),
+                    [config_match.subst.apply(abstract_constraint) for abstract_constraint in abstract.constraints],
+                )
+            )
```
ehildenb, line 249: Does this work?

```suggestion
            abstract_constraints = list(filter(CTerm._is_spurious_constraint, map(config_match.subst, abstract.constraints)))
```

src/pyk/kcfg/explore.py
```diff
@@ -178,6 +189,76 @@ def cterm_implies(
...
+        config_match = self.cterm_implies(CTerm.from_kast(abstract.config), CTerm.from_kast(concrete.config))
+        if config_match is None:
+            failing_cells = []
+            for cell in concrete.cells:
+                concrete_cell = concrete.cell(cell)
+                abstract_cell = abstract.cell(cell)
+                _LOGGER.info(f'concrete_cell: {concrete_cell}')
+                _LOGGER.info(f'abstract_cell: {abstract_cell}')
+                cell_match = abstract_cell.match(concrete_cell)
+                _LOGGER.info(f'cell_match: {cell_match}')
```
ehildenb, line 231: I think we can remove these debugging statements.

```suggestion
                cell_match = abstract_cell.match(concrete_cell)
```

src/pyk/kcfg/explore.py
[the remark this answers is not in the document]
ehildenb: Hmmmm, actually we should probably be iteratively applying the substitution to new `abstract_config` each time....

src/pyk/kcfg/explore.py
```diff
@@ -178,6 +189,76 @@ def cterm_implies(
...
+
+        config_match = self.cterm_implies(CTerm.from_kast(abstract.config), CTerm.from_kast(concrete.config))
+        if config_match is None:
+            failing_cells = []
+            for cell in concrete.cells:
+                concrete_cell = concrete.cell(cell)
+                abstract_cell = abstract.cell(cell)
+                _LOGGER.info(f'concrete_cell: {concrete_cell}')
+                _LOGGER.info(f'abstract_cell: {abstract_cell}')
+                cell_match = abstract_cell.match(concrete_cell)
```
ehildenb, line 230:
```suggestion
        config_match = self.cterm_implies(CTerm.from_kast(abstract.config), CTerm.from_kast(concrete.config))
        if config_match is None:
            failing_cells = []
            curr_cell_match = Subst({})
            for cell in concrete.cells:
                concrete_cell = concrete.cell(cell)
                abstract_cell = abstract.cell(cell)
                _LOGGER.info(f'concrete_cell: {concrete_cell}')
                _LOGGER.info(f'abstract_cell: {abstract_cell}')
                cell_match = abstract_cell.match(concrete_cell)
                curr_cell_match = curr_cell_match.union(cell_match)
```

src/pyk/kcfg/explore.py
```diff
@@ -178,6 +189,73 @@ def cterm_implies(
...
+        if config_match is None:
+            failing_cells = []
+            for cell in concrete.cells:
+                concrete_cell = concrete.cell(cell)
+                abstract_cell = abstract.cell(cell)
+                cell_match = abstract_cell.match(concrete_cell)
+                if cell_match is None:
+                    failing_cell = push_down_rewrites(KRewrite(concrete_cell, abstract_cell))
+                    failing_cell = no_cell_rewrite_to_dots(failing_cell)
+                    failing_cells.append((cell, failing_cell))
```
ehildenb, line 232:
```suggestion
        config_match = self.cterm_implies(CTerm.from_kast(abstract.config), CTerm.from_kast(concrete.config))
        if config_match is None:
            failing_cells = []
            curr_cell_match = Subst({})
            for cell in concrete.cells:
                concrete_cell = concrete.cell(cell)
                abstract_cell = abstract.cell(cell)
                cell_match = abstract_cell.match(concrete_cell)
                if cell_match is not None:
                    _curr_cell_match = curr_cell_match.union(cell_match)
                    if _curr_cell_match is not None:
                        curr_cell_match = _curr_cell_match
                        continue
                failing_cell = push_down_rewrites(KRewrite(concrete_cell, abstract_cell))
                failing_cell = no_cell_rewrite_to_dots(failing_cell)
                failing_cells.append((cell, failing_cell))
```

ehildenb: Perhaps we could accumulate the substitution this way?
It would be good to make sure that with an accumulated substitution, where it succeeds on each cell individually, but fails on some certain specific cells, that the failure is reported correctly.

For example:

```
<k> X </k> <state> 3 |-> X </state>

<k> 5 </k> <state> 3 |-> 6 </state>
```

This test should report a failure on the `<state>` cell likely, because the variable `X` is already bound to `5` by the `<k>` cell.

src/pyk/kcfg/explore.py
nwatson22: It seems so.

src/pyk/kcfg/explore.py
nwatson22: Added the accumulation code and the test. It seems to work (reports only the state cell as failure).

src/pyk/kcfg/explore.py
```diff
@@ -178,6 +189,75 @@ def cterm_implies(
...
+                failing_cell = push_down_rewrites(KRewrite(concrete_cell, abstract_cell))
+                failing_cell = no_cell_rewrite_to_dots(failing_cell)
+                failing_cells.append((cell, failing_cell))
+            failing_cells_str = '\n'.join(
+                f'{cell}: {self.kprint.pretty_print(minimize_term(rew))}' for cell, rew in failing_cells
+            )
+            return (
+                False,
+                f'Structural matching failed, the following cells failed individually (abstract => concrete):\n{failing_cells_str}',
+            )
```
ehildenb, line 244: On line 235, we say `KRewrite(concrete_cell, abstract_cell)`, but we are saying on line 243 that it's `abstract => concrete`, and looking at the test output, it seems that it's backward.

```suggestion
                failing_cell = push_down_rewrites(KRewrite(abstract_cell, concrete_cell))
                failing_cell = no_cell_rewrite_to_dots(failing_cell)
                failing_cells.append((cell, failing_cell))
            failing_cells_str = '\n'.join(
                f'{cell}: {self.kprint.pretty_print(minimize_term(rew))}' for cell, rew in failing_cells
            )
            return (
                False,
                f'Structural matching failed, the following cells failed individually (abstract => concrete):\n{failing_cells_str}',
            )
```

src/tests/integration/kcfg/test_imp.py
```diff
@@ -71,6 +71,54 @@
...
 )
 
+IMPLICATION_FAILURE_TEST_DATA: Final = (
+    (
+        'different-cell',
+        ('int $n ; $n = 0 ;', '.Map'),
+        ('int $n ; $n = 1 ;', '.Map'),
+        (
+            'Structural matching failed, the following cells failed individually (abstract => concrete):\n'
+            'K_CELL: ( int $n , .Ids ; $n = 0 ; => int $n , .Ids ; $n = 1 ; )'
```
ehildenb, line 81:
```suggestion
            'K_CELL: ( int $n , .Ids ; $n = 1 ; => int $n , .Ids ; $n = 0 ; )'
```

src/tests/integration/kcfg/test_imp.py
```diff
@@ -71,6 +71,54 @@
...
+            '#And { true #Equals B:Int <Int 1 } #Implies { true #Equals B:Int <Int 2 }'
+        ),
+    ),
+    (
+        'substitution',
+        ('int $n ; $n = 5 ;', '3 |-> 6'),
+        ('int $n ; $n = X:Int ;', '3 |-> X:Int'),
+        (
+            'Structural matching failed, the following cells failed individually (abstract => concrete):\n'
+            'STATE_CELL: 3 |-> ( 6 => X:Int )'
```
ehildenb, line 118:
```suggestion
            'STATE_CELL: 3 |-> ( X:Int => 6 )'
```

src/pyk/kcfg/explore.py
```diff
@@ -178,6 +189,75 @@ def cterm_implies(
...
+            if type(constraint_bool) is KApply and constraint_bool.label.name == 'notBool_':
+                negative_constraint = constraint_bool.args[0]
+                if type(negative_constraint) is KApply and negative_constraint.label.name == '_andBool_':
+                    constraints = flatten_label('_andBool_', negative_constraint)
+                    cell_constraints = list(filter(_is_cell_subst, constraints))
+                    if len(cell_constraints) > 0:
+                        return True
+            return False
+
+        config_match = self.cterm_implies(CTerm.from_kast(abstract.config), CTerm.from_kast(concrete.config))
```
ehildenb, line 222: I think this should be the other way around right? The `abstract` should be the `consequent`, and the `concrete` should be the `antecedent`?

ehildenb: This worked because all of the tests are either (i) exactly the same on either side of the antecedent/consequent (the 2nd test), or (ii) failing at the matching phase anyway.

So we should add a test that (i) succeeds at the matching phase _with an actual match_, but (ii) fails at the constraint phase. For example:

```
antecedent: <k> X:Int </k> <state> .Map </state> #And { X #Equals 3 }
consequent: <k> Y:Int </k> <state> .Map </state> #And { Y #Equals 4 }
```

This should fail with message:

```
            'Implication check failed, the following is the remaining implication:\n'
            '{ Y #Equals 3 } #Implies { Y #Equals 4 }\n'
```

ehildenb (approved):

src/tests/integration/kcfg/test_imp.py
```diff
@@ -71,6 +71,91 @@
...
+            'K_CELL: ( int $n , .Ids ; $n = 1 ; => int $n , .Ids ; $n = 0 ; )'
+        ),
+    ),
+    (
+        'different-cell-2',
+        ('int $n ; $n = X:Int ;', '.Map'),
+        ('int $n ; $n = 1 ;', '.Map'),
+        (
+            'Structural matching failed, the following cells failed individually (abstract => concrete):\n'
+            'K_CELL: ( int $n , .Ids ; $n = X:Int ; => int $n , .Ids ; $n = 1 ; )'
```
ehildenb, line 90: I guess this needs to be flipped.

Can we also just change it everywhere to say `abstract => consequent` and `concrete => antecedent`? It's confusing to have different terminology for them, should just use the `antecedent #Implies consequent` everwhere.

When outputting this error message, we should do it in implication order, with an actual implication. So we ideally would output:

```suggestion
            'K_CELL: ( int $n , .Ids ; $n = 1 ; #Implies int $n , .Ids ; $n = X:Int ; )'
```

This will require, after doing `push_down_rewrites` and `no_cell_rewrite_to_dots`, to make a routine called `replace_rewrites_with_implies`, and run that on the term. It can look like this:

```
def replace_rewrites_with_implies(kast: KInner) -> KInner:
    def _replace_rewrites_with_implies(_kast: KInner) -> KInner:
        if type(_kast) is KRewrite:
            return mlImplies(_kast.lhs, _kast.rhs)
        return _kast
    return bottom_up(_replace_rewrites_with_implies, kast)
```

PR #416, opened 2023-05-09
`cterm_implies`: always bind free variables in consequent

geo2a:
This PR changes `KCFGExplore.cterm_implies`:
* to always bind variables in the consequent

This change is motivated by #394


ehildenb:
So, actually, upon inspecting the method, it appears that we have that we _always_ return `None` if `result.satisfiable == false`, and we always return a `CSubst` if `result.satisfiable == true`. Is that the case? Then do we really need to change the return type here, or can we just inspect if it's `None` or not?

ehildenb (approved):

PR #419, opened 2023-05-10
Equality proof updates

geo2a:
This PR is factored-out from #394. Changes:

* `EqualityProver.advance_proof` now calls `cterm_implies` to check the implication of form `(dummy_config #And constraints) #Implies ((dummy_config #And LHS) #Equals (dummy_config #And  RHS)))` where `dummy_config` is the configuration of the semantics (in our case, IMP) with all cells holding symbolic variables.
* tests for various edge-cases, specifically when the constraints are `#Bottom`. Is this case, the proof should pass trivially, issuing a warning to the user.


src/pyk/proof/equality.py
```diff
@@ -28,55 +31,52 @@ class EqualityProof(Proof):
...
         rhs_body = extract_rhs(claim.body)
         assert type(lhs_body) is KApply
         sort = defn.return_sort(lhs_body.label)
-        lhs_constraints = flatten_label('_andBool_', claim.requires)
-        rhs_constraints = flatten_label('_andBool_', claim.ensures)
-        return EqualityProof(
-            claim.label, lhs_body, rhs_body, sort, lhs_constraints=lhs_constraints, rhs_constraints=rhs_constraints
-        )
+        constraints = [mlEquals(TRUE, c, arg_sort=BOOL) for c in flatten_label('_andBool_', claim.requires)]
+        return EqualityProof(claim.label, lhs_body, rhs_body, sort, constraints=constraints)
```
ehildenb, line 63: We should add a check that `claim.ensures is None or claim.ensures is TRUE`, and raise an exception if not.

src/pyk/proof/equality.py
```diff
@@ -124,18 +125,30 @@ def dict(self) -> dict[str, Any]:
...
+        lines.append(f'Status: {self.status}')
+        return lines
+
     @property
     def summary(self) -> Iterable[str]:
         return [
             f'EqualityProof: {self.id}',
-            f'    status: {self.status}',
+            f'    satisfiable: {self.satisfiable}',
         ]
```
ehildenb, line 152: We should still print out the `status` as well ,not just `satisfiable`.

src/pyk/proof/equality.py
```diff
@@ -146,13 +159,37 @@ def __init__(self, proof: EqualityProof) -> None:
...
     def advance_proof(self, kcfg_explore: KCFGExplore) -> None:
-        if not self.proof.status == ProofStatus.PENDING:
+        if self.proof.csubst is not None:
             return
 
-        kore = kcfg_explore.kprint.kast_to_kore(self.proof.equality)
+        # to prove the equality, we check the implication of the form `constraints -> LHS ==K RHS`, i.e.
+        # "LHS equals RHS under these constraints"
+        antecedent_kast = mlAnd(self.proof.constraints)
+        consequent_kast = mlEquals(TRUE, self.proof.equality, arg_sort=BOOL, sort=GENERATED_TOP_CELL)
```
ehildenb, line 168: So we're building `CONSTRAINTS #Implies { true #Equals (LHS ==K RHS) }`, did we try instead `CONSTRAINTS #Implies { LHS #Equals RHS }` at some point? Seems like a slightly simpler construction to me.

src/pyk/proof/equality.py
```diff
@@ -89,30 +89,31 @@ def read_proof(id: str, proof_dir: Path) -> EqualityProof:
...
-        elif is_top(self.simplified_equality):
-            return ProofStatus.PASSED
         else:
-            return ProofStatus.FAILED
+            if is_bottom(self.simplified_constraints):  # the proof passes trivially
+                return ProofStatus.PASSED
+            elif self.csubst is not None and is_top(self.csubst.constraint):
+                return ProofStatus.PASSED
+            else:
+                return ProofStatus.FAILED
```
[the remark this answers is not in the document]
ehildenb: I think this should read:

```
if self.simplified_constraints is None:
    return ProofStatus.PENDING
elif self.csubst is None:
    return ProofStatus.FAILED
else:
    return ProofStatus.PASSED
```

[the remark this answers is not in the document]
ehildenb: So we basically use whether `simplified_constraints` is populated to determine if we've _attempted_ simplification, and if we have already attempted simplification, we use `csubst` to determine the results of that.

Notice, it's OK for `csubst` to have a `constraint` attached to it, that's part of the witness. I think it's unlikely that we'll see that case very often, but it doesn't invalidate the proof, we still have that the models of the constraints are also models of the equality we're trying to show.

src/pyk/proof/equality.py
```diff
@@ -146,13 +159,37 @@ def __init__(self, proof: EqualityProof) -> None:
         self.proof = proof
 
     def advance_proof(self, kcfg_explore: KCFGExplore) -> None:
-        if not self.proof.status == ProofStatus.PENDING:
+        if self.proof.csubst is not None:
             return
```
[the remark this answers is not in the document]
ehildenb: See later comment. We should just rely on status here, which will make sure we're not re-running the proof if it's already been run.

src/pyk/proof/equality.py
```diff
@@ -146,13 +159,37 @@ def __init__(self, proof: EqualityProof) -> None:
...
+        dummy_config = kcfg_explore.kprint.definition.empty_config(sort=GENERATED_TOP_CELL)
+        result = kcfg_explore.cterm_implies(
+            antecedent=CTerm(config=dummy_config, constraints=[antecedent_kast]),
+            consequent=CTerm(config=dummy_config, constraints=[consequent_kast]),
+        )
+        if result is None:
+            _LOGGER.warning('cterm_implies returned None, the implication is unsatisfiable')
+        else:
+            self.proof.set_csubst(result)
         self.proof.write_proof()
```
ehildenb, line 195: Some changes I would make:

- Also store a `simplified_consequent`, always compute both.
- Check if `simplified_antecedent` is `#Bottom` or if `simplified_consequent` is `#Top`. If either, just store `CSubst(Subst({}), ())` as the `csubst`, and return (proof passed)..
- If that check fails, check if `simplified_consequent` is `#Bottom`, if it is, don't store anything as `csubst`, and return (proof failed).
- Otherwise build the dummy implication and get the result back. If it's an actual `CSubst`, store it in `csubst` and return (proof passed), if not then store nothing and return (proof failed).

Since we're always computing `simplified_consquent` and `simplified_antecedent`, the fact that those fields are present mark whether we have attempted the proof (no longer pending). Then whether `csubst` is populated or not marks whether teh attempted proof has passed or not.

src/tests/integration/k-files/imp-simple-spec.k
```diff
@@ -5,6 +5,27 @@ module IMP-FUNCTIONAL-SPEC
...
+   claim [concrete-requires-trivial-false-identity]: 0 => 0
+     requires False
+
+   claim [concrete-requires-nontrivial-false-identity]: 0 => 0
+     requires 0 ==Int 1
+
+//   TODO: this should be trivially passing but it fails because cterm_implies returns None
+//   claim [concrete-requires-trivial-false-nonsense]: 0 => 1
+//     requires False
+
```
ehildenb, line 23: Try this one as well:

```suggestion
//   TODO: this should be trivially passing but it fails because cterm_implies returns None
//   claim [concrete-requires-trivial-false-nonsense]: 0 => 1
//     requires False

   claim [concrete-requires-trivial-false-nonsense]: _X:Int => ?_Y:Int
     requires False
```

ehildenb: Likely we need to file a bug about that to the backend too. Even if the consequent is trivially empty, it should still check if the antecedent is empty and pass it trivially, as you say.

src/pyk/proof/equality.py
geo2a: Done.

src/pyk/proof/equality.py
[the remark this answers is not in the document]
geo2a: Done.

src/pyk/proof/equality.py
geo2a: Done.

src/pyk/proof/equality.py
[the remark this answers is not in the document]
geo2a: Done.

src/pyk/proof/equality.py
geo2a: great suggestion, I've implemented that!

src/tests/integration/k-files/imp-simple-spec.k
geo2a: The claim you suggested (concrete-requires-trivial-false-nonsense-undecided( passes with both `False` and `1 == 2` requires, so it indeed seems like the consequent takes priority.

src/pyk/proof/equality.py
geo2a: It works! I've made that change.

src/tests/integration/k-files/imp-simple-spec.k
ehildenb: Hmmm, maybe we need one that looks more like:

```
    claim X => Y requires X ==Int Y +Int 1
```

ehildenb (approved):

src/pyk/proof/equality.py
```diff
@@ -146,13 +176,53 @@ def __init__(self, proof: EqualityProof) -> None:
...
+            # TODO: we should not be forced to include the dummy configuration in the antecedent and consequent
+            dummy_config = kcfg_explore.kprint.definition.empty_config(sort=GENERATED_TOP_CELL)
+            result = kcfg_explore.cterm_implies(
+                antecedent=CTerm(config=dummy_config, constraints=[antecedent_kast]),
+                consequent=CTerm(config=dummy_config, constraints=[consequent_kast]),
+            )
+            if result is None:
+                _LOGGER.warning('cterm_implies returned None, the implication is unsatisfiable')
+            else:
+                self.proof.set_csubst(result)
```
ehildenb, line 227: We can probably clean this up a bit by just:

- Always compute the simplified antecedent and consequent, but do not set `proof.csubst` right away.
- Get rid of intermediate variable `proof_trivially_passed`, and use as the guard on line 216 instead directly `if is_top(self.simplified_consequent) or is_bottom(self.simplified_antecedent): ...`, then in that `if` block set the `proof.csubst`, emit the warning, and return immediately.

ehildenb: Recorded: https://github.com/runtimeverification/pyk/issues/426

src/pyk/proof/equality.py
```diff
@@ -28,54 +31,58 @@ class EqualityProof(Proof):
...
         super().__init__(id, proof_dir=proof_dir)
         self.lhs_body = lhs_body
         self.rhs_body = rhs_body
         self.sort = sort
-        self.lhs_constraints = tuple(lhs_constraints)
-        self.rhs_constraints = tuple(rhs_constraints)
+        self.constraints = tuple(constraints)
+        self.csubst = csubst
+        self.simplified_constraints = simplified_constraints
         self.simplified_equality = simplified_equality
```
ehildenb, line 56: We can probably infer the `sort` from the lhs and rhs directly, by adding `KDefinition.sort(...)`. Recorded: https://github.com/runtimeverification/pyk/issues/426

src/tests/integration/k-files/imp-simple-spec.k
ehildenb: Recorded: https://github.com/runtimeverification/pyk/issues/426

PR #242, opened 2023-03-03
Migrate itegration tests to ephemeral runner

tothtamas28:

PR #421, opened 2023-05-11
Cleanup IMP proof definitions

ehildenb:
This PR cleans up the IMP proof definitions now that we have an `_is_terminal` predicate to use for them. Because of this, we don't need to artificially halt execution with a `#FINISH` rule.

- Remove the `#FINISH` production from `IMP-VERIFICATION`, remove the `IMP-VERIFICATION.halt` rule altogether, and from the test-suite.
- Rename a variable in the `IMP-SIMPLE-SPEC` that was unused to an anonymous variable `_:Bool`.
- Fix a bug in `rename_generated_vars` which was dropping the sort annotation on anonymous variables when renaming them. This is accomplished by allowing the user to specify a sort when calling `abstract_term_safely`.
- Disables using the ephemeral runners because they seem to not exist anymore...


.github/workflows/test-pr.yml
```diff
@@ -72,7 +72,7 @@ jobs:
   integration-tests:
     needs: code-quality-checks
     name: 'Integration Tests'
-    runs-on: [self-hosted, linux, normal-ephemeral]
```
tothtamas28, line 75: Restoring non-ephemeral runners requires that we move these tests back into a Docker image (see https://github.com/runtimeverification/pyk/pull/242).

I'd very much like to avoid this, to me it feels like losing progress that we have already achieved. But if this necessary for some reason, it needs to be taken care of in its own PR.

Why are these runners no longer available?

src/tests/integration/kcfg/test_imp.py
```diff
@@ -293,7 +293,7 @@
         'sum-10',
         None,
         None,
-        ['IMP-VERIFICATION.halt'],
```
tothtamas28, line 296: This value seems to be `[]` in all test cases, so the field itself can be removed from the data structure.

.github/workflows/test-pr.yml
tothtamas28: Hotfix: https://github.com/runtimeverification/pyk/pull/423

tothtamas28: Issue so that we're not stuck with the workaround: #424.

tothtamas28 (approved):

src/tests/integration/kcfg/test_imp.py
ehildenb: Done.

Issue #420, opened 2023-05-10
Refactor `pretty_print_kast` to use helper class


tothtamas28:
Extract a function `PrettyPrinter`:

```python3
class PrettyPrinter:
    symbol_table: SymbolTable

    def print(self, kast: KAst) -> str:
        match kast:
            case KToken():
                return self._print_ktoken(kast)
            case KApply():
                return self._print_kapply(kast)
            # ...
            case _:
                raise AssertionError()

    def _print_ktoken(self, kroken: KToken) -> str:
        ...

    def _print_kapply(self, kapply: KApply) -> str:
        ...
```

Then:

```python3
def pretty_print_kast(kast: KAst, symbol_table: SymbolTable) -> str:
    return PrettyPrinter(symbol_table).print(kast)
```

This enables customizing the default pretty printing logic by overloading cases.

PR #422, opened 2023-05-11
refactor pretty_print into PrettyPrinter class

anvacaru:
Fixes: https://github.com/runtimeverification/pyk/issues/420


src/pyk/ktool/kprint.py
```diff
@@ -387,180 +387,268 @@ def build_symbol_table(
     return symbol_table
 
 
-def pretty_print_kast(kast: KAst, symbol_table: SymbolTable) -> str:
```
tothtamas28, line 390: Don't delete this function, just replace implementation to use `PrettyPrinter`. Then changing all calls to it becomes unnecessary.

tothtamas28: You can call it there directly, sure. My point is not introducing a breaking change (at least in this PR).

tothtamas28 (approved):

Issue #378, opened 2023-04-20
Restrict back-edges in KCFGs


ehildenb:
A bug in how `extract_branches` heuristic led to some branches being automatically trivially proved because thhey had trivial 0-step loop-backs with `Split` edges https://github.com/runtimeverification/pyk/pull/377. We should add restrictions to the generated KCFGs so that this cannot happen (in additioon to the current restriction that each `Node` can have at most one `Successor`).

- Non-`Cover` edges are not allowed to go backwards. This likely isn't possible as a restriction, because a transition system which had `a => b => a` must have a back-loop via an `Edge`. So perhaps another restriction is:
- No zero-length loops. This would stop something where you have `a --> a` with a `Cover` or a `Split`, there must be symbolic execution involved. Another example where restricting zero-length loops would have stopped a bug earlier with a better message: https://github.com/runtimeverification/pyk/issues/398

tothtamas28:
> Non-`Cover` edges are not allowed to go backwards.

I strongly agree with this. In my opinion, the data structure we use to store symbolic execution state should be a tree w. r. t. the step relation (circularities can be encoded using covers). This restriction makes processing the data structure much easier, while the circular representation can still be derived.

>  This likely isn't possible as a restriction, because a transition system which had `a => b => a` must have a back-loop via an `Edge`

This can be implemented if the identity of a node is not equal to the term it stores. So either the hash of a node has to include the hash of the parent node, or the identity should not be the hash (it can be a unique integer for instance). Using this second solution, the hashes can still be used to e.g. search for nodes that store syntactically equal terms.

ehildenb:
I have pulled the second (simpler) change out here: https://github.com/runtimeverification/pyk/pull/428/files

I think it is still worth it to investigate the first change though. They are checking things that overlap, but coer different cases. With just the second check, you still can have a loop `a => b => a` with 1-step rewrites, and with just the second check you still can have a zero-length loop `a -split-> b -cover-> a`.

Perhaps we can somehow name nodes given their tree traversal order? So if we have:

```
a -> b
b -> c
b -> d
c -> a
d -> e
e -> a
```

Then we make nodes:

```
0: hash(a)
0.0: hash(b)
0.0.0: hash(c)
0.0.1: hash(d)
0.0.0.0: hash(a)
0.0.1.0: hash(e)
0.0.1.0.0: hash(a)
```

So that the node identifier is telling us where in teh tree it is (every node must be in the tree).

[...]

PR #428, opened 2023-05-11
Ensure no zero-length loops are created in KCFG

ehildenb:
Part of: https://github.com/runtimeverification/pyk/issues/378

This ensures that no zero-length loops are created in the KCFG by the user, which will catch all sorts of silly things that could happen when adding new edges to the KCFG.

- Function `path_length` is factored out of https://github.com/runtimeverification/pyk/pull/395.
- Function `shortest_path_between` is changed to use `path_length` to determine the shortest path, rather than just the number of successors.
- A check is added at all the edge-creation sites that the new edge will not create a zero-length loop.


tothtamas28 (approved):

src/pyk/kcfg/kcfg.py
```diff
@@ -917,3 +929,16 @@ def reachable_nodes(self, source_id: str, *, reverse: bool = False, traverse_cov
                 worklist.extend(succ.source for succ in self.predecessors(target_id=node.id, covers=traverse_covers))
 
         return visited
+
+
+def path_length(_path: Iterable[KCFG.Successor]) -> int:
+    _path = list(_path)
```
tothtamas28, line 935: A small optimization you can do here:

```suggestion
    _path = tuple(_path)
```

then if `_path` is a `tuple`, then no copying will be performed.

PR #430, opened 2023-05-12
Factor-out read_proof

geo2a:
Part of [evm-semantics#1775](https://github.com/runtimeverification/evm-semantics/issues/1775) and #394

This PR introduces a generic `read_proof` function that creates an object of the appropriate type (i.e. `APRProof`, `APRBMCProof` or `EqualityProof`) depending on the `'type'` JSON field of the serialized proof.

### DIFF
--- a/src/pyk/proof/equality.py
+++ b/src/pyk/proof/equality.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import json
 import logging
 from typing import TYPE_CHECKING
 
@@ -10,7 +9,6 @@ from ..kast.manip import extract_lhs, extract_rhs, flatten_label
 from ..prelude.k import GENERATED_TOP_CELL
 from ..prelude.kbool import BOOL, TRUE
 from ..prelude.ml import is_bottom, is_top, mlAnd, mlEquals
-from ..utils import hash_str
 from .proof import Proof, ProofStatus
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


ehildenb:
So, I don't think this quite accomplishes the goal. If you look at how `KInner.from_dict` works, it returns specifically something in the type of what's read, which is very convenient for type-checking. Are we able to mimic that structure better?

geo2a:
@ehildenb I think I've managed, could you have a look?

src/pyk/proof/proof.py
```diff
@@ -59,6 +61,22 @@ def dict(self) -> dict[str, Any]:
...
+
+        proof_path = proof_dir / f'{hash_str(id)}.json'
+        if Proof.proof_exists(id, proof_dir):
+            proof_dict = json.loads(proof_path.read_text())
+            proof_type = proof_dict['type']
+            _LOGGER.info(f'Reading {proof_type} from file {id}: {proof_path}')
+            if proof_type in Proof._PROOF_TYPES:
+                return locals()[proof_type].from_dict(proof_dict, proof_dir)
+
+        raise ValueError(f'Could not load APRBMCProof from file {id}: {proof_path}')
```
ehildenb, line 78:
```suggestion
        raise ValueError(f'Could not load Proof from file {id}: {proof_path}')
```

ehildenb:
@tothtamas28 what do you think of this approach? Can we do this any better, maybe using a type variable? It looks like here we will get runtime guarantee of the more specific proof type, but we won't get type-checking benefits as far as I can tell.

src/pyk/proof/proof.py
```diff
@@ -59,6 +61,22 @@ def dict(self) -> dict[str, Any]:
     def from_dict(cls: type[Proof], dct: Mapping[str, Any]) -> Proof:
         ...
 
+    @classmethod
+    def read_proof(cls: type[Proof], id: str, proof_dir: Path) -> Proof:
```
tothtamas28, line 65: Consider

```suggestion
    def read_proof(cls: type[P], id: str, proof_dir: Path) -> P:
```

where `P = TypeVar('P', Proof)`.

(This will also require `# type: ignore` for the return statement.)

src/pyk/proof/proof.py
```diff
@@ -59,6 +61,22 @@ def dict(self) -> dict[str, Any]:
...
+        # these local imports allow us to call .to_dict() based on the proof type we read from JSON
+        from .equality import EqualityProof  # noqa
+        from .reachability import APRBMCProof, APRProof  # noqa
```
tothtamas28, line 68: This solution works well if the class hierarchy is "sealed", i. e. you have control of all the subclasses.

If you need more extensibility, consider implementing some sort of registration mechanism, e.g. based on [`__init_subclass__`](https://docs.python.org/3/reference/datamodel.html#object.__init_subclass__).

ehildenb:
@geo2a is this subsumed by: https://github.com/runtimeverification/pyk/pull/485

Issue #319, opened 2023-03-30
Add ability to have pretty-printed nodes sort set and map items


nwatson22:
It would be nice to have the pretty printer in pyk print sets and maps in a definite order. This could be an optional feature if the performance cost of doing this sorting is high. This would improve readability, especially in the case of maps with integer keys, but also allow for less instability when comparing configurations. This idea came up since we are trying to use the KCFG show feature in kevm foundry to test that the KCFGs generated by our tests is not changing in structure when we make updates. Specifically, there is a cell `<activeAccounts>`, which contains a Set, whose items tend to get re-ordered unpredictably, and this can even cause edges to get written with spurious rewrites, like `<setCell> (SetItem(1) => SetItem(2)) (SetItem(2) => SetItem(1)) </setCell>`. See [this PR](https://github.com/runtimeverification/evm-semantics/pull/1663) for reference.

PR #433, opened 2023-05-15
Option to sort collections for KPrint.pretty_print

ehildenb:
Fixes: https://github.com/runtimeverification/pyk/issues/319
~Blocked on: https://github.com/runtimeverification/pyk/pull/466~

This is to address output stability issues when printing K terms as output for testing.

- Tests of sorted output are added, both of normal collections `_Set_`, and cell maps `_AccountCellMapItem_`.
- Functions `sort_assoc_label` and `sort_ac_collections` are added to `kast.manip`.
- Class `PrettyPrinter` gets a new field `_sort_collections`, which marks whether we should be sorting AC collections upon unparsing them. This option is threaded through to `KPrint.pretty_print`, and various places in `KCFGShow`.
- Fix a bug in `KCFGShow.to_module` where the source constraints were being used for both teh initial and target terms when building the rules in the module to display to the user. Factor out several helpers for minimizing configurations and printing here too.
- A new module `prelude.collections` is added with helpers for building things from K builtins regarding sets/lists/maps.


ehildenb:
I've checked with KEVM, and this appears to give more reasonable output.

- The rewrites on teh `<accounts>` cell are smaller (they can now be pushed down below each individual account, because the pre-state and post-state have the accounts in the same order, and we don't have the `AccountCellMapItem` placeholder in the way).
- The `requires` and `ensures` clauses of the rules are properly filled in.
- More of the variables stay present in the minimized output, because they show up in the requires/ensures clauses.

src/pyk/kast/manip.py
```diff
@@ -37,6 +37,38 @@ def flatten_label(label: str, kast: KInner) -> list[KInner]:
     return [kast]
 
 
+# TODO: don't duplicate logic from build_assoc
+def sort_assoc_label(label: str, kast: KInner) -> KInner:
+    res: KInner | None = None
+    if type(kast) is KApply and kast.label.name == label:
+        terms = sorted(flatten_label(label, kast))
+        assert len(terms) > 0
+        for term in reversed(list(terms)):
```
tothtamas28, line 46:
```suggestion
        for term in reversed(terms):
```

src/pyk/kast/manip.py
```diff
@@ -37,6 +37,38 @@ def flatten_label(label: str, kast: KInner) -> list[KInner]:
     return [kast]
 
 
+# TODO: don't duplicate logic from build_assoc
+def sort_assoc_label(label: str, kast: KInner) -> KInner:
+    res: KInner | None = None
+    if type(kast) is KApply and kast.label.name == label:
+        terms = sorted(flatten_label(label, kast))
+        assert len(terms) > 0
```
tothtamas28, line 45: I think this should be a `ValueError` instead.

src/pyk/kast/manip.py
```diff
@@ -37,6 +37,38 @@ def flatten_label(label: str, kast: KInner) -> list[KInner]:
     return [kast]
 
 
+# TODO: don't duplicate logic from build_assoc
```
tothtamas28, line 40: Can this be easily addressed in this PR?

src/pyk/kcfg/show.py
```diff
@@ -325,20 +337,14 @@ def _hide_cells(_k: KInner) -> KInner:
...
-                        init_cterm.add_constraint(c)
-                target_cterm = CTerm(hide_cells(edge.target.cterm.config), ())
-                for c in edge.source.cterm.constraints:
-                    assert type(c) is KApply
-                    if c.label.name == '#Ceil':
-                        _LOGGER.warning(f'Ignoring Ceil condition: {c}')
-                    else:
-                        target_cterm.add_constraint(c)
+                init_constraints = [
+                    c for c in edge.source.cterm.constraints if not (type(c) is KApply and c.label.name == '#Ceil')
```
tothtamas28, line 341:
```suggestion
                    c for c in edge.source.cterm.constraints if type(c) is not KApply or c.label.name != '#Ceil'
```

I personally prefer this variant, but the original is fine too.

src/pyk/kcfg/show.py
```diff
@@ -325,20 +337,14 @@ def _hide_cells(_k: KInner) -> KInner:
...
+                ]
+                init_cterm = CTerm(simplify_config(edge.source.cterm.config), init_constraints)
+                target_constraints = [
+                    c for c in edge.target.cterm.constraints if not (type(c) is KApply and c.label.name == '#Ceil')
```
tothtamas28, line 345:
```suggestion
                    c for c in edge.target.cterm.constraints if type(c) is not KApply and c.label.name != '#Ceil'
```

Consider extracting a helper function for the condition (maybe `_not_ceil`).

src/pyk/ktool/kprint.py
```diff
@@ -313,9 +313,13 @@ def _add_sort_injection(self, pat: Pattern, isort: KSort, osort: KSort) -> Patte
...
-    def pretty_print(self, kast: KAst, *, unalias: bool = True) -> str:
-        if unalias and isinstance(kast, KInner):
-            kast = undo_aliases(self.definition, kast)
+    def pretty_print(self, kast: KAst, *, unalias: bool = True, sort_collections: bool = False) -> str:
+        # TODO: Should apply KInner transformations inside KRule and KClaim as well.
+        if isinstance(kast, KInner):
+            if unalias:
+                kast = undo_aliases(self.definition, kast)
+            if sort_collections:
+                kast = sort_ac_collections(self.definition, kast)
```
tothtamas28, line 322: You can also implement this by subclassing `PrettyPrinter` (I think that would also clear the TODO):

```python3
class SortingPrinter(PrettyPrinter):
    ...

    def _print_kapply(self, kapply: KApply) -> str:
        kapply = sort_ac_collections(self.definition, kast)
        return super().__print_kapply(kapply)
```

Here, `sort_ac_collections` is not recursive.

src/tests/integration/test_kprint.py
```diff
@@ -306,3 +306,57 @@ def test_aliases(self, kprint: KPrint, test_id: str, kast: KInner, expected: str
     def test_nonterm_labels(self, kprint: KPrint, test_id: str, kast: KInner, expected: str) -> None:
         actual = kprint.pretty_print(kast)
         assert actual == expected
+
+
+SORT_COLLECTIONS_TEST_DATA: Iterable[tuple[str, KInner, str]] = (
+    (
+        'single-item',
+        KApply('SetItem', [intToken(1)]),
```
tothtamas28, line 314: You can omit the list for the arguments (for all instances). Also, consider using `token`:

```suggestion
        KApply('SetItem', token(1)),
```

tothtamas28: We need helpers for these collections in `pyk.prelude`.

src/pyk/kast/manip.py
ehildenb: Hmmmm, I guess I'm not sure why, because this case should literally never happen. `flatten_label` is guaranteed to return at least one element, so this is more for type-checking than anything. I actually just removed this assertion, because the one at the end of the function about `res is not None` subsumes it.

src/pyk/kast/manip.py
ehildenb: Done.

src/pyk/kast/manip.py
ehildenb: I guess not, there is some divergences in their behavior now. I'v just removed the TODO.

src/pyk/kcfg/show.py
ehildenb: I factored out a helper.

src/pyk/kcfg/show.py
ehildenb: Done.

src/tests/integration/test_kprint.py
ehildenb: Done.

nwatson22 (approved):

nwatson22 (approved):

PR #438, opened 2023-05-17
Add pretty printing of nonterminal labels and test

### LINKED ISSUE #437: Unparse record productions with the named non-terminals
Now in compiled.txt we'll have the names: https://github.com/runtimeverification/k/pull/3420#pullrequestreview-1429010098

But we already have the field `name: ...` in the `KNonTerminal` nodes in the JSON (emitted [here](https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/unparser/ToJson.java#L385)). We need to just read it [here](https://github.com/runtimeverification/pyk/blob/57be0b08f48b56e882a4a815c07d42bb76071963/src/pyk/kast/outer.py#L192), and use it for unparsing [here](https://github.com/runtimeverification/pyk/blob/57be0b08f48b56e882a4a815c07d42bb76071963/src/pyk/ktool/kprint.py#L376).

nwatson22:
Closes #437.
Currently we use the same `pretty_print` function both for our normal user-facing printing but also for printing claims into k files. This causes issues because the parser for claims does not recognize the nonterminal labels. The approach I'm using right now to solve this problem is to have a flag for the `KPrint` object for whether or not the labels should be printed, and it's off by default.


src/pyk/kast/outer.py
```diff
@@ -184,14 +184,17 @@ def let(
...
 
     @classmethod
     def from_dict(cls: type[KNonTerminal], d: Mapping[str, Any]) -> KNonTerminal:
         cls._check_node(d)
-        return KNonTerminal(sort=KSort.from_dict(d['sort']))
+        name = d['name'] if 'name' in d else None
+        return KNonTerminal(sort=KSort.from_dict(d['sort']), name=name)
 
     def to_dict(self) -> dict[str, Any]:
         return {'node': 'KNonTerminal', 'sort': self.sort.to_dict()}
```
ehildenb, line 200: We should also modify `KNonTerminal.to_dict` and `KNonTerminal.let` to accept the `name` parameter.

src/pyk/ktool/kprint.py
```diff
@@ -190,6 +190,7 @@ class KPrint:
     main_module: str
     backend: str
     _extra_unparsing_modules: Iterable[KFlatModule]
+    nonterm_labels: bool
```
ehildenb, line 193: Let's not bother making it an option. Instead, let's _always_ print out nonterminal names.

src/tests/integration/k-files/imp.k
```diff
@@ -20,7 +20,7 @@ module IMP-SYNTAX
                  | Id "=" AExp ";"            [klabel(_=_;), symbol, strict(2), color(pink), format(%1 %2 %3%4)]
                  | "if" "(" BExp ")"
                    Block "else" Block         [klabel(if(_)_else_), symbol, strict(1), colors(yellow, white, white, yellow), format(%1 %2%3%4 %5 %6 %7)]
-                 | "while" "(" BExp ")" Block [klabel(while(_)_), symbol, colors(yellow,white,white), format(%1 %2%3%4 %5)]
+                 | "while" "(" cond:BExp ")" Block [klabel(while(_)_), symbol, colors(yellow,white,white), format(%1 %2%3%4 %5)]
```
ehildenb, line 23: Hmmm, I would prefer to modify a different definition than this one. Or make a standalone definition.

But here is `k-files/aliases.k` which is a very simple definition you could add a new production to, and it's already wired up to `test_kprint.py`, so not too much effort to make another test for that one.

src/pyk/kast/outer.py
nwatson22: I added these.

src/pyk/ktool/kprint.py
nwatson22: I removed the option. I had thought it was necessary but everything works with it always on when we add `...` before the first named nonterminal.

src/tests/integration/k-files/imp.k
nwatson22: I am now testing with a production inside `aliases.k` instead.

ehildenb: Let's remove this change then.

ehildenb:
```suggestion
                 | "while" "(" BExp ")" Block [klabel(while(_)_), symbol, colors(yellow,white,white), format(%1 %2%3%4 %5)]
```

src/pyk/ktool/kprint.py
```diff
@@ -362,10 +366,16 @@ def unparser_for_production(prod: KProduction) -> Callable[..., str]:
...
+        num_nonterm = len([item for item in prod.items if type(item) is KNonTerminal])
+        num_named_nonterm = len([item for item in prod.items if type(item) is KNonTerminal and item.name != None])
         for item in prod.items:
             if type(item) is KTerminal:
                 result.append(item.value)
             elif type(item) is KNonTerminal and index < len(args):
+                if index == 0 and num_nonterm == num_named_nonterm and num_nonterm > 0:
+                    result.append('...')
+                if num_nonterm == num_named_nonterm and num_nonterm > 0:
+                    result.append(f'{item.name}:')
```
ehildenb, line 378: I guess we don't need to check `num_nonterm > 0` here, because we're inside an iteration of the loop where `type(item) is KNonTerminal`.

```suggestion
                if index == 0 and num_nonterm == num_named_nonterm:
                    result.append('...')
                if num_nonterm == num_named_nonterm:
                    result.append(f'{item.name}:')
```

src/tests/integration/test_kprint.py
```diff
@@ -249,10 +249,45 @@ def test_init_config(self, kprint: KPrint, test_id: str, sort: KSort, expected:
...
+
 
 class TestAliasDefn(KPrintTest):
     KOMPILE_MAIN_FILE = K_FILES / 'aliases.k'
 
+    @staticmethod
+    def _update_symbol_table(symbol_table: SymbolTable) -> None:
+        symbol_table['_,_'] = assoc_with_unit(' , ', '')
+        symbol_table['.List{"_,_"}'] = lambda: ''
+
```
ehildenb, line 290: Why is this change needed?

ehildenb (approved):

PR #440, opened 2023-05-17
kore/tools: kore_print accepts a str as well

ehildenb:
This enables us to skip parsing Kore patterns when the user already has a kore string in hand and calls `kore-print` on it.


src/pyk/kore/tools.py
```diff
@@ -24,14 +19,17 @@ class PrintOutput(Enum):
...
-def kore_print(pattern: Pattern, definition_dir: str | Path, output: str | PrintOutput) -> str:
+def kore_print(pattern: str | Pattern, definition_dir: str | Path, output: str | PrintOutput) -> str:
     definition_dir = Path(definition_dir)
     check_dir_path(definition_dir)
 
     output = PrintOutput(output)
 
     with NamedTemporaryFile(mode='w') as f:
-        pattern.write(f)
+        if type(pattern) is Pattern:
```
tothtamas28, line 29:
```suggestion
        if isinstance(pattern, Pattern):
```

src/pyk/kore/tools.py
```diff
@@ -24,14 +19,17 @@ class PrintOutput(Enum):
...
+            pattern.write(f)
+        elif type(pattern) is str:
+            f.write(pattern)
         f.write('\n')
```
tothtamas28, line 33: I'd do this only for the `Pattern` branch, there we know there's no terminating `\n`.

src/pyk/kore/tools.py
```diff
@@ -24,14 +19,17 @@ class PrintOutput(Enum):
...
     definition_dir = Path(definition_dir)
     check_dir_path(definition_dir)
 
     output = PrintOutput(output)
 
     with NamedTemporaryFile(mode='w') as f:
-        pattern.write(f)
+        if type(pattern) is Pattern:
+            pattern.write(f)
+        elif type(pattern) is str:
```
tothtamas28, line 31: Consider adding an `else` branch with a `TypeError`.

tothtamas28 (approved):

src/pyk/kore/tools.py
ehildenb: Done.

src/pyk/kore/tools.py
ehildenb: Done.

src/pyk/kore/tools.py
ehildenb: Done.

Issue #43, opened 2022-04-26
`kcovr` should use `pyk` and proper argument parsing


ehildenb:
Currently `kcovr` is just a flat Python script.

It should use the pyk libraries functionality for handling definitions, then we wouldn't need `allRules.txt`.

It should also use the same argument parsing logic as pyk.

ehildenb:
You can change wehre the coverage files are dumped: https://github.com/runtimeverification/llvm-backend/pull/250

ehildenb:
This PR introduces `kcovr` binary implemented using pyk library. We should consider deleting the one that the frontend exports.

PR #441, opened 2023-05-17
Add and refactor `kcovr`

tothtamas28:
Related: #43

Fixes linting issues, adds type hints, extracts functions and constants, refactors some parts into simpler code.


src/pyk/kcovr.py
```diff
@@ -0,0 +1,264 @@
...
+        for i, x in enumerate(xs):
+            if x == '++':
+                return xs[:i], xs[i + 1 :]
+        return xs, []
+
+    definition_strs, source_strs = split_at_sep(sys.argv[1:])
+    definition_dirs = tuple(dir_path(s).resolve() for s in definition_strs)
+    source_files = tuple(file_path(s).resolve() for s in source_strs)
+
+    return definition_dirs, source_files
```
ehildenb, line 73: Are we doing manual argument parsing on purpose? Is it because the current interface of `kcovr` won't allow for `argparse`?

ehildenb (approved):

src/pyk/kcovr.py
tothtamas28: Yes, unfortunately, separating two `nargs='+'` arguments with `--` doesn't work as intended in `argparse`.

PR #447, opened 2023-05-22
Change `KCFG.Node.id` from hash `str` to `int`

tothtamas28:
Related: #378

* Changes type of `Node.id` to `int`.
* Adds type alias `NodeIdLike = int | str`, representing an id or an alias, respectively.
* Changes `dict` representation of `splits` and `ndbranches`.
* Drops `get_or_create_node`. A node containing the same `CTerm` as an already existing node can be created without restriction.
* Drops parameter `omit_node_hash` from functions that display the KCFG or its nodes.


tothtamas28:
Tested with KEVM-Foundry (after fixes):
* `make test-foundry-prove`
* `kevm foundry-list`
* `kevm foundry-show`
* `kevm foundry-view-kcfg`

src/pyk/kcfg/kcfg.py
```diff
@@ -355,40 +358,40 @@ def to_json(self) -> str:
...
         return node_strs
 
-    def _resolve_hash(self, id_like: str) -> list[str]:
-        return [node_id for node_id in self._nodes if compare_short_hashes(id_like, node_id)]
-
     def get_unique_init(self) -> Node:
         if len(self.init) > 1:
-            raise ValueError(f'Multiple init nodes found: {[shorten_hash(n.id) for n in self.init]}')
+            raise ValueError(f'Multiple init nodes found: {[n.id for n in self.init]}')
         return self.init[0]
```
ehildenb, line 373: Should we just do: `return single(n.id for n in self.init)` instead?

This usually results in a good enough error message, especially with the stack trace.

src/pyk/kcfg/kcfg.py
```diff
@@ -355,40 +358,40 @@ def to_json(self) -> str:
...
 
     def get_unique_target(self) -> Node:
         if len(self.target) > 1:
-            raise ValueError(f'Multiple target nodes found: {[shorten_hash(n.id) for n in self.target]}')
+            raise ValueError(f'Multiple target nodes found: {[n.id for n in self.target]}')
         return self.target[0]
```
ehildenb, line 378: Same here, should we do the simpler return with using `single`?

src/pyk/kcfg/kcfg.py
```diff
@@ -400,49 +403,43 @@ def _resolve_or_none(self, id_like: str) -> str | None:
...
         cterm = CTerm.from_kast(term)
-        node = KCFG.Node(cterm)
-
-        if node.id in self._nodes:
-            raise ValueError(f'Node already exists: {node.id}')
-
+        node = KCFG.Node(self._next_id, cterm)
+        self._next_id += 1
         self._nodes[node.id] = node
         return node
```
ehildenb, line 440: Perhaps we should expose a property `next_id`, which we use for access everywhere, and it automatically increments `_next_id`:

```
@property
def next_id(self) -> int:
    self._next_id += 1
    return self._next_id
```

Would that be helpful for ensuring that `_next_id` always stays fresh? I guess this is probably one of the only places it's accessed, but it could make it easier to avoid someone using it incorrectly later.

src/pyk/kcfg/kcfg.py
```diff
@@ -265,15 +272,16 @@ def to_dict(self) -> dict[str, Any]:
         nodes = [node.to_dict() for node in self.nodes]
         edges = [edge.to_dict() for edge in self.edges()]
         covers = [cover.to_dict() for cover in self.covers()]
-        splits = dict(sorted((k, s.to_dict()) for k, s in self._splits.items()))
-        ndbranches = dict(sorted((k, b.to_dict()) for k, b in self._ndbranches.items()))
+        splits = [split.to_dict() for split in self.splits()]
+        ndbranches = [ndbranch.to_dict() for ndbranch in self.ndbranches()]
```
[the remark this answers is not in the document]
tothtamas28: I think we didn't need it in the first place. (This current solution is analogous to how `edges` and `covers` is defined.)

Anyways, if it turns out to cause instability in proof hashes (during testing I didn't experience this), we'll know where to fix.

src/pyk/kcfg/tui.py
```diff
@@ -278,30 +278,39 @@ def on_graph_chunk_selected(self, message: GraphChunk.Selected) -> None:
...
+            split = single(self._kcfg.splits(source_id=source_id, target_id=target_id))
             self.query_one('#node-view', NodeView).update(split)
 
         elif message.chunk_id.startswith('ndbranch_'):
             self._selected_chunk = None
             node_source, node_target, *_ = message.chunk_id[8:].split('_')
-            ndbranch = single(self._kcfg.ndbranches(source_id=node_source, target_id=node_target))
+            source_id = int(node_source)
+            target_id = int(node_target)
+            ndbranch = single(self._kcfg.ndbranches(source_id=source_id, target_id=target_id))
```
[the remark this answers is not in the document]
tothtamas28: Yes.

src/pyk/kcfg/kcfg.py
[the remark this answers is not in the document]
ehildenb: Ok, sounds good.

src/pyk/kcfg/kcfg.py
tothtamas28: Good point, done.

src/pyk/kcfg/kcfg.py
tothtamas28: Done.

src/pyk/kcfg/kcfg.py
tothtamas28: Yes, that's a good idea.

I ended up implementing it with an `Iterator[int]`: the only way to get an `int` out of it is to call `next`, which ensures the same number cannot be obtained from it twice.

PR #454, opened 2023-05-24
Create `pytest` plugin

tothtamas28:
Defines a module `pyk.testing` that publishes classes and fixtures used in the `pyk` test suite for downstream projects.


tothtamas28:
The plugin is also listed when running tests:

```
% make test-integration
poetry install
Installing dependencies from lock file

No dependencies to install or update

Installing the current project: pyk (0.1.303)
poetry run pytest src/tests/integration --maxfail=1 --verbose --durations=0 --numprocesses=4 --dist=worksteal
===================================================================== test session starts ======================================================================
platform linux -- Python 3.11.2, pytest-7.2.2, pluggy-1.0.0 -- /home/ttoth/.cache/pypoetry/virtualenvs/pyk-YghnWhM2-py3.11/bin/python
cachedir: .pytest_cache
rootdir: /home/ttoth/git/pyk
plugins: mock-3.10.0, cov-4.0.0, xdist-3.2.1, pyk-0.1.303
```

src/pyk/testing/_kompiler.py
```diff
@@ -0,0 +1,143 @@
...
+
+from typing import TYPE_CHECKING
+
+import pytest
+
+from ..kast.outer import read_kast_definition
+from ..kcfg import KCFGExplore
+from ..kllvm.compiler import compile_runtime
+from ..kllvm.importer import import_runtime
+from ..kore.rpc import KoreClient, KoreServer
```
ehildenb, line 11: Why is this module named `_kompiler`? Is it just for lack of a better name?

Maybe it should be named something like `test` or `pyktest`?

ehildenb (approved):

src/pyk/testing/_kompiler.py
tothtamas28: It defines a class `Kompiler` which is used by all the `*Test` classes.

But it's a private module. To make things as simple as possible for dependent projects, everything is imported in the `pyk.testing` module, so the intended use is:

```python3
from pyk.testing import KompiledTest
```

PR #455, opened 2023-05-24
kprove.py: Pass Haskell RTS options through GHCRTS environment variable

Scott-Guest:
Previously in #451, `KProve.prove(...)` was modified to pass all haskell backend arguments through `--haskell-backend-command` in order to support RTS options.

However, this is problematic because `--debugger` cannot be used with `--haskell-backend-command`.

To resolve this, we partially revert #451, using `KORE_EXEC_OPTS` for the common case and ~~only falling back to `--haskell-backend-command` when RTS options are actually provided~~ using `GHCRTS` for RTS options.


src/pyk/ktool/kprove.py
```diff
@@ -203,8 +204,15 @@ def prove(
...
-        _LOGGER.debug(f'haskell_backend_command={haskell_backend_command}')
+        kore_exec_opts = ' '.join(list(haskell_args) + haskell_log_args)
+        _LOGGER.debug(f'export KORE_EXEC_OPTS={kore_exec_opts!r}')
+        env = os.environ.copy()
+        env['KORE_EXEC_OPTS'] = kore_exec_opts
+
+        haskell_backend_command = None
+        if haskell_rts_args:
+            haskell_backend_command = f'kore-exec +RTS {" ".join(list(haskell_rts_args))} -RTS'
+            _LOGGER.debug(f'--haskell-backend-command={haskell_backend_command}')
```
jberthold, line 215: I just stumbled over this change (and its history)... 
* Maybe we should [consider using the `GHCRTS` environment variable](https://downloads.haskell.org/ghc/latest/docs/users_guide/runtime_control.html#rts-options-environment) to pass RTS options  
* It appears as though the program name `kore-exec` became hard-wired in an earlier change. Might be better if it was still possible to use a binary with a different name and which might not be on the path...

Scott-Guest: Ah I wasn't aware of `GHCRTS`, thank you, that's much cleaner!

I'll submit something separately to un-hardcode `kore-exec`

PR #457, opened 2023-05-25
(fix) path inference

iFrostizz:
When using `foundry-prove`, the process is crashing because for some reason, Python is converting the `Path` variables to a `str` and applies the division operator `/` rather than concatenating the path.
This fixes it by enforcing the Path type.


src/pyk/ktool/kprint.py
```diff
@@ -208,9 +208,9 @@ def __init__(
         self.use_directory = use_directory
         self._definition = None
         self._symbol_table = None
-        with open(self.definition_dir / 'mainModule.txt') as mm:
+        with open(Path(self.definition_dir) / 'mainModule.txt') as mm:
```
tothtamas28, line 211: This is not the right fix: `self.definition_dir` is already supposed to be a `Path`. If you see errors during runtime raised from here, that's because `KPrint` got the wrong type of argument for initialization (or less likely, the attribute got overridden after initialization).

My guess is the bug is in a command line interface, as that's a boundary where we assume types in function signatures. But we need to root-cause this. Can you give reproduction steps for the errors you see?

iFrostizz: In the `evm-semantics` repo, clean any foundry artifacts:
`make foundry-clean`
Then, compile the contracts
`foundry build --root tests/foundry`
And run a kevm kompile
`poetry run -C kevm-pyk/ kevm foundry-kompile --foundry-project-root tests/foundry/`

image

tothtamas28: I cannot reproduce this on `kevm 1.0.182` (source build):

```
% make foundry-clean
rm -rf tests/foundry/cache
rm -rf tests/foundry/out
rm -f  tests/foundry/foundry.debug-log
rm -f  tests/foundry/foundry.k
rm -f  tests/foundry/foundry.rule-profile

% forge build --root tests/foundry
[⠊] Compiling...
[⠰] Compiling 58 files with 0.8.13
[⠒] Solc 0.8.13 finished in 3.26s
Compiler run successful with warnings:
...

% poetry run -C kevm-pyk/ kevm foundry-kompile --foundry-project-root tests/foundry/
WARNING 2023-05-25 20:42:42,547 __main__ - Ignoring command-line option: -O0
WARNING 2023-05-25 20:42:42,547 __main__ - Ignoring command-line option: -O1
WARNING 2023-05-25 20:42:42,547 __main__ - Ignoring command-line option: -O2
WARNING 2023-05-25 20:42:42,547 __main__ - Ignoring command-line option: -O3
[Warning] Compiler: Could not find main syntax module with name
FOUNDRY-CONTRACTS in definition.  Use --syntax-module to specify one. Using
FOUNDRY-MAIN as default.
```

What version did you try?

iFrostizz: Looks like my kevm was outdated, sorry !

tothtamas28: No worries, thank you for the submission.

Issue #458, opened 2023-05-25
add key bindings capabilities to the kcfg-viewer


iFrostizz:
Currently, the kcfg viewer is entirely mouse-driven It would be a nice change to be able to run through this viewer and be able to move using only the keyboard.

Some potential changes ideas:

<kbd>h</kbd>,<kbd>j</kbd>,<kbd>k</kbd>,<kbd>l</kbd>  move in any tab (select next node, scroll down in a window, ...)
<kbd>Ctrl</kbd> + <kbd>d</kbd> go down a page
<kbd>Ctrl</kbd> + <kbd>u</kbd> go up a page
<kbd>Ctrl</kbd> + <kbd>w</kbd> trigger a window change (<kbd>h</kbd>,<kbd>j</kbd>,<kbd>k</kbd>,<kbd>l</kbd>) to actually move
By typing a number, we could store it in a buffer and replicate `x` times the move action.

UI suggestion by @anvacaru keep the highlight focus on the node that is currently selected, even when the mouse isn't hovering it anymore (same using the keyboard)
![image](https://github.com/runtimeverification/pyk/assets/51274081/57b6be93-cce8-4267-ac4f-78cc7db63533)

iFrostizz:
maybe also replace `h` by `f` for fold as it will interfere with the "go left" key binding

iFrostizz:
Mostly done by https://github.com/runtimeverification/pyk/pull/578
Can be reopened if missing features are requested

PR #460, opened 2023-05-25
Add key bindings to the kcfg-viewer

iFrostizz:
Closes: https://github.com/runtimeverification/pyk/issues/458

## TODO

- [x] scroll in nodes with "jk"
- [x] keep node highlight when selected
- [ ] change window
- [ ] naviguate in source and constraint

### DIFF
--- a/src/pyk/kcfg/tui.py
+++ b/src/pyk/kcfg/tui.py
@@ -52,7 +52,7 @@ class GraphChunk(Static):
 
     async def on_click(self, click: Click) -> None:
         await self.emit(GraphChunk.Selected(self, self.id or ''))
-        self.styles.border_left = ('double', 'red')  # type: ignore
+        # self.styles.border_left = ('double', 'red')  # type: ignore
         click.stop()
 
 
@@ -161,14 +161,20 @@ class NodeView(Widget):
         else:
             self.query_one(f'#{field}-view', Horizontal).add_class('hidden')
 
-    def update(self, element: KCFGElem) -> None:
+    def update(self, element: KCFGElem, style: bool = False) -> None:
         self._element = element
+        # if (style):
+        #     self._focus_style()
         self._update()
 
-    def _focus_style(self, element) -> None:
-        self.query_one('#node-view', NodeView).set_styles('border-left: double, red')
+    def _focus_style(self) -> None:
+        # self.query_one('#node-view', NodeView).set_styles('border-left: double, red')
         # self.styles.border_left = None  # type: ignore
-        self.styles.border_left = ('double', 'red')  # type: ignore
+        # self.styles.border_left = ('double', 'red')  # type: ignore
+        if self._element and type(self._element) is KCFG.Node:
+            selected_chunk = 'node_' + str(self._element.id)
+            # self.query_one(f'#{selected_chunk}', GraphChunk).set_styles('border-left: double, red')
+            # self.query_one(f'#{selected_chunk}', GraphChunk)
 
     def _update(self) -> None:
         def _boolify(c: KInner) -> KInner:
@@ -279,6 +285,7 @@ class KCFGViewer(App):
         self._hidden_chunks = []
         self._buffer = []
         self._last_node_idx = 0
+        # TODO: only take cut nodes
         self._node_ids = list(kcfg._nodes.keys())
         self._node_idx = {} # TODO: one-liner for filling this
 
@@ -295,9 +302,6 @@ class KCFGViewer(App):
             self._selected_chunk = None
             self._last_idx = 0
 
-        # self.query_one('#node-view', NodeView).update(self._kcfg.node(node_id))
-        # TODO: update the first node to default to this one
-
     def compose(self) -> ComposeResult:
         yield Horizontal(
             Vertical(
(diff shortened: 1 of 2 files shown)


ehildenb:
Please post a gif with the changes.

Also, instead of changing the boarder on the left of a selected node, perhaps we change the background color, which makes it easier to see without changing the look of the graph.

iFrostizz:
Note: this call is especially quite slow, which makes the app launch proportionally longer as there is more node.
https://github.com/runtimeverification/pyk/blob/d07d63cf75f1981e484538f39bfc261af19ab605/src/pyk/kcfg/tui.py#L286

src/pyk/kcfg/tui.py
```diff
@@ -43,10 +45,10 @@ def __init__(self, id: str, node_text: Iterable[str] = ()) -> None:
         super().__init__(self._node_text, id=id, classes='cfg-node')
 
     def on_enter(self) -> None:
-        self.styles.border_left = ('double', 'red')  # type: ignore
+        self.styles.text_opacity = '75%'  # type: ignore
 
     def on_leave(self) -> None:
-        self.styles.border_left = None  # type: ignore
+        self.styles.text_opacity = '100%'  # type: ignore
```
ehildenb, line 51: Do we still need the `# type: ignore` here?

src/pyk/kcfg/tui.py
```diff
@@ -259,74 +302,268 @@ def __init__(
...
+                if kind == MoveKind.BOUND:
+                    self.query_one(f'#{view}').scroll_page_right(animate=False)
+                else:
+                    self.query_one(f'#{view}').scroll_right(animate=False)
+
+    async def action_keystroke(self, key: str) -> None:
+        if key in ['h', 'j', 'k', 'l']:
+            dir = Direction.dir_of(key)
+            if dir is not None:
+                if self._buffer and self._buffer[-1] in ['change-window']:
```
ehildenb, line 532: It seems that `_buffer` is _only_ used for marking that we're in window changing mode, not for anything else.

Instead, I would have this be called `_mode`, and have it be an `Enum`, so that we have:

```
class MovementMode(Enum):
    SCROLL = 'scroll'
    WINDOW = 'window'
```

Then store that in `_mode` variable, and check that variable to determine if the user is scrolling or changing which window is in focus. Later perhaps we can add more modes. But since we're only using the `_buffer` for marking this one mode, we might as well simplify it.

src/pyk/kcfg/tui.py
```diff
@@ -259,74 +302,268 @@ def __init__(
...
+        self.query_one(f'#{to.value}').set_class(True, 'selected')
+        self._curr_win = to
+
+    def move(self, dir: Direction, kind: MoveKind) -> None:
+        view = self._curr_win.value
+        match dir:
+            case Direction.LEFT:
+                if kind == MoveKind.BOUND:
+                    self.query_one(f'#{view}').scroll_page_left(animate=False)
+                else:
```
ehildenb, line 445: This method is pretty huge and hard to read. Can we extract helpers? Or perhaps we can do something like this:

```
match (dir, kind, self._curr_win):
    case (Direction.LEFT, MoveKind.BOUND, _):
        ....
    ....
    case (Direction.DOWN, Window.BEHAVIOR, Window.TERM | Window.CUSTOM | Window.CONSTRAINT):
        ....
....
```

That way it's much reduced nesting on the branching structure, and maybe a bit easier to read and understand.

src/pyk/kcfg/tui.py
```diff
@@ -259,74 +302,268 @@ def __init__(
...
+                            elif self._curr_win == Window.CONSTRAINT:
+                                self.focus_window(Window.CUSTOM)
+                        case Direction.UP:
+                            if self._curr_win == Window.CUSTOM:
+                                self.focus_window(Window.CONSTRAINT)
+                            elif self._curr_win == Window.CONSTRAINT:
+                                self.focus_window(Window.TERM)
+                        case Direction.RIGHT:
+                            if self._curr_win == Window.BEHAVIOR:
+                                self.focus_window(Window.TERM)
```
ehildenb, line 549: Similarly here, can we do a tuple match statement on `match (dir, self._curr_win):` instead of nested statements?

src/pyk/kcfg/tui.py
iFrostizz: Good suggestion, I initially called it buffer to make it support multiple moves at once through 3j, 5k, ...
Should we rename it for this PR and include the multi move with the variable renamed to buffer in a later one ?

src/pyk/kcfg/tui.py
iFrostizz: Nice catch!

src/pyk/kcfg/tui.py
iFrostizz: Sadly this syntax doesn't exists, I will create helpers

src/pyk/kcfg/tui.py
ehildenb: I don't think we'll need to implement all the movement features of vim, so probably just call it an enum as suggested. If we want ot support numbers later, that can be a separate variable.

ehildenb:
It looks like default keybinding actions already exist, can we just use those? https://textual.textualize.io/api/containers/#textual.containers.ScrollableContainer

PR #464, opened 2023-05-25
Implement very simple sort parameter inference

ehildenb:
Related to: https://github.com/runtimeverification/k/issues/3436

This implements very simple sort parameter inference for the Kast -> Kore conversion in pyk. This inference does not attempt to be clever or smart in any way, it just handles the simplest case:

- All the arguments in parametric sort positions can have their sort determined directly.
- All the arguments in a sort parametric position get the same sort as any other occurrences of that sort parameter.
- All sort parameters get assigned a concrete sort.

This PR:

- Adds two tests of this failing (specifically on `#ite` production), which were crashing before.
- Adds `KDefinition.sort: KInner -> KSort | None`, which best effort returns the sort of the supplied term.
- Makes a small refactoring to `KDefinition.sort_vars_subst` to call out the simple method `greatest_common_subsort`.
- Changes `kast_to_kore` to first call `KDefinition.add_sort_parameters` before converting to Kore.


ehildenb:
In KEVM, running `make tests/specs/function/infinite-gas-spec.k.prove` results in 11 calls to `kast` binary (as fallback or `kast_to_kore`).

image

They range from 1.5s to 7s, with most around 2.5s.

After this PR, there are no remaining calls to `kast` binary (everything can be handled by `kast_to_kore`).

src/tests/integration/konvert/test_simple_proofs.py
```diff
@@ -283,6 +283,49 @@
...
+        )
+        """,
+        KApply(
+            KLabel('#if_#then_#else_#fi_K-EQUAL-SYNTAX_Sort_Bool_Sort_Sort', []),
+            [
+                KVariable('C', sort=KSort('Bool')),
+                KVariable('B1', sort=KSort('K')),
+                KVariable('B2', sort=KSort('K')),
+            ],
+        ),
```
tothtamas28, line 301: You can make these more compact:

```suggestion
        KApply(
            KLabel('#if_#then_#else_#fi_K-EQUAL-SYNTAX_Sort_Bool_Sort_Sort'),
            KVariable('C', sort='Bool'),
            KVariable('B1', sort='K'),
            KVariable('B2', sort='K'),
        ),
```

src/pyk/kast/outer.py
```diff
@@ -1078,6 +1078,33 @@ def _compute_subsorts(self, sort: KSort) -> list[KSort]:
                 _subsorts.extend([_subsort] + self.subsorts(prod.items[0].sort))
         return list(set(_subsorts))
 
+    def sort(self, kast: KInner) -> KSort | None:
+        if type(kast) is KToken:
+            return kast.sort
+        if type(kast) is KVariable:
+            return kast.sort
+        if type(kast) is KRewrite:
```
tothtamas28, line 1086: This is fine as-is. You can also implement such logic very nicely using [pattern matching](https://peps.python.org/pep-0636/):

```python3
match kast:
    case KToken(_, sort) | KVariable(_, sort):
        return sort
    case Rewrite(lhs, rhs):
         ...
```

(Not tested.)

tothtamas28 (approved):

src/pyk/kast/outer.py
ehildenb: Love it! Done!

PR #466, opened 2023-05-26
Move PrettyPrinter class into kast submodule

ehildenb:
This is factored out of: https://github.com/runtimeverification/pyk/pull/433

This PR moves the `PrettyPrinter` class into its own module `kast.pretty`, as well as addressing some minor printing issues. In particular:

- Class `PrettyPrinter` is moved from `ktool.kprint` to `kast.pretty`, and associated helpers are moved too.
- `PrettyPrinter` now takes more arguments associated with unparsing directly in Python (but not with unparsing using `kast`), including:
  - `definition: KDefinition`: used to build a `SymbolTable`.
  - `extra_unparsing_modules: Iterable[KFlatModule]`: used for defining new symbols on the fly and being able to unparse them.
  - `patch_symbol_table: Callable[[SymbolTable], None]`: instead of relying on overriding the classmethod, it's passed as explicit argument.
- All the appropriate arguments are threaded through `KPrint`, `KRun`, and `KProve` to be able to build a `PrettyPrinter`.
- `PrettyPrinter.print` is broken into two separate methods: `PrettyPrinter._print_kouter` and `PrettyPrint._print_kinner`.
  - First it checks which method it should call on the passed in term.
  - Teh sub-methods `PrettyPrinter._print_*` are calling the more speciic methods most of the  time, *except*:
  - When we switch from printing `KOuter` to printing `KInner` or vice-versa, we go back through `PrettyPrinter.print` top-level method. This allows us to do things like call `undo_aliases` _only once_ when we go to unparse a `KInner`, but also have it work inside a `KRule` when unparsing the rule body or requires clause, for example.
- The method `pretty_print_kast` is removed, and relevant tests of that deleted. We no longer expose the functionality of printing directly from a `SymbolTable`, and that becomes a more opaque internal data-structure.


src/pyk/kast/pretty.py
```diff
@@ -0,0 +1,404 @@
...
+            rule_str = rule_str + '[' + kterm.att['label'] + ']:'
+        rule_str = rule_str + ' ' + body
+        atts_str = self.print(kterm.att)
+        if kterm.requires != TRUE:
+            requires_str = 'requires ' + '\n  '.join(self._print_kast_bool(kterm.requires).split('\n'))
+            rule_str = rule_str + '\n  ' + requires_str
+        if kterm.ensures != TRUE:
+            ensures_str = 'ensures ' + '\n  '.join(self._print_kast_bool(kterm.ensures).split('\n'))
+            rule_str = rule_str + '\n   ' + ensures_str
+        return rule_str + '\n  ' + atts_str
```
ehildenb, line 263: Note here for example, we call back into `self.print(...)` instead of `self._print_kinner(...)` in order to make sure any extra processing (like undoing aliases) can be applied to the `KInner` before printing.

This allows us to define passes that happen on `KInner` that will apply both if `PrettyPrinter.print` is called on a `KInner` or a `KOuter`.

tothtamas28: You can achieve the same with a preprocessing step and the straightforward recursion scheme:

```python3
class PrettyPrinter:
    def print(self, kast: KAst) -> str:
        kast = self._preprocess(kast)
        return self._print(kast)

    def _preprocess(self, kast: KAst) -> KAst
        return kast  # or do some transformation

    def _print(self, kast: KAst) -> str:
        match kast:
            ...
```

This decouples transforming the object to print from actually printing it.

tothtamas28 (approved):

src/pyk/kast/pretty.py
ehildenb: I've made a separate function `on_kinner` in `kast.manip`, which maps a `KInner` transformation over any `KAst`.

src/pyk/kast/pretty.py
```diff
@@ -0,0 +1,404 @@
...
+_LOGGER: Final = logging.getLogger(__name__)
+
+SymbolTable = dict[str, Callable[..., str]]
+
+
+class PrettyPrinter:
+    definition: KDefinition
+    _extra_unparsing_modules: Iterable[KFlatModule]
+    _patch_symbol_table: Callable[[SymbolTable], None] | None
+    _unalias: bool
```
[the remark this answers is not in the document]
ehildenb: I'd rather wait for an actual need for that. Separating it is difficult because we don't want to compute the symbol table unless its needed (for performance). As it is currently done, we are computing it lazily, and memoizing it.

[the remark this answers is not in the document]
tothtamas28: > I'd rather wait for an actual need for that.

That's fine.

> Separating it is difficult because we don't want to compute the symbol table unless its needed

You can make `_symbol_table` a method.

ehildenb:
I'm reverting the changes made above to the state that was approved, and will open a follow-up PR, because separating the recursion scheme is proving more tricky because of type-system weaknesses.

src/pyk/kast/pretty.py
ehildenb: See: https://github.com/runtimeverification/pyk/issues/473

src/pyk/kast/pretty.py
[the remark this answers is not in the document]
ehildenb: See: https://github.com/runtimeverification/pyk/issues/473

PR #472, opened 2023-05-31
Sort parameter inference for nested parametric productions

ehildenb:
Related: https://github.com/runtimeverification/k/issues/3436

On KEVM, I noticed a case where our parametric production sort inference was failing and falling back to using `kast`. In this case, we had a nested `#ite ... #ite`, where the inner one was able to determine its sort completely, but the outer one was giving up because it would not consider arguments that were parametric.

In this PR we:

- Add such an example where inner sort parameters are able to be determined as a failing test.
- Strengthens `KDefinition.add_sort_params` to be able to check if it can determine the return sort of a parametric production by lining them up with the given terms klabel sort parameters.


ehildenb:
Before this change on KEVM, it was falling back to using `kast` nested `#ite` like this:

image

Which was taking about 2s at startup time:

image

Following this change, it no longer spends that time calling `kast`.

tothtamas28 (approved):

PR #478, opened 2023-06-03
Add basic parsers for CLI args

### LINKED ISSUE #260: Pull out argparse library from KEVM foundry integration
Currently, KEVM Foundry has the most advanced usage of pyk library for proving. It has a bunch of CLI options which are generic and re-usable among different projects.

For example, this PR factors out some common argument parsers: https://github.com/runtimeverification/evm-semantics/pull/1625

We should upstream the helper parsers here which make sense: https://github.com/runtimeverification/evm-semantics/blob/master/kevm-pyk/src/kevm_pyk/__main__.py#L457

Perhaps new modules: `kcfg/cli.py`, and `ktool/cli.py`?

Then, downstream users can import these generic options, and get similar (if not the same) user interfaces across different semantics.

ehildenb:
Fixes: https://github.com/runtimeverification/pyk/issues/260

This adds basic CLI parsers from KEVM to here.

- The helpers `cli_utils` is moved to `cli.utils` instead, to create the submodule.
- The helpers for parsing individual arguments from KEVM are upstreamed to `cli.utils`.
- Promote several things from `cli.utils` to `utils`, including: `run_process`, `BugReport`, and `check_*`.
- A new module `cli.args` is added with `KCLIArgs`, which contains the basic common K cli tool arguments from KEVM.
- The pyk cli is adapted to use the new upstreamed argument parsers. Unfortunately, this does not mean very much of the CLI is captured here, because the pyk cli has not been fleshed out to handle all the ktools yet.


ehildenb:
Related (it will make us have to implement more of the CLI): https://github.com/runtimeverification/pyk/issues/483

tothtamas28 (approved):

PR #480, opened 2023-06-05
Bump textual to latest

iFrostizz:
Anticipating for https://github.com/runtimeverification/pyk/pull/460
This will allow us to use `VerticalScroll` widgets to scroll programatically.


src/pyk/kcfg/tui.py
```diff
@@ -49,8 +48,8 @@ def on_enter(self) -> None:
     def on_leave(self) -> None:
         self.styles.border_left = None  # type: ignore
 
-    async def on_click(self, click: Click) -> None:
-        await self.emit(GraphChunk.Selected(self, self.id or ''))
+    def on_click(self, click: Click) -> None:
+        self.post_message(GraphChunk.Selected(self.id or ''))
         click.stop()
```
ehildenb, line 53: Is this making the messages synchronous? Does that affect performance on KEVM negatively?

iFrostizz: This is the new method, pyk was using a really outdated version of textual
I couldn't notice any significant slow down with or without this change

ehildenb (approved):

PR #484, opened 2023-06-07
Possible bug/fix with `KCFG` next node ID incrementing during `to_dict`

nwatson22:
It looks like whenever `to_dict` is called on a `KCFG`, the `"next"` field is incremented. I'm guessing this is not intended and rather the `dict` version of the `KCFG` should be able to be requested multiple times and get the same result.

I added a test to explicitly test the `to_dict` method and another that calls `to_dict` twice and compares the output. This latter test exercises the behavior I described.

The issue is "fixed" by using a plain `int` to track the next node ID instead of a `count()` iterator, although it would be preferable to continue using `count()`. I'm just not sure how to make that work.


tothtamas28 (approved):
Nice catch!

PR #485, opened 2023-06-07
Port proof methods and associated tests from #394

nwatson22:
This PR takes some methods for the `Proof` abstract class from #394 into a separate pull request.

Adds:
- `Proof.digest()`
- `Proof.up_to_date(check_method)`
- `Proof.read_proof(id, proof_dir)`, Generic way to read a proof from a file, regardless of type
- `hash_file(file)` to compute the hash of a file

A test for `read_proof` with test cases for each of the 3 existing proof types

This test should currently be failing due to #484.


ehildenb (approved):

src/tests/unit/test_proof.py
```diff
@@ -0,0 +1,87 @@
...
+        'proof_param': 1,
+    },
+]
+
+
+@pytest.fixture(scope='function', params=PROOF_TEST_DATA)
+def sample_proof(
+    request: pytest.FixtureRequest,
+    proof_dir: Path,
+) -> Proof:
```
ehildenb, line 47: These can maybe be parametrized in a bit nicer way, see for example: https://github.com/runtimeverification/pyk/blob/6191b958fef40e5db222cb5e825e51656ea0293c/src/tests/integration/kcfg/test_simple.py#L62

nwatson22: Done. I wish there was a way to just directly call the `proof_dir` fixture in the `PROOF_TEST_DATA` initialization.

Issue #471, opened 2023-05-31
Off-by-one error in BMC prover


ehildenb:
After a discussion with @PetarMax , it seems that we are doing one too few loop unrolls in the BMC prover.

Issue #413, opened 2023-05-05
Proof-specific viewers for Each proof type


ehildenb:
Currently, we have `kcfg/tui.py` which implements the "KCFG Viewer". This is really geared towards visualizing `APRProof` instead. Resources:

- `pretty_segments`: main workhorse for viewing a KCFG. https://github.com/runtimeverification/pyk/blob/cb1bb221abbd4586331f8899475aef1b049997ef/src/pyk/kcfg/show.py#L39. This stores a list of tuples of `str, Iterable[str]`, so that each chunk can be shoved into a "GraphChunk": https://github.com/runtimeverification/pyk/blob/cb1bb221abbd4586331f8899475aef1b049997ef/src/pyk/kcfg/tui.py#L32. A `GraphChunk` is most importantly clickable, and can have its styles modified.
- These `GraphChunk` are built up in the `BehaviorView`: https://github.com/runtimeverification/pyk/blob/cb1bb221abbd4586331f8899475aef1b049997ef/src/pyk/kcfg/tui.py#L57. This behavior view just renders the graph-chunks all in order, according to the "vertical" layout styling: https://github.com/runtimeverification/pyk/blob/cb1bb221abbd4586331f8899475aef1b049997ef/src/pyk/kcfg/style.css#L5
- The `NodeView` is on the rigth of the viewer, and contains: (i) info about how the display is, (ii) the "term" view, and (iii) the "constraint" view, and (iv) the "custom" view. https://github.com/runtimeverification/pyk/blob/cb1bb221abbd4586331f8899475aef1b049997ef/src/pyk/kcfg/tui.py#L88. The custom view is currently just used for displaying Solidity source mapping info.
- This is all wrapped up into the `KCFGViewer` class: https://github.com/runtimeverification/pyk/blob/cb1bb221abbd4586331f8899475aef1b049997ef/src/pyk/kcfg/tui.py#L236

What we want is to make these things more library/widget like, rather than a big monolithic thing. In particular:

- The `BehaviorView` is just the `KCFGViewer`, as a widget which can be embedded into other applications.
- The `NodeView` is a viewer for _any_ `GraphChunk`, but really it should be made more specialized for different types of graph chunks. A viewer for Node, a viewer for Split, for Edge, for NDBranch, and for Cover. Each one custom, with it's own properties. Likely all supporting the `minimize` flag.
- The `CustomView` is not a thing anymore built in to the app.
[...]

PR #486, opened 2023-06-08
Code re-organizations prepping for KCFG refactors

ehildenb:
Part of #471, #413

As I was working towards the goal of better BMC prover visualization, it's clear we need:

- Proof specific viewers,
- Fixing the BMC checker off-by-one-error, and
- Trimming of the structure stored in the KCFG (removing the notion of `target`, `expanded`).

This makes the first least controversial steps towards those changes, including:

- Method `KCFG.replace_node` now replaces the node in place, so that the node identifier doesn't change. This is possible now because the node identifier is not tied to the contents of the node, and makes tracking things like `init/target/terminal/bounded` sets externally in the `*Proof` classes more feasible even if nodes are being simplified for example.
- The notion of a unique `target` and `init` is added to the `APRProof` class, in preparation for removing the `target` notion from `KCFG`.
  - Add fields `target` and `init` to `APRProof` state, which explicitly call out the unique init/target states.
  - Move `KCFGExplore.target_subsume => APRProver._check_subsumed`, which makes use of the explicitly called out `init/target` states.
  - Moves `KCFG.path_constraints => APRProof.path_constraints`, because it relies on a unique target node.
- The notion of `stuck` nodes is moved into `APRProof` state as `terminal` nodes.
  - The field `_terminal_nodes` is added to `APRProof` state.
  - `APRProver._check_terminal` directly manipulates this set instead of `stuck/expanded` set.
  - The notion of `APRProof.terminal` is added, which returns the set of nodes in the `APRProof._terminal_nodes` set.
  - The notion of `APRProof.pending`  is added, which returns the set of `KCFG.frontier` nodes that are _not_ in the proofs `terminal_nodes` set.
  - The `APRProof.terminal` and `APRProof.pending` sets are used for determining `APRProof.status`.
  - The `APRBMCProof.bounded` set is added, which similarly tracks the bounded set of nodes directly. This is used in computing `APRBMCProof.pending` and `APRBMCProof.status`.
- Methods `APRProof.from_claim` and `APRBMCProof.from_claim_with_bmc_depth` are added for reading these proof types directly from claims (and tests are updated appropriately).
- Method `APRBMCProof.dict` is refactored to use `APRProof.dict` directly first, then add the needed information to it.
- Methods `KCFG.node_attrs => KCFGShow.node_attrs` and `KCFG.node_short_info => KCFGShow.node_short_info` are moved to localize them to use site.
- Adds `Proof.json` as shortcut for `json.dumps(Proof.dict)`.
- For each of the `integration/kcfg` tests that do `APRProof`s, adds logging at `info` level for the generated KCFG, to make debugging them easier by just throwing a flag.


tothtamas28:
> The notion of `stuck` nodes is moved into `APRProof` state as `terminal` nodes

Do you also plan to remove `stuck` from `KCFG`? It is reachability information that I think should be stored there. Proof classes can interpret that information accordingly.

> The notion of a unique `target` and `init` is added to the `APRProof` class, in preparation for removing the `target` notion from KCFG.

Subsumption into the target will still be retained in the KCFG though, right?

ehildenb:
After discussion, we'll be removing the notion of `expanded`, in favor of an explicit `stuck` set (to mark that the backend has attempted symbolic execution from that node and it will not be able to make progress). But `target` will be removed.

PR #489, opened 2023-06-12
APRProof specific proof state visualization

ehildenb:
Fixes: #413 

This PR introduces proof-specific viewers, and does a few refactors to make that possible:

- Helpers are added to `KCFG` for turning into modules, including:
  - `KCFG.Split.covers` which returns each individual split as a reverse cover.
  - `KCFG.NDBranch.edges`, which returns each non-deterministic step as an edge.
  - `KCFG.Edge.to_rule`, which converts an edge into a K rule (or claim).
- Small refactorings to `KCFGShow` structure:
  - `KCFGShow.dot` returns a `Digraph` directly (for further manipulation) instead of a list of `str`.
  - `KCFGShow.to_module` which returns a `KFlatModule` is factored out and used directly in `KCFGShow.show`.
  - Helper methods embedded in other methods in `KCFGShow` are factored out directly to staticmethods of `KCFGShow` so they can be reused easily.
  - We just treat `stuck` as a node attribute now, printed at the beginning of a node, instead of printing it a second time following a node too.
- The way that `KCFGShow` calculates node attributes is adjusted and factor out.
  - A new class `NodePrinter` is factored out, which by default just uses the attributes the `KCFG` knows about for printing out nodes.
  - That class is extended by `APRProofNodePrinter` and `APRBMCProofNodePrinter`, which add the additional information about the nodes `pending,terminal,bounded` state. If a node is `terminal` or `bounded`, then the `stuck` attribute is no longer printed for less confusion.
  - `KCFGShow` now takes an optional `*NodePrinter` as an argument (or builds on itself), so instantiating that class is how you can customize how nodes are displayed in text.
  - The tests are adapted to use the more specific `*NodePrinter` classes, and the output checked that the more specific node information is indeed being included.
- Adds helper predicates `APRProof.is_{pending,terminal}` and `APRProof.is_bounded`, which should behave better in case there are qualified node ids passed in.


ehildenb:
Running `bmc-loop-symbolic-1` test with log level set to `INFO` now prints out something looking like this for BMC depth-bounded nodes:

image

You can see that it includes the `bounded` label in that node ID, which makes it easier to spot that it's `stuck` because it's `depth-bounded`.

src/pyk/kcfg/kcfg.py
```diff
@@ -79,6 +80,22 @@ def to_dict(self) -> dict[str, Any]:
                 'depth': self.depth,
             }
 
+        def to_rule(self, defn: KDefinition, label: str, claim: bool = False, priority: int | None = None) -> KRuleLike:
```
nwatson22, line 83: Does this function need to take a KDefinition?
```suggestion
        def to_rule(self, label: str, claim: bool = False, priority: int | None = None) -> KRuleLike:
```

nwatson22 (approved):

src/tests/unit/test_kcfg.py
```diff
@@ -519,17 +516,16 @@ def _x_equals(i: int) -> KInner:
         '┃  ┃ (1 step)\n'
         '┃  ┣━━┓\n'
         '┃  ┃  │\n'
-        '┃  ┃  └─ \033[1m24 (frontier, leaf)\033[0m\n'
+        '┃  ┃  └─ 24 (frontier, leaf)\n'
```
tothtamas28, line 519: `id`-s are no longer **bold**, is this deliberate?

src/pyk/kcfg/kcfg.py
ehildenb: Ohhhhh, nice catch!

ehildenb: Fixed.

PR #498, opened 2023-06-15
Add ability to pass in custom function to KCFGExplore for modifying new nodes

nwatson22:
This PR adds the ability to pass in a function simplify_node: CTerm->CTerm to `KCFGExplore()` which modifies new nodes as they are added during exploration. This is intended to be used to accomplish https://github.com/runtimeverification/evm-semantics/issues/1909.


ehildenb:
Let's call it `abstract_node` instead of `simplify_node`, and make it an argument to `APRProver` instead of `KCFGExplore`.

We have have another block like `_check_subsume` or `_check_terminal`, which is `_check_abstract`. https://github.com/runtimeverification/pyk/blob/master/src/pyk/proof/reachability.py#L316. It should:

- See if auto_abstraction exists, if it does, try applying it.
- If applying it changes the cterm, then insert the new cterm into the KCFG, and _also_ insert a cover from the old node to the new one. Then skip the rest of `advance_proof` (return `true` from `_check_abstract` so that it can `continue` just like the other `_check*` methods).
- If `auto_bastraction` doesn't exist, or doesn't result in a change in the cterm, continue with checking `extract_branches`.
- Make sure to thread through the new argument to the `APRBMCProver` too.

src/pyk/proof/reachability.py
```diff
@@ -257,16 +257,19 @@ class APRProver:
...
         self,
         proof: APRProof,
         is_terminal: Callable[[CTerm], bool] | None = None,
         extract_branches: Callable[[CTerm], Iterable[KInner]] | None = None,
+        abstract_node: Callable[[CTerm], CTerm] | None = None,
     ) -> None:
         self.proof = proof
         self._is_terminal = is_terminal
         self._extract_branches = extract_branches
+        self._abstract_node = abstract_node
```
ehildenb, line 272: Make sure the extra argument is threaded through ot `APRBMCProver` as well.

ehildenb (approved):

src/pyk/proof/reachability.py
ehildenb: Ah wait, my mistake, it is!

ehildenb (approved):

PR #499, opened 2023-06-16
Fix sort vars's handling of quantifiers

virgil-serbanuta:
When adding sorts to variables, pyk takes all variables in an expression, regardless of whether they are free or not, groups them by name and merges their sorts.

This creates problems when using quantifiers, e.g. in this term:
```
exists x:A . term1(x)
#And
exists x:B . term2(x)
```
pyk will attempt to create a single variable x having A and B as sorts, which will fail.

This PR makes quantified variables "visible" only under their quantifier when assigning sorts.


src/pyk/kast/inner.py
```diff
@@ -696,3 +698,40 @@ def build_cons(unit: KInner, label: str | KLabel, terms: Iterable[KInner]) -> KI
         return KApply(label, (fst, build_cons(unit, label, it)))
     except StopIteration:
         return unit
+
+
+def fold(folder: Callable[[KInner, list[A]], A], to_fold: KInner) -> A:
```
tothtamas28, line 703: What does this function do?

src/pyk/kast/inner.py
```diff
@@ -696,3 +698,40 @@ def build_cons(unit: KInner, label: str | KLabel, terms: Iterable[KInner]) -> KI
...
+        children.append(fold(folder, child))
+        return child
+
+    children: list[A] = []
+    to_fold.map_inner(lambda c: child_collect(children, c))
+    return folder(to_fold, children)
+
+
+def unzip(l: list[tuple[A, B]]) -> tuple[list[A], list[B]]:
+    return ([a for a, _ in l], [b for _, b in l])
```
tothtamas28, line 714: `zip(*l)` should do the same I think.

src/pyk/kast/inner.py
```diff
@@ -696,3 +698,40 @@ def build_cons(unit: KInner, label: str | KLabel, terms: Iterable[KInner]) -> KI
...
+
+
+def fold_transform(
```
tothtamas28, line 717: What does this function do?

src/pyk/kast/outer.py
```diff
@@ -1125,29 +1125,42 @@ def greatest_common_subsort(self, sort1: KSort, sort2: KSort) -> KSort | None:
...
-            if type(_kast) is KSequence and _kast.arity > 0:
-                for a in _kast.items[0:-1]:
-                    if type(a) is KVariable:
-                        _var_sort_occurrences[a.name].append(a.let_sort(KSort('KItem')))
-                last_a = _kast.items[-1]
-                if type(last_a) is KVariable:
-                    _var_sort_occurrences[last_a.name].append(last_a.let_sort(KSort('K')))
-
-        collect(_sort_contexts, kast)
+    def sort_vars(self, kast: KInner, sort: KSort | None = None) -> KInner:
```
tothtamas28, line 1128: How does this new implementation work? in particular, how does it handle variable shadowing?

src/pyk/kast/inner.py
virgil-serbanuta: That's probably correct, but it messes up types:
```
src/pyk/kast/inner.py:714: error: Incompatible return value type (got "zip[Any]", expected "Tuple[List[A], List[B]]")
```

src/pyk/kast/outer.py
virgil-serbanuta: It does a bottom-up transformation of a KInner, computing the list of free variables in the transformed part at the same time.

To be specific, it does the following in a bottom-up traversal:
* Whenever it finds a variable, it returns it, both as a transformed term, and as a singleton list of free variables. E.g. it returns (Variable, {Variable}).
* Whenever it finds a quantifier for a variable X with a child C (e.g. exists X . C), it takes the list of free variables for its child C (i.e., freevars(C)), assigns a sort to X and replaces it in the entire `exists X . C` expression. Then it returns (replace(X := sorted-X, exists X . C), freevars(C) - {X}).
* For all other terms, it returns the term together with the union of its children's free vars.

virgil-serbanuta (commented):
FWIW, I'm not sure why

src/pyk/kast/outer.py
tothtamas28: Makes sense, thank you.

src/pyk/kast/inner.py
tothtamas28: Yes, probably a less abstract implementation would be more concise and readable in this case.

src/pyk/kast/outer.py
```diff
@@ -1125,29 +1125,42 @@ def greatest_common_subsort(self, sort1: KSort, sort2: KSort) -> KSort | None:
...
+                    if key not in result:
+                        result[key] = []
+                    result[key] += value
+            if isinstance(term, KVariable):
+                add_var_to_dict(term, result)
+            elif isinstance(term, KApply):
+                if term.label.name in ML_QUANTIFIERS:
+                    var = get_quantifier_variable(term)
+                    add_var_to_dict(var, result)
+            return result
```
tothtamas28, line 1160: Consider using `collections.deafultdict`:

```python3
result: dict[str, list[KVariable]] = defaultdict(list)
...
result[var.name].append(var)
```

Then `add_var_to_dict` can be removed.

src/pyk/kast/inner.py
tothtamas28: Please move it into `fold_transform` then.

src/pyk/kast/inner.py
tothtamas28: Please try to eliminate these functions (`fold`, `fold_transform`) by a direct implementation of the recursion. If you think that would lead to less readable code, then move them to `outer.py` with a `_` prefix to mark them private.

virgil-serbanuta: I'm sorry, somehow I didn't notice this comment. I replaced this with a bottom_up_with_summary function. Is it reasonable, or should I try to make it more clear/less abstract?

tothtamas28 (approved):

src/pyk/kast/inner.py
tothtamas28: Looks good, thanks.

PR #502, opened 2023-06-16
Fix LHS and RHS extraction from rules.

virgil-serbanuta:
When normalizing a rewrite, pyk represents something like "(a => .K) ~> stuff" as "(a => #EmptyK) ~> stuff". When extracting the RHS of such a rewrite, pyk produces "#EmptyK ~> stuff" which breaks various tools.

This PR replaces "#EmptyK" with an actual empty KSequence (i.e., KSequence([])), which, for the example above, will produce just "stuff" as the RHS.


ehildenb:
Notice that here: https://github.com/runtimeverification/pyk/blob/b1f057583b4fb335b68258a5b6278ebb8f12b1f7/src/pyk/kast/inner.py#L586

We flatten nested `KSequence` at construction time. Maybe immediately after doing that, you can run this filter, so that we just never build a `KSequence` with an empty item it in.

src/pyk/kast/inner.py
```diff
@@ -696,3 +701,9 @@ def build_cons(unit: KInner, label: str | KLabel, terms: Iterable[KInner]) -> KI
         return KApply(label, (fst, build_cons(unit, label, it)))
     except StopIteration:
         return unit
+
+
+def is_empty_k(item: KInner) -> bool:
+    if not isinstance(item, KApply):
+        return False
+    return item.label.name == EMPTY_K_LABEL
```
tothtamas28, line 709: I'd just move `EMPTY_K` here and compare for equality, then this and `EMPTY_K_LABEL` become unnecessary.

src/pyk/kast/manip.py
```diff
@@ -148,12 +159,21 @@ def simplify_bool(k: KInner) -> KInner:
     return new_k
 
 
+def remove_empty_k(term: KInner) -> KInner:
```
tothtamas28, line 162: Ideally calling this function should never be necessary (except perhaps for terms coming from the outside). Can you root cause where the malformed terms are coming from in your use case?

src/pyk/kast/inner.py
virgil-serbanuta: Actually, I removed it EMPTY_K.

src/pyk/kast/manip.py
virgil-serbanuta: pyk was adding this in `push_down_rewrites`. I rewrote this PR.

ehildenb (approved):

ehildenb:
@virgil-serbanuta I've approved, but please do two things:

- Add a description to the main PR comment above.
- Add a test of `push_down_rewrites`: https://github.com/runtimeverification/pyk/blob/94c3f6928d5dbdafd064de4ad63fbd940c47bb2a/src/tests/unit/kast/test_manip.py#L44

PR #505, opened 2023-06-19
Deduplicate productions

virgil-serbanuta:


src/pyk/kast/outer.py
```diff
@@ -1033,6 +1033,23 @@ def production_for_klabel(self, klabel: KLabel) -> KProduction:
...
+                equivalent = False
+                trimmed_prod = prod.let_att(KAtt({}))
+                for _, t in unique_no_att:
+                    if t == trimmed_prod:
+                        equivalent = True
+                        break
+                if equivalent:
+                    continue
+                unique_no_att.append((prod, trimmed_prod))
+            _prods = [p for p, _ in unique_no_att]
```
tothtamas28, line 1049: If the goal is just to bin productions w. r. t. equivalence (not taking `att` in account), you can do that in linear time using a `dict`. Or if the attributes are not interesting for the result then even `Counter`:

```python3
from collections import Counter

from .kast import EMPTY_ATT

cnt = Counter(prod.let_att(EMPTY_ATT) for prod in prods)
```

Then `list(cnt)` is the list of unique productions, `len(cnt)` is the number of bins, `len(cnt[prod])` is the caridnality of `prod`-s bin, etc.

src/pyk/kast/outer.py
```diff
@@ -1033,6 +1033,23 @@ def production_for_klabel(self, klabel: KLabel) -> KProduction:
...
+            if len(_prods) < len(prods):
+                prods = _prods
+                _LOGGER.warning(f'Discarding {len(prods) - len(_prods)} equivalent productions')
```
tothtamas28, line 1052: Wouldn't this always log
```
Discarding 0 equivalent productions
```

ehildenb:
@virgil-serbanuta do you have a minimal example where multiple productions are produced? I guess that causes a crash, because in several places we assert that a single production is produced, so w eshould add a test of that to our integration tests, to make sure the filtering is working correctly.

I haven't seen a case where this happens yet, but maybe I haven't used fancy enough import structure yet. Maybe we also just need to change the way we recursively collect sentences from imported modules instead of deduplicating the productions after the fact. Right now `KDefinition.sentences`, for example, just grabs all the sentences in all the modules. It could instead, for example, walk the import DAG, build a traversal that makes sense given the import structure that only visits each module once, then collect the sentences from the modules in that order.

virgil-serbanuta:
> @virgil-serbanuta do you have a minimal example where multiple productions are produced? 
a.k:
```
module A
  imports B
  imports C
endmodule

module B
  syntax Int
endmodule

module C
  syntax Int
endmodule
```

```
kompile a.k --backend haskell --emit-json
```

Search for
```
{"node":"KProduction","klabel":{"node":"KLabel","name":"isInt","params":[]}
```
in a-kompiled/compiled.json

src/pyk/kast/outer.py
virgil-serbanuta: Fixed. Initially (before removing attributes) I tried doing something smarter when filtering duplicates and got an error about productions not being hashable. I assumed that dictionaries would also fail.

However, it seems to work right now, so that either dictionaries do not use hashes, or removing attributes made them hashable.

virgil-serbanuta:
I also added a test which fails on master.

ehildenb:
I still don't understand how we are getting duplicate productions in the first place, which is what I think needs to be addressed rather than just filtering them out.

Here is where we gather the productions: https://github.com/runtimeverification/pyk/blob/8b8f791d136b5307ed093850c3318401186b4a91/src/pyk/kast/outer.py#L985.

Which uses the gathering of the modules: https://github.com/runtimeverification/pyk/blob/8b8f791d136b5307ed093850c3318401186b4a91/src/pyk/kast/outer.py#L981.

Which uses the gathering of the module names: https://github.com/runtimeverification/pyk/blob/8b8f791d136b5307ed093850c3318401186b4a91/src/pyk/kast/outer.py#L966

Which as far as I can tell, should only gather each module a single time (as long as it gets the same name). So in the example of your test,  the module `INT` should only be collected a single time, not twice as it traverses both sides of the import path.

src/pyk/kast/outer.py
```diff
@@ -1029,10 +1029,20 @@ def production_for_klabel(self, klabel: KLabel) -> KProduction:
...
+            # Automatically defined symbols like isInt may get multiple
+            # definitions in different modules.
+            unique_no_att: dict[KProduction, KProduction] = {}
+            for prod in prods:
+                trimmed_prod = prod.let_att(EMPTY_ATT)
+                unique_no_att[trimmed_prod] = prod
+            _prods = list(unique_no_att.values())
+            if len(_prods) < len(prods):
+                _LOGGER.warning(f'Discarding {len(prods) - len(_prods)} equivalent productions')
+                prods = _prods
```
ehildenb, line 1045: Which attributes are ending up being different between different productions? Is it just the source/locaiton attributes? If so, then we should remove _just_ those attributes and do the comparison I think, rather than removing _all_ attributes. Some other attributes may be needed semantically to distinguish productions. Here is an example of removing the source maps: https://github.com/runtimeverification/pyk/blob/master/src/pyk/kast/manip.py#L495

virgil-serbanuta: Sure, although the existing deduplication code did not work anymore (`TypeError: unhashable type: 'dict'`), so I had to change it.

tothtamas28: That's a bug. this patch should fix it:

```diff
--- a/src/pyk/kast/kast.py
+++ b/src/pyk/kast/kast.py
@@ -88,7 +88,7 @@ class KAtt(KAst, Mapping[str, Any]):
             if isinstance(m, (int, str, tuple, FrozenDict, frozenset)):
                 return m
             elif isinstance(m, list):
-                return tuple(v for v in m)
+                return tuple(_freeze(v) for v in m)
             elif isinstance(m, dict):
                 return FrozenDict((k, _freeze(v)) for (k, v) in m.items())
             raise ValueError(f"Don't know how to freeze attribute value {m} of type {type(m)}.")
```

src/pyk/kast/outer.py
```diff
@@ -1323,3 +1340,12 @@ def _remove_config_var_lookups(_kast: KInner) -> KInner:
 def read_kast_definition(path: str | PathLike) -> KDefinition:
     with open(path) as f:
         return kast_term(json.loads(f.read()), KDefinition)
+
+
+def remove_source_map_from_katt(att: KAtt) -> KAtt:
```
tothtamas28, line 1345: This can be made a method on `KAtt` (`drop_source`, `remove_source`, ...)

src/pyk/kast/outer.py
```diff
@@ -1029,10 +1029,27 @@ def production_for_klabel(self, klabel: KLabel) -> KProduction:
...
             if len(_prods) < len(prods):
-                prods = _prods
                 _LOGGER.warning(
                     f'Discarding {len(prods) - len(_prods)} productions with `unparseAvoid` attribute for label: {klabel}'
                 )
+                prods = _prods
+            # Automatically defined symbols like isInt may get multiple
+            # definitions in different modules.
+            unique_no_att: list[tuple[KProduction, KProduction]] = []
+            for prod in prods:
```
tothtamas28, line 1039: After fixing `_freeze`, a `Counter`-base solution should work (not thoroughly tested):

```python
prod_cnt = Counter(prod.let_att(remove_source_map_from_katt(prod.att)) for prod in prods)
try:
    prod = single(prod_cnt)
except ValueError as err:
    raise ValueError(f'Expected a single production for label {klabel}, not: {list[prod_cnt]}') from err

num_eq_prods = prod_cnt[prod] - 1
if num_eq_prods:
    _LOGGER.warning(f'Discarding {num_eq_prods} equivalent productions')

self._production_for_klabel[klabel] = prod
```

This version always returns a production without source map. If retaining it is important (in the case whe there is a single production), a `dict`-based variant can be implemented with.

src/pyk/kast/outer.py
virgil-serbanuta: Thanks, fixed.

src/pyk/kast/kast.py
```diff
@@ -133,6 +133,14 @@ def update(self, atts: Mapping[str, Any]) -> KAtt:
     def remove(self, atts: Iterable[str]) -> KAtt:
         return KAtt({k: v for k, v in self.atts.items() if k not in atts})
 
+    def drop_source(self) -> KAtt:
+        new_atts = {
+            key: value
+            for key, value in self.atts.items()
+            if key != 'org.kframework.attributes.Source' and key != 'org.kframework.attributes.Location'
```
tothtamas28, line 140:
```suggestion
            if key != self.SOURCE and key != self.LOCATION
```

src/pyk/kast/outer.py
```diff
@@ -1029,10 +1030,17 @@ def production_for_klabel(self, klabel: KLabel) -> KProduction:
...
                 _LOGGER.warning(
                     f'Discarding {len(prods) - len(_prods)} productions with `unparseAvoid` attribute for label: {klabel}'
                 )
+                prods = _prods
+            # Automatically defined symbols like isInt may get multiple
+            # definitions in different modules.
+            counter = Counter(prod.let_att(prod.att.drop_source()) for prod in prods)
+            _prods = list(counter)
+            if len(_prods) < len(prods):
+                _LOGGER.warning(f'Discarding {len(prods) - len(_prods)} equivalent productions')
```
tothtamas28, line 1042: `len(_prods)` is the number of keys in `counter`. `counter` is a `Mapping[KProduction, int]` where the key is the representative of the equivalence class (i.e. the production without the source attribute), and the value is the number of productions that fall into the equivalence class (i.e. those that equal the representative when dropping the source attribute).

virgil-serbanuta: Yes, I think that's correct. I'm not sure why you explained this, you probably wanted to point out something that I'm still missing.

All I'm saying here is that if I have 'n' production, which I group in 'm' equivalence classes, and I keep one production form each class, this means that I dropped all the other productions, i.e. I dropped n-m productions.

tothtamas28: > dropped n-m productions

I see, thanks for the clarification.

For me, the following order seems more logical though:

* If there's more than one equivalence class, fail.
* Otherwise, emit the log message about the number of discarded equivalent productions from the single class.

By the way, using this `n-m` trick, the `Counter` is unnecessary for both orderings: you can just use a `set` instead (converting to list is not necessary).

tothtamas28 (approved):

src/pyk/kast/outer.py
virgil-serbanuta: I changed it to a set.

To me, this code looks like a series of filters (remove 'unparseAvoid', remove equivalent productions), and other filters may be added in the future. To me, it makes sense to log each filter's activity, then move to the next step (and, for the last filter, the next step is to check how many productions are left and return)

ehildenb:
@virgil-serbanuta please get in the habit of adding a description to the main PR comment body (which gets included in the git changelog). It helps the reviewer, and it helps people who need to look through the history to find out why a change was introduced.

PR #507, opened 2023-06-20
Implement APRProofViewer as an APRProof specific TUI

ehildenb:
Following https://github.com/runtimeverification/pyk/pull/489, it turns out that KEVM's TUI is broken because it relies on the old way of doing Node printing.

This makes a new `APRProofViewer` class, which largely inherits from `KCFGViewer`, but adjusts it to handle the more spceific node viewers.

I have tested this with both KEVM+Foundry and raw KEVM specs, and it seems to work.


src/pyk/proof/style.css
```diff
@@ -0,0 +1,61 @@
...
+    height: auto;
+}
+#constraint {
+    width: auto;
+    height: auto;
+}
+#custom {
+    width: auto;
+    height: auto;
+}
```
nwatson22, line 61: Is this just a clone of `src/pyk/kcfg/style.css`? I wonder if it would be better to have `APRProofViewer` inherit the styling from its superclass.

nwatson22 (approved):

src/pyk/proof/style.css
ehildenb: Hmmmmm, that would be a good idea, sorry I didn't see this before it was merged.

PR #512, opened 2023-06-26
Adjust BMC loop detection for languages with coinciding loop heads and branches

ehildenb:
In GOTO style languages (which use `JUMP` instructions for control flow), the loop head and the branch points coincide. One example of such a language is EVM. In these languages, when detecting prior loop heads for BMC loop detection, we were counting each loop-head twice, because we were simply walking backwards in the KCFG and looking for nodes that were in the `_same_loop_head` heuristic relation. But because these loop heads were also branches, we were counting both the pre- and post-branch nodes as loop heads, even though it should only count a single time.

To fix this, we basically don't count nodes which are "zero-distance" from each other as distinct loop heads, which results in us only keeping the pre-branch node as a loop head.

In particular, we:

- Add some helpers to KCFG/APRProof path detection/manipulation:
  - `APRProof.shortest_path_to` is a shortcut for `KCFG.shortest_path_between(APRProof.init, SOME_NODE)`, utilizing the fact that every node in a `APRProof` should be guaranteed to be reachable from the `init` node.
  - KCFG.shortest_distance_between` is added as a helper which returns either `None` (no path between), or the shortest distance between two nodes (in either diretion).
  - `KCFG.zero_distance_between` is a function that returns `True` if there is a zero-length path between two nodes.
  - `APRProof.nonzero_depth` is refactored to use `KCFG.zero_distance_between`.
- The prior-loop detection for `APRBMCProver` is adjusted so that it:
  - Calculates an actual path in order from the root to the current node, instead of just looking at prior reachable nodes.
  - Only includes nodes in that path that are (i) not zero-depth from any of the other nodes in the current path, and (ii) not zero-depth from the current node. This prevents double-counting nodes.
- Adds a test of the correct loop-head detection:
  - The language `GOTO` is added, inspired by `EVM`, with trimmed down functionality.
  - A proof of the program `sum(N)` is added in language `GOTO`, and a BMC test harness is added for this proof, showing that it achieves the correct final branching factor of 2 instead of ending early because it over-detects loops.


src/tests/integration/kcfg/test_goto.py
```diff
@@ -0,0 +1,134 @@
...
+        return []
+
+    @staticmethod
+    def _same_loop(cterm1: CTerm, cterm2: CTerm) -> bool:
+        k_cell = cterm1.cell('K_CELL')
+        pc_cell_1 = cterm1.cell('PC_CELL')
+        pc_cell_2 = cterm2.cell('PC_CELL')
+        if pc_cell_1 == pc_cell_2 and type(k_cell) is KSequence and len(k_cell) > 0 and type(k_cell[0]) is KApply:
+            return k_cell[0].label.name == 'jumpi'
+        return False
```
nwatson22, line 84: Am I correct in understanding that for a more complex language that takes multiple K steps to jump, you would  also need to examine the cterm2 K cell when providing this heuristic? I just noticed here that this example is based entirely on the two terms having the same PC and only checks if one term is on a jump instruction and wondered if there was some kind of implicit meaning to the argument order in the `_same_loop` idea.

src/tests/unit/test_kcfg.py
```diff
@@ -257,21 +258,6 @@ def test_cover_then_remove() -> None:
...
-    # When
-    new_edge = cfg.create_edge(1, 1, 1)
-
-    # Then
-    assert new_edge == edge(1, 1)
-    assert set(cfg.nodes) == {node(1)}
-    assert set(cfg.edges()) == {edge(1, 1)}
-    assert cfg.edge(1, 1) == edge(1, 1)
-
-
```
nwatson22, line 274: Why are we removing this test?

nwatson22 (approved):

src/tests/unit/test_kcfg.py
ehildenb: It has a 0-length branch in it, so doesn't make sense with the new restrictions on KCFGs that 0-length loops aren't allowed.

src/tests/integration/kcfg/test_goto.py
ehildenb: In the case of this GOTO language, you only need to inspect the program counter to determine if it's teh same loop head, because there are no other control-flow structures (like in EVM, we have to also inspect things like call depth, see here: https://github.com/runtimeverification/evm-semantics/blob/1762c255c9650cdddf878114deb432dcd20fbf4d/kevm-pyk/src/kevm_pyk/kevm.py#L183).

So we are just checking: (i) is it the same program counter in both loop heads, and (ii) is it a JUMPI instruction (which loop heads should be jumpi instrucitons).

PR #514, opened 2023-06-27
Add BoosterServer to pyk and re-organise bug_report.tar

goodlyrottenapple:
In order to collect all the necessary files needed for running the new `kore-rpc-booster` when using the bug report flag, pyk needs to be aware of the llvm_definition folder when launching the booser, in order to include it in `bug_report.tar`.

This PR also re-organises the bug report tar in the following way:
image

server_instance.json:
```json
{"exe": "kore-rpc-booster", "module": "FOUNDRY-MAIN", "extra_args": ["--llvm-backend-library", "out/kompiled-llvm/interpreter.dylib", "--smt-timeout", "125", "--smt-retry-limit", "4"]}
```

server_version.txt (produced by `kore-rpc{-booster} --version`):
```
hs-backend-booster version:
  revision:	UNKNOWN
  branch:	UNKNOWN
  last commit:	UNKNOWN
```

when the llvm_definiton folder is supplied, we copy `definition.kore` and the `dt/` folder into the tar file, as we need those to build the shared lib when processing the bug report

we also only copy the definition.kore from the haskell kompiled folder, as that's the only other file we need...


src/pyk/ktool/kprint.py
```diff
@@ -176,6 +176,7 @@ def _build_arg_list(
 
 class KPrint:
     definition_dir: Path
+    llvm_definition_dir: Path | None
```
tothtamas28, line 179: `llvm_definition_dir` seems unrelated to the responsibility of the `KPrint` class. Why is this addition needed?

src/pyk/utils.py
```diff
@@ -511,3 +512,13 @@ def add_definition(self, defn_path: Path) -> None:
...
+
+
+class Kernel(Enum):
+    LINUX = 'Linux'
+    DARWIN = 'Darwin'
+
+    @staticmethod
+    def get() -> Kernel:
+        uname = run_process(('uname', '-s'), pipe_stderr=True, logger=_LOGGER).stdout.strip()
+        return Kernel(uname)
```
tothtamas28, line 524: It would be nice to cache the result.

src/pyk/ktool/kprint.py
goodlyrottenapple: Hmm we could pass it directly into KCFG explore I guess? but I included it in KPrint because that's where the code that emits the definition into the bug report for the `definition_dir` lives:

https://github.com/runtimeverification/pyk/blob/6cbf94f969dd2b7044d188e3abec3a1678abd953/src/pyk/ktool/kprint.py#L216-L217

goodlyrottenapple: based on the data in that class, would renaming to something like `KDefinitionSources` not be appropriate?

src/pyk/utils.py
goodlyrottenapple: I'm afraid I don't know enough about python to do that... can you combine `@cached_property` with `@staticmethod`??

tothtamas28: > can you combine `@cached_property` with `@staticmethod`

Probably not, but you can do a custom implementation:

```python3
from typing import ClassVar

class C:
    c: ClassVar[int | None] = None

    @classmethod
    def get_c(cls) -> int:
        if cls.c is None:
            print('Setting c')
            cls.c = 1
        return cls.c
```

src/pyk/ktool/kprint.py
tothtamas28: > Hmm we could pass it directly into KCFG explore I guess

Yes, that's the ideal solution I think.

> based on the data in that class, would renaming to something like `KDefinitionSources`/`KDefinitionData` not be appropriate?

This class probably has grown too big with time, but it's original responsibility is to do format conversions over a single definition.

src/pyk/utils.py
```diff
@@ -506,8 +507,16 @@ def _remap_arg(_a: str) -> str:
...
+
+class Kernel(Enum):
+    LINUX = 'Linux'
+    DARWIN = 'Darwin'
+
+    cached: ClassVar[Kernel | None] = None
+
+    @classmethod
+    def get(cls) -> Kernel:
+        if cls.cached is None:
```
tothtamas28, line 519:
```suggestion
        if not cls.cached:
```

src/pyk/kore/rpc.py
```diff
@@ -687,11 +689,25 @@ def __init__(
...
+
+            version_info = run_process((command[0], '--version'), pipe_stderr=True, logger=_LOGGER).stdout.strip()
+            bug_report.add_file_contents(version_info, Path('server_version.txt'))
+            server_instance = {
+                'exe': command[0],
+                'module': module_name,
+                'extra_args': command[1:] + smt_server_args + haskell_log_args,
+            }
+            bug_report.add_file_contents(json.dumps(server_instance), Path('server_instance.json'))
+
```
tothtamas28, line 708: Consider extracting a private static method.

src/pyk/kore/rpc.py
```diff
@@ -736,3 +752,63 @@ def close(self) -> None:
         self._proc.send_signal(SIGINT)
         self._proc.wait()
         _LOGGER.info(f'KoreServer stopped: {self.host}:{self.port}, pid={self.pid}')
+
+
+class BoosterServer(KoreServer):
```
tothtamas28, line 757: Please add a few tests that exercise this class.

src/pyk/utils.py
jberthold: Tried this, but ... 
```
poetry run mypy src
src/pyk/utils.py:521: error: Cannot assign to final attribute "cached"  [misc]
Found 1 error in 1 file (checked 152 source files)
make: *** [Makefile:74: check-mypy] Error 1
```

src/pyk/kore/rpc.py
jberthold: Sorry... I am a bit lost on this one - how is the server set up for the tests in `src/test/integration`? I could not find any use of `KoreServer` in there except in two unit test modules.
Could we postpone this or you give me some pointers? I would just want to run some/all integration tests against booster.

src/pyk/utils.py
jberthold: Well... that error was actually a good hint. The assignment to class variable `cached` did not work at all (seems it is implicitly `final` as a `ClassVar`). I am now running `uname -s` and assigning the final value when initialising.

tothtamas28: I think the problem is not with `ClassVar`, but that

```python3
cached = ...
```

is interpreted as an enum item definition.

> I am now running `uname -s` and assigning the final value when initialising.

I'd like to avoid that, because this code will run when `pyk.utils` is imported for the first time.

This works:

```python3
class Kernel(Enum):
    LINUX = 'Linux'
    DARWIN = 'Darwin'

    _cached: ClassVar[Kernel]

    @classmethod
    def get(cls) -> Kernel:
        if not hasattr(cls, '_cached'):
            uname = run_process(('uname', '-s'), pipe_stderr=True, logger=_LOGGER).stdout.strip()
            cls._cached = Kernel(uname)
        return cls._cached
```

src/pyk/kore/rpc.py
tothtamas28: There is this class:
https://github.com/runtimeverification/pyk/blob/d1492aecd4f3dc805e1582739592f55eac95ca8f/src/pyk/testing/_kompiler.py#L138

It defines a fixture that starts a `KoreServer` and provides a client for it.

src/pyk/utils.py
```diff
@@ -506,8 +507,13 @@ def _remap_arg(_a: str) -> str:
...
+
+class Kernel(Enum):
+    LINUX = 'Linux'
+    DARWIN = 'Darwin'
+
+    cached: ClassVar[Kernel] = run_process(('uname', '-s'), pipe_stderr=True, logger=_LOGGER).stdout.strip()
+
+    @classmethod
+    def get(cls) -> Kernel:
+        return cls.cached
```
jberthold, line 519:
```suggestion
    _cached: ClassVar[Kernel]

    @classmethod
    def get(cls) -> Kernel:
        if not hasattr(cls, '_cached'):
            uname = run_process(('uname', '-s'), pipe_stderr=True, logger=_LOGGER).stdout.strip()
            cls._cached = Kernel(uname)
        return cls._cached
```

src/pyk/kore/rpc.py
jberthold: I added some tests (running the test cases for `execute`, minus the `branching` case where the `next-states` in the response have a different order).

src/pyk/kore/rpc.py
```diff
@@ -687,18 +689,43 @@ def __init__(
...
         self._host, self._port = self._get_host_and_port(self._pid)
         if port:
             assert self.port == port
         _LOGGER.info(f'KoreServer started: {self.host}:{self.port}, pid={self.pid}')
 
+    @staticmethod
+    def gather_server_report(
+        module_name: str,
+        command: list[str],
+        bug_report: BugReport,
```
tothtamas28, line 714:
```suggestion
    def _gather_server_report(
        bug_report: BugReport,
        module_name: str,
        command: list[str],
```

tothtamas28: Consider passing `command: str` and `extra_args: list[str]` directly.

src/tests/integration/kore/test_kore_client.py
```diff
@@ -211,6 +211,35 @@ def state(n: int) -> State:
...
+        self,
+        booster_client: KoreClient,
+        test_id: str,
+        n: int,
+        params: Mapping[str, Any],
+        expected: ExecuteResult,
+    ) -> None:
+        if test_id == 'branching':
+            # order of next-states array differs in booster
+            pytest.skip()
```
tothtamas28, line 234: Would comparing as a set work?

src/pyk/testing/_kompiler.py
```diff
@@ -150,6 +163,21 @@ def kore_client(self, definition_dir: Path, bug_report: BugReport) -> Iterator[K
         server.close()
 
 
+class BoosterClientTest(KompiledTest, LlvmLibKompiledTest):
```
tothtamas28, line 166: Consider defining this class without subclassing. It's more verbose, but also more  symmetric in naming:

```python3
class BoosterTest:
    MAIN_FILE: ClassVar[str | Path]
    MODULE_NAME: ClassVar[str]
    HASKELL_ARGS: ClassVar[Mapping[str, Any]] = {}
    LLVM_ARGS: ClassVar[Mapping[str, Any]] = {}

    @pytest.fixture(scope='class')
    def haskell_dir(self, kompile: Kompiler) -> Path:
        ...

    @pytest.fixture(scope='class')
    def llvm_dir(self, kompile: Kompiler) -> Path:
        ...

    @pytest.fixture
    def booster_client(self, haskell_dir: Path, llvm_dir: Path, bug_report: BugReport) -> Iterator[KoreClient]:
        ...
```

tothtamas28: You can also probably extract `BugReport` from `KompiledTest`.

jberthold: I thought the double inheritance reflects very nicely the fact that we need two compilations... if I inherit only from `KompiledTest` (for `bug_report`) we lose this aspect, and if I don't inherit I need to duplicate the `bug_report` code. :shrug:

tothtamas28: > if I don't inherit I need to duplicate the `bug_report` code

You don't if you do this:

> You can also probably extract `BugReport` from `KompiledTest`.

> the double inheritance reflects very nicely the fact that we need two compilations

That's true, but when implementing a test, which set of attributes would you rather fill out? This:

```
MAIN_FILE
MODULE_NAME
```

or this:

```
KOMPILE_MAIN_FILE
KORE_MODULE_NAME
LLVM_LIB_MAIN_FILE  # same as KOMPILE_MAIN_FILE
```

src/tests/integration/kore/test_kore_client.py
jberthold: Of course... just need to find out how to convert the `tuple` to a `set` and whether that exists in python base... :book:

tothtamas28: `set` takes an `Iterable`:

```python3
>>> set((1, 2, 3, 1, 2))
{1, 2, 3}
```

jberthold: yeah, just that we have an _optional_ set. A bit ugly but I just pushed the change

src/pyk/testing/_kompiler.py
jberthold: IDK what you mean by 
> you can also probably extract `BugReport` from `KompiledTest`

tothtamas28: Right, sorry for the confusion. I mean lift it from the class scope, and make in a global fixture (in `plugin.py`).

At some point we'll probably introduce a command line argument to control whether it returns `None` or a `BugReport`.

jberthold: If you insist on the non-inheritance I will just make it `None` for now to get this in.

tothtamas28: Works for me.

PR #515, opened 2023-06-27
Refactor Proof/Prover class hierarchy

nwatson22:
Changes:
- `Prover`, a superclass of `EqualityProver`, `RefutationProver`, `APRProver`, and `APRBMCProver`
- `ImpliesProver`, a superclass of `EqualityProver` and `RefutationProver`
- `ImpliesProof`, a superclass of `EqualityProof` and `RefutationProof`
- Moves shared `advance_proof()` logic of `EqualityProver` and `RefutationProver` into `ImpliesProver`
- Removes unused `kcfg_explore` argument from `refute_node()`
- `kcfg_explore` argument is now consistently passed in on `Prover` construction


src/pyk/proof/equality.py
```diff
@@ -263,30 +271,37 @@ def pretty(self, kprint: KPrint) -> Iterable[str]:
...
 
 
-class EqualityProver:
-    proof: EqualityProof
+class ImpliesProver(Prover):
+    proof: ImpliesProof
+    _antecedent_kast: KInner
+    _consequent_kast: KInner
+    simplified_antecedent: KInner
+    simplified_consequent: KInner
```
ehildenb, line 279: I think we should store these simplified fields as part of the `ImpliesProof` state instead, because it's part of the proof state.

ehildenb (approved):

src/pyk/proof/equality.py
nwatson22: Would the same not also apply to the non-simplified `_antecedent_kast` and `_consequent_kast` fields as well?

ehildenb: Yeah, I guess it would.

PR #517, opened 2023-06-28
Add `failure_info` method for showing why `APRProof` failed

### LINKED ISSUE #463: Add `Proof.failure_info` for printing out more detailed information about proof failures
From: https://github.com/runtimeverification/evm-semantics/pull/1716/files#r1206006385

We should have a new method `APRProof.failure_info: KCFGExplore -> list[str]`, which prints out more detailed information about why a reachability proof failed. Just like the linked funciton about, it should print out more detailed information about the current frontier/stuck nodes in teh KCFG.

nwatson22:
Closes #463

This method was previously in `evm-semantics` and this upstreams it into `pyk`.
Adds a class `APRFailureInfo` which stores the IDs of the pending nodes, and the IDs of the failing nodes, along with the failure reason (`implication_failure_reason`) and path condition (`path_constraints`).
Also adds a test which checks the output of this function for `IMP-SIMPLE-SPEC.failing-if`.


src/tests/conftest.py
```diff
@@ -0,0 +1,22 @@
...
+        '--update-expected-output',
+        action='store_true',
+        default=False,
+        help='Write expected output files for proof tests',
+    )
+
+
+@pytest.fixture
+def update_expected_output(request: FixtureRequest) -> bool:
+    return request.config.getoption('--update-expected-output')
```
tothtamas28, line 22: To be honest, I'm not a big fan of this pattern, as it encourages writing tests that compare string output, which leads to a brittle test suite. This is fine for a few test cases where the string output actually matters (like this current one, though this feature could also be made more testable by introducing an intermediate data structure), but that does not require first-class support from the test harness. Please remove this.

tothtamas28 (changes requested):

src/tests/conftest.py
tothtamas28: Exactly, looks pretty good.

src/pyk/proof/reachability.py
```diff
@@ -587,6 +587,59 @@ def construct_node_refutation(self, node: KCFG.Node) -> RefutationProof | None:
         self.proof.add_subproof(refutation)
         return refutation
 
+    def failure_info(self) -> APRFailureInfo:
+        return APRFailureInfo(self.proof, self.kcfg_explore)
+
+
+class APRFailureInfo:
```
tothtamas28, line 594: Consider making this an immutable dataclass:

```python3
@dataclass(frozen=True)
class APRFailureInfo:
    failing: FrozenDict[int, tuple[str, str]]
    pending: frozenset[int]

    def __init__(self, failing: Mapping[int, tuple[str, str]], pending: iterable[int])
         ...

    @staticmethod
    def from_proof(proof: APRProof, explore: KCFGExpore) -> APRFailureInfo:
        ...
```

tothtamas28 (approved):

src/pyk/proof/reachability.py
nwatson22: Done. Do the changes look ok?

tothtamas28: Looks good. Consider adding `__init__` with a slightly weaker signarutre:

```python
def __init__(self, failing_nodes: Mapping[int, tuple[str, str]], pending_nodes: iterable[int]):
    object.__setattr__(self, 'failing_nodes', FrozenDict(failing_nodes))
    object.__setattr__(self, 'pending_nodes', frozenset(pending_nodes))
```

src/pyk/proof/reachability.py
```diff
@@ -587,6 +588,62 @@ def construct_node_refutation(self, node: KCFG.Node) -> RefutationProof | None:
...
+            simplified_node, _ = kcfg_explore.cterm_simplify(node.cterm)
+            simplified_target, _ = kcfg_explore.cterm_simplify(target.cterm)
+            node_cterm = CTerm.from_kast(simplified_node)
+            target_cterm = CTerm.from_kast(simplified_target)
+            _, reason = kcfg_explore.implication_failure_reason(node_cterm, target_cterm)
+            path_condition = kcfg_explore.kprint.pretty_print(proof.path_constraints(node.id))
+            failing_nodes[node.id] = (reason, path_condition)
+        return APRFailureInfo(failing_nodes=FrozenDict(failing_nodes), pending_nodes=frozenset(pending_nodes))
+
+    def print(self) -> list[str]:
```
tothtamas28, line 615: What's missing is a few simple unit tests for this method.

Once that's done, the original tests will be refactored into:
* A heavier integration test that compares data structures, hence more robust;
* A unit test that does string comparison but runs in an instant, hence easy to iterate on.

Also, this will be the only place where we need to compare these string outputs, so once this PR is merged, the `evm-semantics` test suite can be refactored accordingly.

src/pyk/proof/reachability.py
nwatson22: Done.

src/pyk/proof/reachability.py
nwatson22: Added a test for `print()`.

PR #519, opened 2023-06-30
Add `cterm_get_model` to `KCFGExplore` for counterexample generation

palinatolmach:
Closes the pyk-related part of runtimeverification/evm-semantics#1927, which is aimed to generate and output concrete assignment to variables corresponding to the failed Foundry test.

Adds `cterm_get_model(cterm: CTerm) -> Subst | None` function to `KCFGExplore`, which returns a variable substitution identified via the `get-model` RPC request. Also adds two IMP tests that check the output of `cterm_get_model(cterm)` (one of a successful call, one of an unsuccessful).


src/pyk/kcfg/explore.py
```diff
@@ -176,6 +176,32 @@ def kast_simplify(self, kast: KInner) -> tuple[KInner, tuple[LogEntry, ...]]:
...
+            model_subst = self.kprint.kore_to_kast(result.model)
+            _subst: dict[str, KInner] = {}
+            for subst_pred in flatten_label('#And', model_subst):
+                subst_pattern = mlEquals(KVariable('###VAR'), KVariable('###TERM'))
+                m = subst_pattern.match(subst_pred)
+                if m is not None and type(m['###VAR']) is KVariable:
+                    _subst[m['###VAR'].name] = m['###TERM']
+                else:
+                    raise AssertionError(f'Received a non-substitution from get-model endpoint: {subst_pred}')
+            return Subst(_subst)
```
ehildenb, line 201: I guess this logic can be factored out as `Subst.from_pred` near here: https://github.com/runtimeverification/pyk/blob/49d861944bad260be137f68f4adc37b360e25771/src/pyk/kast/inner.py#L127

ehildenb (approved):

src/pyk/kcfg/explore.py
palinatolmach: Thank you @ehildenb! I tried, but it generates several circular dependency issues. Would you suggest to leave it as it, or define `from_pred(subst_pred: KInner) -> Subst` elsewhere?

PR #522, opened 2023-07-04
Improve `anti_unify_with_constraints`

nwatson22:
Needed for https://github.com/runtimeverification/evm-semantics/pull/1934.
Adds an additional option to `anti_unify_with_constraints` which adds an additional constraint which is an `_orBool_` of the two anti-unifying substitutions.
Adds a test for the method with this option which tests a new variable is created to abstract and the `orBool` constraint is added as described.


src/pyk/kast/inner.py
```diff
@@ -136,6 +136,21 @@ def ml_pred(self) -> KInner:
...
         return ml_term
 
+    @property
+    def pred(self) -> KInner:
+        items = []
+        for k in self:
+            val = self[k]
+            if type(val) is not KVariable or val.name != KVariable(k).name:
+                items.append(KApply('_==K_', [KVariable(k), val]))
+
```
nwatson22, line 146:
```suggestion
```

src/pyk/kast/inner.py
```diff
@@ -136,6 +136,20 @@ def ml_pred(self) -> KInner:
...
+        for k in self:
+            val = self[k]
+            if type(val) is not KVariable or val.name != KVariable(k).name:
+                items.append(KApply('_==K_', [KVariable(k), val]))
+        if len(items) == 0:
+            return KToken('true', 'Bool')
+        term = items[0]
+        for _i in items[1:]:
+            term = KApply('_andBool_', [term, _i])
+        return term
```
tothtamas28, line 151: Consider using `functools.reduce`:

```suggestion
        conjuncts = [
            KApply('_==K_', KVariable(name), val) for name, val in self.items()
            if type(val) is not KVariable or val.name != k
        ]
        if not items:
            return KToken('true', 'Bool')
        return reduce(KLabel('__andbool__'), conjuncts)
```

(Not tested.)

src/pyk/kast/manip.py
```diff
@@ -598,8 +598,18 @@ def _rewrites_to_abstractions(_kast: KInner) -> KInner:
...
+    constrained_term_2: KInner,
+    implications: bool = False,
+    constraint_disjunct: bool = False,
+    abstracted_disjunct: bool = False,
 ) -> KInner:
+    def disjunction_from_substs(substs: list[Subst]) -> KInner:
+        term = substs[0].pred
+        for _i in substs[1:]:
+            term = KApply('_orBool_', [term, _i.pred])
+        return mlEqualsTrue(term)
```
tothtamas28, line 611: It looks like the special case of `len(substs) == 2` is sufficient here:

```suggestion
    def disjunction_from_substs(subst1: Subst, subst2: Subst) -> KInner:
        return mlEqualsTrue(KApply('_orBool_', (subst1.pred, subst2.pred)))
```

tothtamas28 (approved):

src/pyk/kast/manip.py
tothtamas28: Also, if you want as compact a constraint as possible, you need to handle the special case of a `subst` being `true`.

src/pyk/kast/inner.py
nwatson22: Now using `reduce`.

src/pyk/kast/manip.py
nwatson22: Simplified as suggested, and also handles the case where `subst.pred` is true.

PR #523, opened 2023-07-05
Add parameter `verbose` to `kompile`

### LINKED ISSUE #521: Showing `verbose` output for `kompile`
At the moment, the `--verbose` option in `kevm foundry-kompile` is not showing the detailed output for the `kompile` command that is invoked within `foundry-kompile`. Consider adding the ability to show verbose output for `kompile` as well.

tothtamas28:
Fixes #521


radumereuta (commented):
Is the output from kompile collected and printed at the end?
Or is it printed immediately?
In make you have the `-o, --output-sync` option to choose.
Since we're running tasks in sequence, I don't think this matters, but if we're to add parallel compilation, then it would be helpful.

tothtamas28:
> Is the output from kompile collected and printed at the end?

Modified accordingly.

palinatolmach (approved):

PR #524, opened 2023-07-05
Evaluate pending nodes in parallel

iFrostizz:
When having a lot of branching, it may take a lot of time to evaluate proofs on pending nodes because this is done sequentially.
In this case, we can run them on the pending leaves by running the proofs in parallel at a small cost.
This feature adds a new parameter `max_workers` with a default of 1 for the old behavior in the `advance_proof` of the `Prover` and will cap the amount of proving processes running in parallel whenever there are these pending nodes.

### DIFF
--- a/src/pyk/kcfg/explore.py
+++ b/src/pyk/kcfg/explore.py
@@ -55,7 +55,7 @@ class KCFGExplore(ContextManager['KCFGExplore']):
     _bug_report: BugReport | None
 
     _kore_server: KoreServer | None
-    _kore_clients: list[KoreClient]
+    _kore_clients: list[KoreClient] = []
     _rpc_closed: bool
     _trace_rewrites: bool
 
--- a/src/pyk/kcfg/explore.py
+++ b/src/pyk/kcfg/explore.py
@@ -55,7 +55,7 @@ class KCFGExplore(ContextManager['KCFGExplore']):
     _bug_report: BugReport | None
 
     _kore_server: KoreServer | None
-    _kore_client: KoreClient | None
+    _kore_clients: list[KoreClient]
     _rpc_closed: bool
     _trace_rewrites: bool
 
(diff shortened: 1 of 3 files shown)


iFrostizz:
This may lead to significant efficiency improvements in proofs that are branching a lot !
For the `ArithmeticTest.test_max1` which has one branch:

Before (one kore client executing requests sequentially):
image

After (with 3 clients in parallel, although only two were used):
image

tothtamas28 (commented):
Thank you for this PR. Parallelizing KCFG exploration is a feature we've been contemplating for a long time but so far no one actually attempted to do it. This is a step in the right direction.

Based on your submission though I now think that introducing this feature in the codebase might require further planning, and possibly multiple enabling changes. Below are a few thoughts I have on the issue.

---

1. We need an approach that makes locking in the client unnecessary. Each thread in the pool should be assigned its own server (to be more precise, a port number to a running server). Instantiating a client is not a big overhead so it can be done on-demand inside the handler. We can even introduce an abstraction for this so that algorithms can be coded against it (e.g. `KoreServerPool`).

2. Class `KCFGExplore` has too many responsibilities. On one hand, it implements logic, on the other hand, it manages resources. Resource management should be factored out of this class (i.e. `KCFGExplore` should receive a `KoreClient` as a constructor argument). If this is done, instantiating `KCFGExplore` inside the handler will also become an option.

3. We need a thread-safe data model. It must be ensured that no race condition occurs on a data write.

4. If I understand correctly, parallelism in this PR is utilized on branching, i.e. at each moment, nodes submitted to the pool have a common parent. The ideal model is where a pending node is submitted to the pool as soon as it is discovered. (This is remark for the future though, the branching parallelism is already an improvement.)
```

tothtamas28:
@ehildenb, this is probably interesting for you too.

PR #527, opened 2023-07-07
Refactor KoreClientError

anvacaru:
Closes: https://github.com/runtimeverification/evm-semantics/issues/1943

The implementation of `KoreClientError` interfered with the pickling process required for inter-process communication.

Modified the `KoreClientError` class to avoid the use of `object.__setattr__()`


src/pyk/kore/rpc.py
```diff
@@ -164,17 +164,12 @@ def _check(response: Mapping[str, Any]) -> None:
 
 
 @final
-@dataclass(frozen=True)
+@dataclass()
```
tothtamas28, line 167:
```suggestion
@dataclass
```

src/pyk/kore/rpc.py
```diff
@@ -164,17 +164,12 @@ def _check(response: Mapping[str, Any]) -> None:
...
 class KoreClientError(Exception):  # TODO refine
-    message: str
-    code: int
-    data: Any
-
-    def __init__(self, message: str, code: int, data: Any = None):
-        object.__setattr__(self, 'code', code)
-        object.__setattr__(self, 'message', message)
-        object.__setattr__(self, 'data', data)
+    def __init__(self, message: str = '', code: int = 1, data: Any = None):
```
tothtamas28, line 169:
```suggestion
    def __init__(self, message: str, code: int, data: Any = None):
```

src/pyk/kore/rpc.py
tothtamas28, line 1: Note that `JsonRpcError` is also a frozen `dataclass`. If the exception trace is also pickled, this might cause the same issue.

anvacaru: Refactored `JsonRpcError` as well.

src/pyk/kore/rpc.py
anvacaru: Done.

tothtamas28 (approved):

PR #528, opened 2023-07-07
Add support for vacuous nodes

iFrostizz:
Advances towards fixing https://github.com/runtimeverification/evm-semantics/issues/1944
We allow marking nodes as "vacuous" if they match the `is_vacuous()` rule, avoiding to check their subsumption into the target state.


tothtamas28 (commented):
I feel like `vacuous` is semantic information that based on recent changes should go into proof classes. @ehildenb, what's your opinion?

ehildenb:
Yes, let's store `vacuous` in the `APRProof` datastructure instead.

- Class `APRProof` gets new `_vacuous_nodes` set (similar to `_terminal_nodes`), and `is_vacuous` predicate. Make sure to serialize/deserialize the `_vacuous_nodes` set in `dict` and `from_dict`.
- Modify `is_pending` to not include nodes that have `is_vacuous` true about them, and similar for `is_failing`. This must happen for both `APRProof` and `APRBMCProof`.
- Modify `summary` to include the vacuous node count, this needs to happen in both `APRProof` and `APRBMCProof`.
- `APRProof.advance_proof` can check [here](https://github.com/runtimeverification/pyk/blob/b055dbbd749c10a2b7afef97b6457d725494fc02/src/pyk/proof/reachability.py#L492) if a node is vacuous, by calling `KCFGExplore.cterm_simplify` on it, and seeing if it simplifies to `#Bottom`. If it does, then it adds that node id to the proofs `_vacuous_nodes` list.
- Modify `APRProofNodePrinter.node_attrs` to include `vacuous` as a node attribute (as you've already done for `NodePrinter`, but that can be removed now).

If we decide later that this deserves to be first-class information in `KCFG`, we can make that change then. But for now let's store in in `APRProof`. And now we don't need the user to supply an `is_vacuous` predicate, we can already determine that directly with `KCFGExplore.cterm_simplify`.

PR #531, opened 2023-07-11
getModel(): do not crash on empty substitution

h0nzZik:
The [documentation](https://github.com/runtimeverification/haskell-backend/blob/89b42ddbcaa050228c13a27edaffd68e5ce04402/docs/2022-07-18-JSON-RPC-Server-API.md) for the `get-model` endpoint says:
> Exception: if the predicates are trivially satisfied without any variable assignments, the trivial (empty) substitution is omitted.

This PR handles that case.


src/pyk/kore/rpc.py
```diff
@@ -491,7 +491,7 @@ def from_dict(dct: Mapping[str, Any]) -> GetModelResult:
             case 'Unsat':
                 return UnsatResult()
             case 'Sat':
-                return SatResult(model=kore_term(dct['substitution'], Pattern))  # type: ignore
+                return SatResult(model=kore_term(dct.get('substitution'), Pattern))  # type: ignore
```
tothtamas28, line 494:
```suggestion
                substitution = dct.get('substitution')
                return SatResult(model=kore_term(substitution, Pattern)) if substitution else None  # type: ignore
```

tothtamas28 (approved):

src/pyk/kore/rpc.py
tothtamas28: Otherwise, `kore_term` might be called with `None`.

src/pyk/kore/rpc.py
```diff
@@ -491,7 +491,7 @@ def from_dict(dct: Mapping[str, Any]) -> GetModelResult:
             case 'Unsat':
                 return UnsatResult()
             case 'Sat':
-                return SatResult(model=kore_term(dct.get('substitution'), Pattern))  # type: ignore
+                return SatResult(model=kore_term(dct['substitution'], Pattern) if 'substitution' in dct else None)  # type: ignore
```
tothtamas28, line 494: The other version is more robust as it handles

```json
{"substitution": null}
```

PR #532, opened 2023-07-12
Store nodes as separate JSON files

nwatson22:
Attempts to improve performance: there was a possible bottleneck where it was taking too much time every block to write the entire KCFG file when a new node was added. To rectify this, we are now storing the nodes each in their own file `[proof_id]/kcfg/[node_id].json`. Each node should only be written to when created or updated. The functions `read_node_data` and `write_node_data` read from/write to the new proof storage layout. The `omit_nodes` option can be included to get this faster save, provided the nodes have been saved individually.


src/pyk/kcfg/kcfg.py
```diff
@@ -196,6 +204,7 @@ def edges(self) -> tuple[KCFG.Edge, ...]:
 
     _node_id: int
     _nodes: dict[int, Node]
+    _node_digests: dict[int, str]
```
ehildenb, line 207: Do we need the `_node_digests` dict? Since the `Node` itself is carrying a `Node.digest`, is it really necessary?

We can always instead of doing `self._node_digests[nid]` do `self._nodes[nid].digest`, right?

RaoulSchaffranek:
Hey Noah, I tried the PR on my example and can see the new proof artifacts. However, the old artifact is also still present. Consequently, it caused a performance regression on my computer.

Notice that I'm using the bounded model checker. It seems that `APRBMCProver` is still calling the old `write_proof`. https://github.com/runtimeverification/pyk/blob/c190dead61c29266fe50d2424894826d0cb66a19/src/pyk/proof/reachability.py#L752

Also, the APRBMCProver saves the proof three times per step. 
The first time, before it explores the next pending notes:
https://github.com/runtimeverification/pyk/blob/c190dead61c29266fe50d2424894826d0cb66a19/src/pyk/proof/reachability.py#L752
And then two times when it calls into the parent `APRProver`:
Here:https://github.com/runtimeverification/pyk/blob/c190dead61c29266fe50d2424894826d0cb66a19/src/pyk/proof/reachability.py#L548
https://github.com/runtimeverification/pyk/blob/c190dead61c29266fe50d2424894826d0cb66a19/src/pyk/proof/reachability.py#L587

src/pyk/kcfg/kcfg.py
```diff
@@ -214,6 +217,9 @@ def __init__(self) -> None:
         self._stuck = set()
         self._aliases = {}
         self._lock = RLock()
+        self.cfg_dir = cfg_dir
+        if self.cfg_dir is not None and not self.cfg_dir.exists():
+            self.cfg_dir.mkdir()
```
ehildenb, line 222:
```suggestion
        self.cfg_dir = cfg_dir
        if self.cfg_dir is not None:
            ensure_dir_path(self.cfg_dir)
```

This can be imported as: `from ..utils import ensure_dir_path`

src/pyk/kcfg/kcfg.py
```diff
@@ -268,8 +274,8 @@ def uncovered(self) -> list[Node]:
         return [node for node in self.nodes if not self.is_covered(node.id)]
 
     @staticmethod
-    def from_claim(defn: KDefinition, claim: KClaim) -> tuple[KCFG, NodeIdLike, NodeIdLike]:
-        cfg = KCFG()
+    def from_claim(defn: KDefinition, claim: KClaim, cfg_dir: Path) -> tuple[KCFG, NodeIdLike, NodeIdLike]:
+        cfg = KCFG(cfg_dir=cfg_dir)
```
ehildenb, line 278: Can we make `cfg_dir` optional here instead? Is there a reason its mandatory?

```suggestion
    def from_claim(defn: KDefinition, claim: KClaim, cfg_dir: Path | None = None) -> tuple[KCFG, NodeIdLike, NodeIdLike]:
        cfg = KCFG(cfg_dir=cfg_dir)
```

src/pyk/kcfg/kcfg.py
nwatson22: Reworked to no longer use digests, as @tothtamas28 suggested to go with the simpler route of just storing the nodes under their IDs directly.

src/pyk/kcfg/kcfg.py
nwatson22: Made it optional.

nwatson22:
@RaoulSchaffranek It should work with APRBMC proofs now, and I removed the call to `write_proof` in `APRBMCProver.advance_proof()` since it is already getting called in the parent method.

src/pyk/kcfg/kcfg.py
```diff
@@ -434,6 +444,7 @@ def create_node(self, cterm: CTerm) -> Node:
         node = KCFG.Node(self._node_id, cterm)
         self._node_id += 1
         self._nodes[node.id] = node
+        self.write_node_data(node)
         return node
 
     def remove_node(self, node_id: NodeIdLike) -> None:
```
tothtamas28, line 450: Should this call `delete_node_data`?

src/pyk/kcfg/kcfg.py
```diff
@@ -860,3 +872,102 @@ def reachable_nodes(
...
+        if self.cfg_dir is None:
+            return
+        nodes_dir = self.cfg_dir / 'nodes'
+        ensure_dir_path(nodes_dir)
+        node_json = nodes_dir / (str(node.id) + '.json')
+        node_dict = node.to_dict()
+        node_json.write_text(json.dumps(node_dict))
+
+    def delete_node_data(self, node: KCFG.Node) -> None:
+        assert self.cfg_dir is not None
```
tothtamas28, line 913: There's an asymmetry here: if `self.cfg_dir is None`, method `write_node_data` will do nothing, whereas `delete_node_data` will raise `AssertionError`.

tothtamas28: Are these functions intentionally public? My understanding is that calling them can mess up consistency.

src/pyk/kcfg/kcfg.py
tothtamas28, line 1: What are the intended ways to interact with the class after this change?

My understanding is that if `cfg_dir` is not specified, then it will not touch the disk on modifications, so that's basically the same behavior as before. Can such an instance still be saved to disk though?

Suppose`cfg_dir` is set. What assumptions can one make w.r.t. the consistency between the data structure in memory and on disk? Or is there a sequence of methods that has to be called to achieve that?

src/pyk/kcfg/kcfg.py
nwatson22: I didn't think so. If there is already a node with that ID, which there shouldn't be, it would just be overwritten by `write_node_data`.

src/pyk/kcfg/kcfg.py
nwatson22: >  My understanding is that if cfg_dir is not specified, then it will not touch the disk on modifications, so that's basically the same behavior as before. Can such an instance still be saved to disk though?

I'm pretty sure this is correct. I think if you changed your mind later about not caring about saving to disk you could set the cfg_dir field to a value and then call `write_cfg_data`.

> Suppose cfg_dir is set. What assumptions can one make w.r.t. the consistency between the data structure in memory and on disk? Or is there a sequence of methods that has to be called to achieve that?

Assuming I implemented it right, you do need to call `write_cfg_data` to insure consistency of the kcfg.json file. The node files should be automatically consistent assuming you modify the node structure only through `KCFG`'s public interface. So, now that you mention it, there isn't really a need for the `omit_nodes` option.

src/pyk/kcfg/kcfg.py
nwatson22: Is the convention to just prefix "private" methods with `_`? I did so to both, and changed the inconsistency where now `delete_node_data` will now also just pass.

src/pyk/kcfg/kcfg.py
tothtamas28: Sorry for the confusion. My question is: for `remove_node`, shouldn't `delete_node_data`be called?

src/pyk/kcfg/kcfg.py
tothtamas28: So based on my understanding, these are the correct ways to interact with the class:

### Non-persistent mode

```python3
kcfg = KCFG()
kcfg.create_node(...)
kcfg.create_edge(...)
```

### Persistent mode, create node

```python3
kcfg = KCFG(kcfg_dir)
kcfg.create_node(...)  # node file is written
kcfg.write_cfg_data(omit_nodes=True)  # the node id still has to be written
```

### Persistent mode, create edge

```python3
kcfg = KCFG(kcfg_dir)
kcfg.create_edge(...)
kcfg.write_cfg_data(omit_nodes=True)  # the edge data has to be written
```

### Saving a non-persistent KCFG
```python3
kcfg = KCFG()
...
kcfg.cfg_dir = cfg_dir
kcfg.write_cfg_data(omit_nodes=False)
```

(In this last scenario, `cfg_dir` might not exist, so it would be better to expose a setter that ensures it is created.)

tothtamas28: I do not want to block this PR due to the urgency of the feature, so this is just a suggestion.

You can simplify the interface if the data structure tracks the diff from the persistent state. So for example:

```python3
cfg = read_cfg(cfg_path)  # cfg is in a clean state

cfg.create_node(1)  # _created = {1}
cfg.create_node(2)  # _created = {1, 2}
cfg.create_node(3)  # _created = {1, 2, 3}
cfg.create_edge(1, 2)  # _created = {1, 2, 3}, _dirty = True
cfg.remove_node(3)  # _created = {1, 2}, _dirty = True
cfg.remove_node(4)  # _created = {1, 2}, _dirty = True, _removed = {4}

# Creates node files for __created.
# Removes node files for __removed.
# Also writes the main JSON file as (_dirty or _created or _removed)
cfg.write_cfg(cfg_path)  # cfg is in a clean state
```

which is the exact same interface that we used to have in the single file setting.

src/pyk/kcfg/kcfg.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import os
```
tothtamas28, line 4: Where possible, use the corresponding `pathlib` function: https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

src/pyk/kcfg/kcfg.py
nwatson22: Oh, sorry. I see. It deletes it on disk now.

src/pyk/kcfg/kcfg.py
nwatson22: I changed it so it now only saves nodes on `write_cfg_data`, but only those nodes that were marked for creation or deletion, as suggested. The `omit_nodes` option has been removed, so now except that detail the example you gave is the correct interaction.

src/pyk/kcfg/kcfg.py
nwatson22: Fixed.

src/pyk/kcfg/kcfg.py
tothtamas28: The node has to be marked for writing on modifications as well. Are those cases handled?

tothtamas28: This would be a great target for unit testing: you mock disk access (`_write_node_data` in particular), and check interactions on the mock.

tothtamas28 (approved):
Before merging, please check if this change works with KEVM-Foundry.

src/pyk/kcfg/kcfg.py
nwatson22: On modifications (`replace_node()`) it is added to the `_created_nodes` set. It should work with just `_created_nodes` and `_deleted_nodes` since node creation will overwrite the old node file.
I've also added a test for `KCFG.write_cfg_data()` which tracks `_write_node_data` and `_delete_node_data` calls.

src/pyk/kcfg/kcfg.py
```diff
@@ -860,3 +878,113 @@ def reachable_nodes(
...
+        dct['edges'] = edges
+        dct['covers'] = covers
+        dct['splits'] = splits
+        dct['ndbranches'] = ndbranches
+        dct['stuck'] = stuck
+        dct['aliases'] = aliases
+        cfg_json.write_text(json.dumps(dct))
+
+        print(self.nodes)
+        print(self._deleted_nodes)
```
nwatson22, line 906:
```suggestion
```

RaoulSchaffranek:
@nwatson22 I've tested again with the symbolic debugger - and it works now 🎉

src/pyk/proof/equality.py
```diff
@@ -42,6 +44,19 @@ def __init__(
...
+        if not self.proof_dir:
+            return
+        ensure_dir_path(self.proof_dir)
+        ensure_dir_path(self.proof_dir / self.id)
+        proof_path = self.proof_dir / self.id / 'proof.json'
+        if not self.up_to_date:
+            proof_json = json.dumps(self.dict)
+            proof_path.write_text(proof_json)
+            _LOGGER.info(f'Updated proof file {self.id}: {proof_path}')
+            print(proof_path)
```
nwatson22, line 58:
```suggestion
```

src/pyk/proof/reachability.py
```diff
@@ -255,6 +272,65 @@ def summary(self) -> CompositeSummary:
...
+            circularity=circularity,
+            admitted=admitted,
+            proof_dir=proof_dir,
+            subproof_ids=subproof_ids,
+            node_refutations=node_refutations,
+        )
+
+    def write_proof_data(self) -> None:
+        print('in APRProof.write_proof_data')
+        print(self.id)
```
nwatson22, line 307:
```suggestion
```

src/pyk/proof/reachability.py
```diff
@@ -292,6 +366,68 @@ def __init__(
...
+            bounded_nodes=bounded_nodes,
+            bmc_depth=bmc_depth,
+            proof_dir=proof_dir,
+            subproof_ids=subproof_ids,
+            node_refutations=node_refutations,
+        )
+
+    def write_proof_data(self) -> None:
+        print('in APRBMCProof.write_proof_data')
+        print(self.id)
```
nwatson22, line 403:
```suggestion
```

nwatson22:
I tested with the foundry-* commands on https://github.com/runtimeverification/evm-semantics/pull/1956 and confirmed that it is working with kevm.

PR #533, opened 2023-07-12
Use a tuple to concretize generator

iFrostizz:
Updating `map_inner` functions to pass a `tuple` to avoid reaching recursion depth limits caused by generators.


src/pyk/kast/inner.py
```diff
@@ -433,7 +433,7 @@ def let(self, *, label: str | KLabel | None = None, args: Iterable[KInner] | Non
         return KApply(label=label, args=args)
 
     def map_inner(self: KApply, f: Callable[[KInner], KInner]) -> KApply:
-        return self.let(args=(f(arg) for arg in self.args))
+        return self.let(args=(tuple(f(arg) for arg in self.args)))
```
tothtamas28, line 436:
```suggestion
        return self.let(args=tuple(f(arg) for arg in self.args))
```

src/pyk/kast/inner.py
```diff
@@ -638,7 +638,7 @@ def let(self, *, items: Iterable[KInner] | None = None) -> KSequence:
         return KSequence(items=items)
 
     def map_inner(self: KSequence, f: Callable[[KInner], KInner]) -> KSequence:
-        return self.let(items=(f(item) for item in self.items))
+        return self.let(items=(tuple(f(item) for item in self.items)))
```
tothtamas28, line 641:
```suggestion
        return self.let(items=tuple(f(item) for item in self.items))
```

tothtamas28 (approved):

PR #534, opened 2023-07-12
refactor: summary of proofs

iFrostizz:
Make `Proof.summary` returning a proper data structure rather than a `Iterable[str]` will allow us to harden tests such as `foundry-list`: https://github.com/runtimeverification/evm-semantics/pull/1952


src/pyk/proof/equality.py
```diff
@@ -351,3 +355,47 @@ def advance_proof(self) -> None:
         super().advance_proof()
         assert type(self.proof) is RefutationProof
         self.proof.set_simplified_constraints(self.proof.simplified_consequent)
+
+
+class EqualitySummary(ProofSummary[EqualityProof]):
```
tothtamas28, line 360: You can make these classes a `dataclass`.

src/pyk/proof/equality.py
```diff
@@ -338,6 +331,17 @@ def advance_proof(self) -> None:
...
 
+class ProofSummary(Generic[T], ABC):
+    @staticmethod
+    @abstractmethod
+    def from_proof(proof: T) -> ProofSummary[T]:
+        ...
+
+    @abstractmethod
+    def to_str(self) -> Iterable[str]:
+        ...
```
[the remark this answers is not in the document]
iFrostizz: Thank you!

src/pyk/proof/equality.py
```diff
@@ -178,12 +178,12 @@ def pretty(self, kprint: KPrint) -> Iterable[str]:
         return lines
 
     @property
-    def summary(self) -> Iterable[str]:
-        return [
-            f'EqualityProof: {self.id}',
-            f'    status: {self.status}',
-            f'    admitted: {self.admitted}',
-        ]
+    def summary(self) -> CompositeSummary:
```
tothtamas28, line 181: Method `summary` used to return a single `EqualityProof: ...` section, returning `CompositeSummary` with subproofs modifies that behavior.

src/pyk/proof/equality.py
```diff
@@ -258,11 +258,12 @@ def from_dict(cls: type[RefutationProof], dct: Mapping[str, Any], proof_dir: Pat
...
-        return [
-            f'RefutationProof: {self.id}',
-            f'    status: {self.status}',
-        ]
+    def summary(self) -> CompositeSummary:
+        subproofs_summaries = [subproof.summary for subproof in self.subproofs]
+        return CompositeSummary(
+            RefutationSummary(self.id, self.status),
+            subproofs_summaries,
+        )
```
tothtamas28, line 266: Same here.

src/pyk/proof/equality.py
```diff
@@ -351,3 +352,31 @@ def advance_proof(self) -> None:
         super().advance_proof()
         assert type(self.proof) is RefutationProof
         self.proof.set_simplified_constraints(self.proof.simplified_consequent)
+
+
+@dataclass
+class EqualitySummary(ProofSummary):
```
tothtamas28, line 358: Move this right under `EqualityProof`.

src/pyk/proof/proof.py
```diff
@@ -175,9 +176,29 @@ def json(self) -> str:
...
+
+    @property
+    def lines(self) -> list[str]:
+        return [f'Proof: {self.id}', f'    status: {self.status}']
+
+    def __str__(self) -> str:
+        return '\n'.join(self.lines)
+
+
+class CompositeSummary:
```
tothtamas28, line 197: `CompositeSummary` should implement of `ProofSummary`. It can also be a `dataclass`.

src/pyk/proof/proof.py
```diff
@@ -175,9 +176,29 @@ def json(self) -> str:
         return json.dumps(self.dict)
 
     @property
-    def summary(self) -> Iterable[str]:
+    def summary(self) -> CompositeSummary:
```
tothtamas28, line 179: This should be `ProofSummary`.

src/pyk/proof/reachability.py
```diff
@@ -722,3 +724,62 @@ def advance_proof(
 
         self.proof.write_proof()
         return self.proof.kcfg
+
+
+@dataclass
+class APRSummary(ProofSummary):
```
tothtamas28, line 730: We need some unit tests to make sure `summary` output is not broken for these `Proof` classes.

src/pyk/proof/equality.py
```diff
@@ -1,26 +1,26 @@
...
     from collections.abc import Iterable, Mapping
     from pathlib import Path
-    from typing import Any, Final, TypeVar
 
     from ..kast.outer import KClaim, KDefinition
     from ..kcfg import KCFGExplore
     from ..ktool.kprint import KPrint
 
-    T = TypeVar('T', bound='Proof')
+T = TypeVar('T', bound='Proof')
```
tothtamas28, line 23: Is this used anywhere?

src/pyk/proof/equality.py
```diff
@@ -178,13 +178,27 @@ def pretty(self, kprint: KPrint) -> Iterable[str]:
...
+    @property
+    def lines(self) -> list[str]:
         return [
             f'EqualityProof: {self.id}',
             f'    status: {self.status}',
             f'    admitted: {self.admitted}',
         ]
 
+    def __str__(self) -> str:
+        return '\n'.join(self.lines)
```
tothtamas28, line 200: No need to override, superclass defines it the same way.

src/pyk/proof/proof.py
```diff
@@ -175,9 +176,37 @@ def json(self) -> str:
         return json.dumps(self.dict)
 
     @property
-    def summary(self) -> Iterable[str]:
+    def summary(self) -> ProofSummary:
         subproofs_summaries = [subproof.summary for subproof in self.subproofs]
-        return chain([f'Proof: {self.id}', f'    status: {self.status}'], *subproofs_summaries)
+        return CompositeSummary(ProofSummary(self.id, self.status), subproofs_summaries)
```
tothtamas28, line 181: I agree this is a bit weird. I'd resolve it as follows:

Keep the interface minimal:

```python
class ProofSummary(ABC):

    @property
    @abstractmethod
    def lines(self) -> list[str]:
        ...

    def __str__(self) -> str:
        return '\n'.join(self.lines)
```

Note how, `ProofSummary is truly abstract, there's no way to instantiate it..

Then create a minimal implementation:

```python3
@dataclass
class BaseSummary(ProofSummary):
    id: str
    status: ProofStatus

    @property
    def lines(self) -> list[str]:
        return [f'Proof: {self.id}', f'    status: {self.status}']
```

then

```suggestion
        return CompositeSummary(BaseSummary(self.id, self.status), subproofs_summaries)
```

You can even make `_BaseSummary` private or nest it within `Proof.summary`.

src/pyk/proof/proof.py
```diff
@@ -175,9 +176,37 @@ def json(self) -> str:
...
+        return '\n'.join(self.lines)
+
+
+@dataclass
+class CompositeSummary(ProofSummary):
+    id: str
+    status: ProofStatus
+    summaries: tuple[ProofSummary, ...]
+
+    def __init__(self, summary: ProofSummary, subproofs_summaries: Iterable[ProofSummary]):
```
tothtamas28, line 203: Consider

```suggestion
    def __init__(self, summaries: Iterable[ProofSummary]):
```

src/pyk/proof/proof.py
```diff
@@ -175,9 +176,37 @@ def json(self) -> str:
...
+        self.id = summary.id
+        self.status = summary.status
+        self.summaries = tuple(chain([summary], subproofs_summaries))
+
+    def __str__(self) -> str:
+        return '\n'.join(str(summary) for summary in self.summaries)
```
tothtamas28, line 209: This should not be overridden. Instead, implement `lines`.

src/pyk/proof/equality.py
```diff
@@ -178,7 +176,18 @@ def pretty(self, kprint: KPrint) -> Iterable[str]:
         return lines
 
     @property
-    def summary(self) -> Iterable[str]:
+    def summary(self) -> EqualitySummary:
+        return EqualitySummary(self.id, self.status, self.admitted)
+
+
+@dataclass
```
tothtamas28, line 183: Can be frozen:

```suggestion
@dataclass(frozen=True)
```

src/pyk/proof/proof.py
```diff
@@ -175,9 +176,52 @@ def json(self) -> str:
...
+            @property
+            def lines(self) -> list[str]:
+                return [f'Proof: {self.id}', f'    status: {self.status}']
+
         subproofs_summaries = [subproof.summary for subproof in self.subproofs]
-        return chain([f'Proof: {self.id}', f'    status: {self.status}'], *subproofs_summaries)
+        return CompositeSummary([_BaseSummary(self.id, self.status), *subproofs_summaries])
+
+
+@dataclass
```
tothtamas28, line 193: Shouldn't be a dataclass.

```suggestion
```

src/pyk/proof/proof.py
```diff
@@ -175,9 +176,52 @@ def json(self) -> str:
...
+class ProofSummary(ABC):
+    id: str
+    status: ProofStatus
+
+    @property
+    @abstractmethod
+    def lines(self) -> list[str]:
+        return [f'Proof: {self.id}', f'    status: {self.status}']
```
tothtamas28, line 201: Let's keep this class proper abstract.

```suggestion
class ProofSummary(ABC):
    @property
    @abstractmethod
    def lines(self) -> list[str]:
        ...
```

src/pyk/proof/proof.py
```diff
@@ -175,9 +176,52 @@ def json(self) -> str:
...
+class CompositeSummary(ProofSummary):
+    id: str
+    status: ProofStatus
+    summaries: tuple[ProofSummary, ...]
+
+    def __init__(self, _summaries: Iterable[ProofSummary]):
+        summary = list(_summaries)[0]
+        self.id = summary.id
+        self.status = summary.status
+        self.summaries = tuple(chain(_summaries))
```
tothtamas28, line 217:
```suggestion
    summaries: tuple[ProofSummary, ...]

    def __init__(self, _summaries: Iterable[ProofSummary]):
        self.summaries = tuple(chain(_summaries))
```

src/pyk/proof/proof.py
```diff
@@ -175,9 +176,52 @@ def json(self) -> str:
...
+
+    def __str__(self) -> str:
+        return '\n'.join(str(summary) for summary in self.summaries)
+
+    @property
+    def lines(self) -> list[str]:
+        return [line for lines in [summary.lines for summary in self.summaries] for line in lines]
```
tothtamas28, line 224:
```suggestion
        return [line for lines in (summary.lines for summary in self.summaries) for line in lines]
```

src/pyk/proof/reachability.py
```diff
@@ -722,3 +728,62 @@ def advance_proof(
...
```
tothtamas28, line 734: Move these classes right under the respective prover classes.

src/pyk/proof/equality.py
iFrostizz: @tothtamas28 This is not possible since writing `EqualitySummary` class as frozen requires to mark `ProofSummary` the same, and thus `CompositeSummary` as well because it inherits from `ProofSummary`, which cannot be frozen because of the custom `__init__`.

tothtamas28: > writing `EqualitySummary` class as frozen requires to mark `ProofSummary` the same

I don't think so, `ProofSummary` shouldn't even be a `dataclass`.

> which cannot be frozen because of the custom `__init__`

It can, you can use `object.__setattr__` to set attributes: https://docs.python.org/3/library/dataclasses.html#frozen-instances

iFrostizz: Makes sense, thank you !

src/tests/unit/test_proof.py
```diff
@@ -279,3 +279,94 @@ def test_print_failure_info() -> None:
 Join the Runtime Verification Discord server for support: https://discord.gg/CurfmXNtbN"""
 
     assert actual_output == expected_output
+
+
+def test_apr_proof_summary(proof_dir: Path) -> None:
```
tothtamas28, line 284: In these test you can assert the instance directly:

```python3
assert proof.summary = CompositeSummary([...])
```

tothtamas28 (approved):

PR #535, opened 2023-07-12
Fix unquantified variables in predicates for krule_to_kore

virgil-serbanuta:
These seem to be generated by the Haskell backend when producing "remainder" predicates of the type #Not (#Exists V . stuff(V))


src/pyk/konvert.py
```diff
@@ -95,10 +95,10 @@ def kast_to_kore(
 
 
 # 'krule' should have sorts on variables
-def krule_to_kore(kompiled_kore: KompiledKore, krule: KRule) -> Axiom:
+def krule_to_kore(kast_defn: KDefinition, kompiled_kore: KompiledKore, krule: KRule) -> Axiom:
```
tothtamas28, line 98: I see this function calls `_kast_to_kore`. Wouldn't calling `kast_to_kore` solve the issue with the variable sorts?

virgil-serbanuta: That's correct, fixed.

src/pyk/konvert.py
```diff
@@ -95,30 +95,32 @@ def kast_to_kore(
...
     krule_body = krule.body
     krule_lhs = CTerm(extract_lhs(krule_body), [bool_to_ml_pred(krule.requires)])
     krule_rhs = CTerm(extract_rhs(krule_body), [bool_to_ml_pred(krule.ensures)])
 
+    top_level_kore_sort = SortApp(name='SortGeneratedTopCell', sorts=())
```
tothtamas28, line 103: This works:

```suggestion
    top_level_kore_sort = SortApp('SortGeneratedTopCell')
```

tothtamas28 (approved):

PR #538, opened 2023-07-14
Implement outer lexer

tothtamas28:
Implements a function `outer_lexer` that returns an iterator over K outer syntax tokens.

The implementation lexes `domains.k` (obtained from `domains.md` by stripping Markdown syntax) in ~0.05s.

There are some small differences compared to the way the K Frontend lexes K input:
* In `BUBBLE` state `'rule/**/'` lexes as a rule token and not as a bubble (for justification see https://github.com/runtimeverification/k/issues/3501)
* In `BUBBLE` and `KLABEL` states a `'/*'` without a closing `'*/'` will not lex. This makes implementation much simpler. Also, even if the lexer allowed this, a block comment following the token would break the file anyways.


ehildenb (approved):

ehildenb:
Merging to unblock, any issues found can be addressed in follow-ups.

PR #542, opened 2023-07-18
Add `KoreServerPool`

tothtamas28:
Adds a class `KoreServerPool` that wraps a `ThreadPoolExecutor` and provides a `KoreServer` instance per thread. As `ThreadPoolExecutor` handles threads economically (i.e. it won't start a new thread if there's a free thread that has been started earlier), only the minimal necessary number of servers will be started (up to `max_workers`).

The pool can be submitted functions that have `int` as first parameter. The server pool will forward the handler to the thread pool with the corresponding server port number substituted for that parameter.


src/tests/integration/kore/test_pool.py
```diff
@@ -0,0 +1,96 @@
...
+        expected = {'', 'a', 'b', 'aa', 'ab', 'ba', 'bb', 'aaa', 'aab', 'aba', 'abb', 'baa', 'bab', 'bba', 'bbb'}
+
+        # When
+        actual = explore(server_pool)
+
+        # Then
+        assert actual == expected
+
+
+def explore(pool: KoreServerPool) -> set[str]:
```
tothtamas28, line 49: Simple example for exploring the semantics in parallel.

ehildenb (approved):

PR #543, opened 2023-07-18
Re-enable `--no-implication-every-step` flag

RaoulSchaffranek:
This PR fixes a regression that caused the `--no-implication-every-step` flag to be ignored.


ehildenb (approved):
@RaoulSchaffranek I assume you tested this with KEVM. If so, feel free to add automerge.

PR #546, opened 2023-07-20
Factoring of semantics-specific functionalities

PetarMax:
This PR creates a KCFGSemantics class, which holds semantics-specific functionalities, specifically:

- the `is_terminal(c)` predicate, which checks whether or not the passed-in node is terminal;
- the `extract_branches(c)` function, which understands whether the passed-in node can be split;
- the `abstract_node(c)` function, which abstracts the passed-in node 
- the `same_loop(c1, c2)` function, which identifies if the two passed-in nodes represent the same loop head;
- the cut-point rules of the semantics; and
- the terminal rules of the semantics;

and refactors the code to use this class appropriately.


deps/k_release
```diff
@@ -1 +1 @@
-6.0.24
+6.0.27
```
ehildenb, line 1: Please back out the change to the K depnedency, we should let the auto-updater take care of that.

src/pyk/kcfg/explore.py
```diff
@@ -60,9 +61,12 @@ class KCFGExplore(ContextManager['KCFGExplore']):
     _rpc_closed: bool
     _trace_rewrites: bool
 
+    semantics: KCFGSemantics
+
     def __init__(
         self,
         kprint: KPrint,
+        semantics: KCFGSemantics,
```
tothtamas28, line 69:
```suggestion
        semantics: KCFGSemantics | None,
```

src/pyk/kcfg/explore.py
```diff
@@ -78,6 +82,7 @@ def __init__(
     ):
         self.kprint = kprint
         self.id = id if id is not None else 'NO ID'
+        self.semantics = semantics
```
tothtamas28, line 85:
```suggestion
        self.semantics = semantics or DefaultSemantics()
```

If the type checker complains:

```suggestion
        self.semantics = semantics if semantics is not None else DefaultSemantics()
```

src/pyk/kcfg/kcfg.py
```diff
@@ -31,7 +31,6 @@
     from ..kast.inner import KInner
     from ..kast.outer import KClaim, KDefinition, KImport, KRuleLike
 
-
```
tothtamas28, line 34: You can opt out of this change to keep the set of changed files minimal.

src/pyk/testing/_kompiler.py
```diff
@@ -128,12 +129,16 @@ def _update_symbol_table(symbol_table: SymbolTable) -> None:
 
 
 class KCFGExploreTest(KProveTest):
+    SEMANTICS: ClassVar[KCFGSemantics]
```
tothtamas28, line 132:
```suggestion
    KCFG_SEMANTICS: ClassVar[KCFGSemantics | None] = None
```

src/tests/integration/kcfg/test_cell_map.py
```diff
@@ -46,13 +47,14 @@ class State(NamedTuple):
...
-APR_PROVE_TEST_DATA: Iterable[tuple[str, Path, str, str, int | None, int | None, Iterable[str]]] = (
-    ('cell-map-no-branch', K_FILES / 'cell-map-spec.k', 'CELL-MAP-SPEC', 'cell-map-no-branch', 2, 1, []),
+APR_PROVE_TEST_DATA: Iterable[tuple[str, Path, str, str, int | None, int | None]] = (
+    ('cell-map-no-branch', K_FILES / 'cell-map-spec.k', 'CELL-MAP-SPEC', 'cell-map-no-branch', 2, 1),
 )
 
 
 class TestCellMapProof(KCFGExploreTest):
     KOMPILE_MAIN_FILE = K_FILES / 'cell-map.k'
+    SEMANTICS = DefaultSemantics()
```
tothtamas28, line 57: After setting the defaults, this should be unnecessary.

src/tests/integration/kcfg/test_equality_proof.py
```diff
@@ -113,6 +114,7 @@
 
 class TestImpEqualityProof(KCFGExploreTest):
     KOMPILE_MAIN_FILE = K_FILES / 'imp-verification.k'
+    SEMANTICS = DefaultSemantics()
```
tothtamas28, line 117: And this.

src/tests/integration/kcfg/test_multiple_definitions.py
```diff
@@ -22,6 +23,7 @@
 
 class TestMultipleDefinitionsProof(KCFGExploreTest):
     KOMPILE_MAIN_FILE = K_FILES / 'multiple-definitions.k'
+    SEMANTICS = DefaultSemantics()
```
tothtamas28, line 26: And this.

src/tests/integration/kcfg/test_non_det.py
```diff
@@ -24,16 +25,17 @@
...
-APR_PROVE_TEST_DATA: Iterable[tuple[str, Path, str, str, int | None, int | None, Iterable[str]]] = (
-    ('test-nondet', K_FILES / 'non-det-spec.k', 'NON-DET-SPEC', 'non-det', 8, 1, []),
+APR_PROVE_TEST_DATA: Iterable[tuple[str, Path, str, str, int | None, int | None]] = (
+    ('test-nondet', K_FILES / 'non-det-spec.k', 'NON-DET-SPEC', 'non-det', 8, 1),
 )
 
 
 class TestNonDetProof(KCFGExploreTest):
     KOMPILE_MAIN_FILE = K_FILES / 'non-det.k'
+    SEMANTICS = DefaultSemantics()
```
tothtamas28, line 35: And this.

src/tests/integration/kcfg/test_simple.py
```diff
@@ -39,6 +40,7 @@
 
 class TestSimpleProof(KCFGExploreTest):
     KOMPILE_MAIN_FILE = K_FILES / 'simple-proofs.k'
+    SEMANTICS = DefaultSemantics()
```
tothtamas28, line 43: And this.

src/tests/integration/kcfg/test_goto.py
```diff
@@ -113,15 +126,13 @@ def test_all_path_bmc_reachability_prove(
         prover = APRBMCProver(
             proof,
             kcfg_explore=kcfg_explore,
-            same_loop=TestGoToProof._same_loop,
-            is_terminal=TestGoToProof._is_terminal,
-            extract_branches=TestGoToProof._extract_branches,
+            same_loop=kcfg_explore.semantics.same_loop,
+            is_terminal=kcfg_explore.semantics.is_terminal,
+            extract_branches=kcfg_explore.semantics.extract_branches,
```
tothtamas28, line 131: This can be made internal to `APRBMCProver`.

src/tests/integration/kcfg/test_refute_node.py
```diff
@@ -116,7 +162,10 @@ def test_apr_proof_refute_node(
             f'{spec_module}.{claim_id}', kcfg_pre, init=init_node, target=target_node, logs={}, proof_dir=proof_dir
         )
         prover = APRProver(
-            proof, kcfg_explore, is_terminal=TestAPRProof._is_terminal, extract_branches=TestAPRProof._extract_branches
+            proof,
+            kcfg_explore,
+            is_terminal=kcfg_explore.semantics.is_terminal,
+            extract_branches=kcfg_explore.semantics.extract_branches,
```
tothtamas28, line 168: This too.

src/tests/integration/kcfg/test_imp.py
```diff
@@ -869,14 +850,13 @@ def qualify(module: str, label: str) -> str:
             prover = APRProver(
                 proof,
                 kcfg_explore=kcfg_explore,
-                is_terminal=TestImpProof._is_terminal,
-                extract_branches=lambda cterm: TestImpProof._extract_branches(kprove.definition, cterm),
+                is_terminal=kcfg_explore.semantics.is_terminal,
+                extract_branches=kcfg_explore.semantics.extract_branches,
```
tothtamas28, line 854: And also this.

src/tests/integration/kcfg/test_imp.py
```diff
@@ -919,15 +897,13 @@ def _node_printer(cterm: CTerm) -> list[str]:
         prover = APRProver(
             proof,
             kcfg_explore=kcfg_explore,
-            is_terminal=TestImpProof._is_terminal,
-            extract_branches=lambda cterm: TestImpProof._extract_branches(kprove.definition, cterm),
+            is_terminal=kcfg_explore.semantics.is_terminal,
+            extract_branches=kcfg_explore.semantics.extract_branches,
```
tothtamas28, line 901: And this.

src/tests/integration/kcfg/test_imp.py
```diff
@@ -965,14 +939,12 @@ def test_all_path_bmc_reachability_prove(
         prover = APRBMCProver(
             proof,
             kcfg_explore=kcfg_explore,
-            same_loop=TestImpProof._same_loop,
-            is_terminal=TestImpProof._is_terminal,
+            same_loop=kcfg_explore.semantics.same_loop,
+            is_terminal=kcfg_explore.semantics.is_terminal,
```
tothtamas28, line 943: And this.

src/tests/integration/kcfg/test_imp.py
```diff
@@ -1010,7 +982,7 @@ def test_failure_info(
         prover = APRProver(
             proof,
             kcfg_explore=kcfg_explore,
-            is_terminal=TestImpProof._is_terminal,
+            is_terminal=kcfg_explore.semantics.is_terminal,
```
tothtamas28, line 985: And this.

src/pyk/kcfg/explore.py
PetarMax: Yes, I wanted to have it like this first so that I can make sure that the semantics is passed correctly everywhere.

src/pyk/kcfg/semantics.py
```diff
@@ -0,0 +1,76 @@
...
+    def same_loop(c1: CTerm, c2: CTerm) -> bool:
+        ...
+
+    @abstractproperty
+    def cut_point_rules(self) -> Iterable[str]:
+        ...
+
+    @abstractproperty
+    def terminal_rules(self) -> Iterable[str]:
+        ...
```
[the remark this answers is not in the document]
PetarMax: Yes

src/tests/integration/kcfg/test_goto.py
PetarMax: Yes, definitely, this will be happening but I'm not there yet, have some other things to fix.

deps/k_release
PetarMax: I certainly didn't change that, I'm not sure what happened...

src/pyk/kcfg/explore.py
```diff
@@ -60,9 +61,12 @@ class KCFGExplore(ContextManager['KCFGExplore']):
     _rpc_closed: bool
     _trace_rewrites: bool
 
+    kcfg_semantics: KCFGSemantics
```
tothtamas28, line 64: It looks like `KCFGExplore` does not use `KCFGSemantics`, just stores it. Consider moving it out to where it is actually used (to prover classes I guess).

src/pyk/proof/reachability.py
```diff
@@ -385,9 +385,6 @@ def summary(self) -> Iterable[str]:
 
 class APRProver(Prover):
     proof: APRProof
-    _is_terminal: Callable[[CTerm], bool] | None
-    _extract_branches: Callable[[CTerm], Iterable[KInner]] | None
-    _abstract_node: Callable[[CTerm], CTerm] | None
```
tothtamas28, line 390: Related to the previous comment:

This is where the corresponding callables got removed, so it makes sense to replace them by an (optional) `KCFGSemantics`.

src/pyk/proof/reachability.py
```diff
@@ -652,26 +639,17 @@ def print(self) -> list[str]:
...
     _checked_nodes: list[int]
 
     def __init__(
         self,
         proof: APRBMCProof,
         kcfg_explore: KCFGExplore,
-        same_loop: Callable[[CTerm, CTerm], bool],
-        is_terminal: Callable[[CTerm], bool] | None = None,
-        extract_branches: Callable[[CTerm], Iterable[KInner]] | None = None,
-        abstract_node: Callable[[CTerm], CTerm] | None = None,
```
tothtamas28, line 665: Similarly here.

src/pyk/testing/_kompiler.py
```diff
@@ -128,10 +129,15 @@ def _update_symbol_table(symbol_table: SymbolTable) -> None:
 
 
 class KCFGExploreTest(KProveTest):
+    @abstractmethod
+    def semantics(self, definition: KDefinition) -> KCFGSemantics:
```
tothtamas28, line 133: Then all this mechanics becomes unnecessary as well.

src/pyk/kcfg/semantics.py
```diff
@@ -0,0 +1,42 @@
...
+    from ..kast.inner import KInner
+
+
+class KCFGSemantics(ABC):
+    @abstractmethod
+    def is_terminal(self, c: CTerm) -> bool:
+        ...
+
+    @abstractmethod
+    def extract_branches(self, c: CTerm) -> Iterable[KInner]:
```
tothtamas28, line 19:
```suggestion
    def extract_branches(self, c: CTerm) -> list[KInner]:
```

src/pyk/kcfg/semantics.py
```diff
@@ -0,0 +1,42 @@
...
+        return False
+
+    def extract_branches(self, c: CTerm) -> Iterable[KInner]:
+        return []
+
+    def abstract_node(self, c: CTerm) -> CTerm:
+        return c
+
+    def same_loop(self, c1: CTerm, c2: CTerm) -> bool:
+        return False
```
tothtamas28, line 42: Should this not be a reflexive relation?

PetarMax: I don't think so, it should hold only for loop heads (e.g., while loops in IMP), not arbitrary CTerms... @ehildenb?

src/pyk/kcfg/explore.py
PetarMax: Ah, that one's on me - it will use it, some of the Prover functionalities should move to KCFGExplore... will ping when that is done.

src/pyk/kcfg/semantics.py
ehildenb: Hmmmm, I guess it should be reflexive, yes. So maybe the default implementation should be `return c1 == c2`. Then teh user can loosen the relation as needed.

ehildenb:
@PetarMax can you mark the conversations that are done as resolved?

src/pyk/kcfg/explore.py
PetarMax: The functionalities are now as they should be, this requires another pass.

src/pyk/testing/_kompiler.py
PetarMax: This is needed now that the functionalities are in the correct places.

PetarMax:
> @PetarMax can you mark the conversations that are done as resolved?

Done.

src/pyk/kcfg/kcfg.py
```diff
@@ -201,6 +200,7 @@ def edges(self) -> tuple[KCFG.Edge, ...]:
     _splits: dict[int, Split]
     _ndbranches: dict[int, NDBranch]
     _stuck: set[int]
+    _terminal: set[int]
     _aliases: dict[str, int]
     _lock: RLock
```
ehildenb, line 206: Why are we making changes to `terminal` and `stuck` here? I guess `stuck` should stay in this class, but `terminal` should remain in `APRProof`?

src/pyk/kcfg/semantics.py
ehildenb: These both do look reflexive to me actually....

ehildenb: Ah wait, I guess it's not purely relexive. It's more like "if it's a loop head, then it's reflexive".

ehildenb: Maybe it makes sense to codify this a bit? That is, make a predicate `def is_loop_lead(cterm: CTerm) -> bool:`, and then the default implementation of this can be: `return is_loop_head(cterm1) and cterm1 == cterm2`.

Not sure we get much benefit from that, but maybe it does make sense to call this out as a specific call.

src/pyk/kcfg/kcfg.py
ehildenb: A main reason to _not_ have `terminal` in the `KCFG` datastructure is because it's semantics speciic, and there isn't a way to pass this predicate into the backend without hardcoding it. Eventually (already soon, in fact), the bottleneck will be communication between the pyk and the backend, and the backend will have to implement thigns like the `KCFG` datastructure directly, but will not have a way to mark things such as `is_terminal` because it's defined by a Python function. So I'd prefer not to use anything that is user-supplied heuristics for marking data in the `KCFG`, rather store it externally.

PetarMax: I thought `stuck` was as semantics-specific as `terminal`? The reason I put `terminal` in the KCFG was that the check for terminal nodes has moved from the prover to `KCFGExplore`, but the prover was supposed to keep track of them. Let me try to refactor

src/pyk/proof/reachability.py
```diff
@@ -483,25 +464,17 @@ def advance_pending_node(
         terminal_rules: Iterable[str] = (),
         implication_every_block: bool = True,
     ) -> None:
-        if implication_every_block or self._is_terminal is None or self._is_terminal(node.cterm):
+        if implication_every_block:
             if self._check_subsume(node):
+                # Shouldn't the node be marked somehow here?
+                # Otherwise, this subsumption call will be repeated?
```
RaoulSchaffranek, line 470: When the node can be subsumed in the target node, a cover will be created. Now, the node is no longer a leaf; more specifically, it is no longer pending. Hence, it won't be queued up for further advancement.

This is under the assumption that `advance_pending_node` will only be called for pending nodes. Maybe we should check this assumption explicitly at the beginning of the function?

src/pyk/proof/reachability.py
```diff
@@ -483,25 +464,17 @@ def advance_pending_node(
...
                 return
 
-        if self._check_terminal(node):
-            return
-
-        if self._check_abstract(node):
+        # Is this logic correct?
+        if self._check_terminal(node) and not self._check_subsume(node):
+            self.proof.kcfg.add_stuck(node.id)
```
RaoulSchaffranek, line 475: My intuition is that a terminal node that cannot be subsumed in the target state is not stuck but failing.

src/pyk/kcfg/explore.py
```diff
@@ -431,6 +437,15 @@ def section_edge(
             new_depth += section_depth
         return tuple(new_nodes)
 
+    def _check_abstract(self, node: KCFG.Node, kcfg: KCFG) -> bool:
+        new_cterm = self.kcfg_semantics.abstract_node(node.cterm)
+        if new_cterm == node.cterm:
+            return False
```
RaoulSchaffranek, line 443: My understanding is that you want to test if that abstracted node is strictly more general than the original node. I'm not sure that purely syntactical equality is sufficient for this check.

Alternatively, what if we make this property a requirement of the `abstract_node` function? For example, "return None if no abstraction step could be taken"?

src/pyk/kcfg/explore.py
```diff
@@ -446,6 +461,18 @@ def extend(
         if kcfg.is_stuck(node.id):
             raise ValueError(f'Cannot extend stuck node {self.id}: {node.id}')
 
+        if self._check_abstract(node, kcfg):
+            return
+
+        if len(kcfg.splits(target_id=node.id)) == 0:
```
tothtamas28, line 467:
```suggestion
        if not kcfg.splits(target_id=node.id):
```

src/pyk/kcfg/explore.py
```diff
@@ -446,6 +461,18 @@ def extend(
...
+            branches = list(self.kcfg_semantics.extract_branches(node.cterm))
+            if len(branches) > 0:
```
tothtamas28, line 469:
```suggestion
            if branches:
```

src/pyk/kcfg/explore.py
```diff
@@ -446,6 +461,18 @@ def extend(
         if kcfg.is_stuck(node.id):
             raise ValueError(f'Cannot extend stuck node {self.id}: {node.id}')
 
+        if self._check_abstract(node, kcfg):
+            return
+
+        if len(kcfg.splits(target_id=node.id)) == 0:
+            branches = list(self.kcfg_semantics.extract_branches(node.cterm))
```
tothtamas28, line 468:
```suggestion
            branches = self.kcfg_semantics.extract_branches(node.cterm)
```

src/tests/integration/kcfg/test_cell_map.py
```diff
@@ -46,14 +49,17 @@ class State(NamedTuple):
...
+    ('cell-map-no-branch', K_FILES / 'cell-map-spec.k', 'CELL-MAP-SPEC', 'cell-map-no-branch', 2, 1),
 )
 
 
 class TestCellMapProof(KCFGExploreTest):
     KOMPILE_MAIN_FILE = K_FILES / 'cell-map.k'
 
+    def semantics(self, definition: KDefinition) -> KCFGSemantics:
+        return DefaultSemantics()
+
```
tothtamas28, line 62:
```suggestion
```

src/tests/integration/kcfg/test_multiple_definitions.py
```diff
@@ -23,6 +26,9 @@
 class TestMultipleDefinitionsProof(KCFGExploreTest):
     KOMPILE_MAIN_FILE = K_FILES / 'multiple-definitions.k'
 
+    def semantics(self, definition: KDefinition) -> KCFGSemantics:
+        return DefaultSemantics()
+
```
tothtamas28, line 31:
```suggestion
```

src/tests/integration/kcfg/test_non_det.py
```diff
@@ -18,22 +19,27 @@
...
+    ('test-nondet', K_FILES / 'non-det-spec.k', 'NON-DET-SPEC', 'non-det', 8, 1),
 )
 
 
 class TestNonDetProof(KCFGExploreTest):
     KOMPILE_MAIN_FILE = K_FILES / 'non-det.k'
 
+    def semantics(self, definition: KDefinition) -> KCFGSemantics:
+        return DefaultSemantics()
+
```
tothtamas28, line 40:
```suggestion
```

src/tests/integration/kcfg/test_simple.py
```diff
@@ -40,6 +43,9 @@
 class TestSimpleProof(KCFGExploreTest):
     KOMPILE_MAIN_FILE = K_FILES / 'simple-proofs.k'
 
+    def semantics(self, definition: KDefinition) -> KCFGSemantics:
+        return DefaultSemantics()
+
```
tothtamas28, line 48:
```suggestion
```

src/pyk/kcfg/kcfg.py
ehildenb: `stuck` is not semantics specific in the sense that the backend doesn't need any extra input to determine if a node is stuck. `terminal` requires a heuristic supplied by the user in python.

So I guess the better distinction is "requires no Python code to determine"

src/pyk/kcfg/explore.py
ehildenb: The next few lines, which do `kcfg.create_cover` will enforce this. `create_cover` fails if it cannot find a `CSubst` which instantiates the more general node to the more specific one.

src/pyk/kcfg/kcfg.py
```diff
@@ -619,8 +618,13 @@ def edge_likes(self, *, source_id: NodeIdLike | None = None, target_id: NodeIdLi
...
 
     def add_stuck(self, node_id: NodeIdLike) -> None:
+        self._stuck.add(self._resolve(node_id))
+
+    def remove_stuck(self, node_id: NodeIdLike) -> None:
         node_id = self._resolve(node_id)
-        self._stuck.add(node_id)
+        if node_id not in self._stuck:
+            raise ValueError(f'Node is not stuck: {node_id}')
+        self._stuck.remove(node_id)
```
ehildenb, line 627:
```suggestion
        self._stuck.remove(node_id)

    def discard_stuck(self, node_id: NodeIdLike) -> None:
        node_id = self._resolve(node_id)
        self._stuck.discard(node_id)
```

ehildenb (approved):

src/pyk/kcfg/explore.py
RaoulSchaffranek: But what if the two terms are equivalent modulo alpha-renaming? In this case, we don't have pure syntactic equality, i.e., `new_cterm == node.cterm` is false, and `kcfg.create_cover` will succeed, where the substitution gives us the renamings. Strictly speaking, the new term is not strictly more abstract than the original term. This could lead to some infinite loop, where we keep abstracting terms forever without progressing, no?

PetarMax:
That is true, but this is a user-supplied semantics -specific function, so
if the user chooses to do something like that, so be it - I don't think we
should be protecting against.

PR #555, opened 2023-07-26
kcfg/explore: make kcfg_semantics optional

ehildenb:
Make the `kcfg_semantics` field optional in `KCFGExplore`.


ehildenb:
I'm not sure if there was a reason this wasn't optional before, because it has type `_ |None`, which seems to indicate that it should be optional.

PetarMax (approved):

PR #558, opened 2023-07-28
Add fail_fast option for APRProofs

nwatson22:
Adds an option to abort proof execution early, leaving remaining nodes pending, when a node is detected that meets the following condition:
1. It is terminal (based on the `is_terminal` predicate
2. It is failing (failed subsumption into target).


src/pyk/proof/reachability.py
```diff
@@ -609,14 +609,20 @@ def advance_pending_node(
         cut_point_rules: Iterable[str] = (),
         terminal_rules: Iterable[str] = (),
         implication_every_block: bool = True,
-    ) -> None:
+        fail_fast: bool = False,
+    ) -> bool:
```
tothtamas28, line 613: What does the return value stand for?

nwatson22: It was just a flag to tell the calling function to break out of the proof, but that was messy, so I figured out a way where that is unnecessary now.

src/pyk/proof/reachability.py
```diff
@@ -655,6 +656,15 @@ def advance_proof(
                 terminal_rules=terminal_rules,
                 implication_every_block=implication_every_block,
             )
+            if (
+                fail_fast
+                and self.kcfg_explore.kcfg_semantics.is_terminal(curr_node.cterm)
+                and (not self.proof.kcfg.is_covered(curr_node.id))
```
tothtamas28, line 662:
```suggestion
                and not self.proof.kcfg.is_covered(curr_node.id)
```

tothtamas28 (approved):

src/pyk/proof/reachability.py
nwatson22: Done.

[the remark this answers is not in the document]
nwatson22: Done.

src/pyk/proof/reachability.py
```diff
@@ -655,6 +656,15 @@ def advance_proof(
                 terminal_rules=terminal_rules,
                 implication_every_block=implication_every_block,
             )
+            if (
+                fail_fast
+                and not self.proof.kcfg.is_covered(curr_node.id)
+                and self.kcfg_explore.kcfg_semantics.is_terminal(curr_node.cterm)
+            ):
```
ehildenb, line 663: @nwatson22 can't we just say `if (fail_fast and self.proof.status == ProofStatus.FAILING):` and leave it up to the proof type to determine whether it's failing or not?

PR #559, opened 2023-07-31
Restructuring Provers

PetarMax:
This PR introduces:
- The `ExplorationProof` and `ExplorationProver` classes, which will power the equivalence checker, and which are, in essence, `APR`s without target nodes and associated functionalities
- Reformulation of how some auxiliary functions are computed  to minimise recomputation
- Refactoring of `APRProof` and `APRProver`to be subclasses of their `Exploration` counterparts
- Corrections to the BMC proof and prover to also accept terminal nodes


src/pyk/proof/exploration.py
```diff
@@ -0,0 +1,328 @@
...
+        if node.id not in self._checked_terminals:
+            _LOGGER.info(f'Checking terminal: {shorten_hashes(node.id)}')
+            if self.kcfg_explore.kcfg_semantics.is_terminal(node.cterm):
+                _LOGGER.info(f'Terminal node: {shorten_hashes(node.id)}.')
+                self.proof.add_terminal(node.id)
+                self._checked_terminals[node.id] = True
+            else:
+                self._checked_terminals[node.id] = False
+
+        return self._checked_terminals[node.id]
```
ehildenb, line 266: Why are we tracking if we've already checked this node for terminal? It's a pretty cheap check isn't it?

ehildenb: Hmmm, I see that we have `_update_terminals` now which goes through every node in `pending` to check it.

src/pyk/proof/exploration.py
```diff
@@ -0,0 +1,328 @@
...
+
+    def _update_terminals(self) -> None:
+        pending = self.proof.pending
+
+        for node in pending:
+            self._check_terminal(node)
```
ehildenb, line 272:
```suggestion
    def _update_terminals(self) -> None:
        for node in self.proof.pending:
            self._check_terminal(node)
```

src/pyk/proof/exploration.py
```diff
@@ -0,0 +1,328 @@
...
+                execute_depth=execute_depth,
+                cut_point_rules=cut_point_rules,
+                terminal_rules=terminal_rules,
+                implication_every_block=implication_every_block,
+            )
+
+            self._update_terminals()
+
+        self.proof.write_proof_data()
+        return self.proof.kcfg
```
ehildenb, line 328: Do we really need to return the `kcfg`? I guess that's legacy from how the codebase used to work, but it seems unneccessary, since the `proof` object that the usre passed in is benig updated in place anyway.

src/pyk/proof/proof.py
```diff
@@ -23,8 +23,9 @@
 
 
 class ProofStatus(Enum):
-    PASSED = 'passed'
+    COMPLETED = 'completed'
     FAILED = 'failed'
+    PASSED = 'passed'
     PENDING = 'pending'
```
ehildenb, line 29: If a proof is `COMPLETED` but not `PASSED` or `FAILED`, what could it be?

src/pyk/proof/exploration.py
```diff
@@ -0,0 +1,328 @@
...
+
+    def is_terminal(self, node_id: NodeIdLike) -> bool:
+        return self.kcfg._resolve(node_id) in self._terminal_node_ids
+
+    def is_pending(self, node_id: NodeIdLike) -> bool:
+        return self.kcfg.is_leaf(node_id) and not self.is_terminal(node_id) and not self.kcfg.is_stuck(node_id)
+
+    @property
+    def pending(self) -> list[KCFG.Node]:
+        return [nd for nd in self.kcfg.leaves if not self.is_terminal(nd.id) and not self.kcfg.is_stuck(nd.id)]
```
ehildenb, line 74:
```suggestion
    @property
    def pending(self) -> list[KCFG.Node]:
        return [nd for nd in self.kcfg.leaves if self.is_pending(nd.id)]
```

src/pyk/proof/reachability.py
```diff
@@ -91,56 +80,25 @@ def __init__(
...
-        self._terminal_nodes.append(self.kcfg._resolve(node_id))
-
-    def remove_terminal(self, node_id: NodeIdLike) -> None:
-        node_id = self.kcfg._resolve(node_id)
-        if node_id not in self._terminal_nodes:
-            raise ValueError(f'Node is not terminal: {node_id}')
-        self._terminal_nodes.remove(node_id)
+    @property
+    def pending(self) -> list[KCFG.Node]:
+        return [nd for nd in self.kcfg.leaves if self.is_pending(nd.id)]
```
ehildenb, line 97: With the suggested change to `ExplorationProof.pending`, this is now redundant and can be removed.

src/pyk/proof/reachability.py
```diff
@@ -91,56 +80,25 @@ def __init__(
...
 
-    def shortest_path_to(self, node_id: NodeIdLike) -> tuple[KCFG.Successor, ...]:
-        spb = self.kcfg.shortest_path_between(self.init, node_id)
-        assert spb is not None
-        return spb
+    @property
+    def failing(self) -> list[KCFG.Node]:
+        return [nd for nd in self.kcfg.leaves if self.is_failing(nd.id)]
```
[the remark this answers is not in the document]
ehildenb: Ah, nevermind, this one must stay.

src/pyk/proof/reachability.py
```diff
@@ -333,236 +273,116 @@ def write_proof_data(self) -> None:
...
-        logs = {int(k): [l.to_dict() for l in ls] for k, ls in self.logs.items()}
-        dct['logs'] = logs
-        dct['circularity'] = self.circularity
-        return dct
+    def __init__(
+        self,
+        proof: APRProof,
+        kcfg_explore: KCFGExplore,
+    ) -> None:
+        Prover.__init__(self, kcfg_explore)
```
ehildenb, line 373: Sholud we be calling `super().__init__(self, proof, kcfg_explore)` instead? Then, we wouldn't need to re-init all the things taht `ExplorationProver` is already initializing, such as `_checked_terminals` and `proof`.

src/pyk/proof/exploration.py
PetarMax: `self.is_pending(nd.id)` has an additional check whether or not the node is a leaf, the current solution is more efficient. This does mean that if the notion of `pending` would change, then this function would have to be updated as well, but this is not likely to happen.

src/pyk/proof/exploration.py
PetarMax: This does not work as intended, in fact, since it would lead to not checking all of the pending nodes in cases a node other than the last node in the `pending` list is terminal. 

For example, this is what happens when one iterates over a list and in each iteration removes the head of the list:
```
0
[0, 1, 2, 3, 4, 5, 6]
2
[1, 2, 3, 4, 5, 6]
4
[2, 3, 4, 5, 6]
6
[3, 4, 5, 6]
```
In particular, 1, 3, and 5 are never examined.

src/pyk/proof/reachability.py
PetarMax: Yes, I found this unsatisfactory as well. The problem I had was that the call to `super` would do this:
```
        self.kcfg_explore.add_dependencies_module(
            self.main_module_name,
            self.dependencies_module_name,
            [],
            priority=1,
        )
```
and then our constructor would do this on top:
```
        self.kcfg_explore.add_dependencies_module(
            self.main_module_name,
            self.dependencies_module_name,
            dependencies_as_claims,
            priority=1,
        )
```
and I wasn't able to figure out for sure if these were in conflict or not.

src/pyk/proof/exploration.py
ehildenb: Ah I see.

ehildenb: What if we just iterated over _all_ nodes then. I think even if a node is not a pending/leaf node, it's useful to mark it as `terminal`, and we can still track the list of ones we've already checked.

src/pyk/proof/exploration.py
ehildenb: No, we should definitely use the predicate, that's how these are designed to be.

src/pyk/proof/reachability.py
ehildenb: No, the point is to use the `is_...` predicates directly, so that we don't risk them disagreeing.

These do not need to be efficient, they ahve never showed up as cost centers in profile. Better to have them be simple and correct:

src/pyk/proof/reachability.py
ehildenb: Can w efactor out the `add_dependencies_as_claims` functionality and do it only here?

src/pyk/proof/exploration.py
PetarMax: Very well, presumably the duplicated computation won't affect the speed ever.

src/pyk/proof/reachability.py
PetarMax: Very well.

src/pyk/proof/exploration.py
PetarMax: This would be ok for the `ExplorationProver` in principle, but `APRProver` will expand `_check_terminal` with a check that the newly found terminal node is subsumed into the target node, and I don't think we'd want to do that for terminal nodes that were not pending (say, for the target node itself).

src/pyk/proof/reachability.py
PetarMax: I've tried to do that, please take a look @ehildenb

src/pyk/proof/exploration.py
ehildenb: I think it's useful to mark the `target` node as `terminal` or not. `terminal` is just a descriptor for nodes, so I don't really see why it can't be applied to all nodes.

In particular, we'll want to check if the `target` node has `terminal` being true about it, and if so, then we'll only check subsumption into it for nodes that are also `terminal`.

So not sure why it doesn't make sense to not just check `terminal` for all nodes.

PetarMax: In the context of the `APRProver`, this is the `_update_terminals` function:
```
    def _update_terminals(self) -> None:
        pending = self.proof.pending
        for node in pending:
            self._check_terminal(node)
```
and this is the `_check_terminal` function:
```
    def _check_terminal(self, node: KCFG.Node) -> bool:
        result = super()._check_terminal(node)

        if result:
            self._check_subsume(node)

        return result
```

If we start checking whether or not nodes other than `pending` are `terminal`, then we'll be creating covers for those nodes. In particular, we will creata a cover for `target`. I assume that this is not something that we want. I also don't see a use case for checking nodes other than `pending` and `target` for being terminal, and until there is such a use case, I don't see a reason why we should be doing that, at least not in this particular place.

If we examine how the prover works in more detail, by construction we hae that:

a) stuck nodes will not be terminal, so they need not be checked
b) refuted nodes should not exist in the first place, so they need not be checked
c) nodes marked as terminal via `_terminal_node_ids` are terminal by construction, so they need not be checked
d) all nodes created during the execution will be checked

which means (given d) specifically) that there is no need to check all nodes on every iteration of the prover

and also that only:

1) `target`; and 
2) nodes that were non-pending, non-stuck, non-refuted, non-target, and not-marked-as-terminal at the creation time of the prover 

might not end up being checked. As for 1), `target` we can check straightforwardly on initialisation of the prover, and we will, but not for this PR - this will happen in the subsequent PR on `implication-every-block`.

As for 2), if there is a such a node in the initial KCFG, then that initial KCFG was malformed. In particular, by definition, such a node is not a `leaf`. Then, if that node:

1) has any `edge`s, `splits`, or `ndbranches` coming out of it, we have an error in which the execution somehow continued from the node, but should not have as the node is `terminal`;
2) otherwise, if the node has a cover into `target`, we have an error in which an actual `terminal` node was not passed in via `_terminal_node_ids` on prover initialisation - that one we can fix by adding the node there;
3) otherwise, the node has a cover into some node other than target, and I don't know what that means but I assume that that's also a malformation.

So, in summary, we could potentially do a `terminal` check for all nodes when initialising the APRProver, only once and as a sanity check, throwing an error if there is some malformation going on.

src/pyk/proof/reachability.py
```diff
@@ -59,24 +55,17 @@ def __init__(
...
         circularity: bool = False,
         admitted: bool = False,
     ):
-        super().__init__(id, proof_dir=proof_dir, subproof_ids=subproof_ids, admitted=admitted)
-        self.kcfg = kcfg
-        self.init = init
-        self.target = target
-        self.logs = logs
+        super().__init__(id, kcfg, init, logs, proof_dir=proof_dir, terminal_node_ids=terminal_node_ids)
+        Proof.__init__(self, id, proof_dir=proof_dir, subproof_ids=subproof_ids, admitted=admitted)
```
tothtamas28, line 64: Is `Proof` initialization not taken care of by `ExplorationProof`?

tothtamas28: > `ExplorationProofs` aren't really proofs, they are just KCFG explorations with a bit on top.

Would this be a solution then?

```python
class Exploration:
    kcfg: KCFG
    ...

class APRProof(Proof):
    exploration: Exploration
```

src/pyk/proof/exploration.py
```diff
@@ -44,8 +44,14 @@ def __init__(
         logs: dict[int, tuple[LogEntry, ...]],
         proof_dir: Path | None = None,
         terminal_node_ids: Iterable[NodeIdLike] | None = None,
+        subproof_ids: Iterable[str] = (),
+        admitted: bool = False,
     ):
-        super().__init__(id, proof_dir=proof_dir, subproof_ids=[], admitted=False)
+        if type(self) == ExplorationProof:
+            assert subproof_ids == ()
+            assert not admitted
```
tothtamas28, line 52: Can an `ExplorationProof` be directly instantiated, or is it abstract?

Also, if it is part of the public interface, then `assert`-s should be replaced by `ValueError`-s.

src/pyk/proof/reachability.py
PetarMax: > > `ExplorationProofs` aren't really proofs, they are just KCFG explorations with a bit on top.
> 
> Would this be a solution then?
> 
> ```python
> class Exploration:
>     kcfg: KCFG
>     ...
> 
> 
> class APRProof(Proof):
>     exploration: Exploration
>     ...
> ```

I don't think this would work, I need `ExplorationProofs` for the equivalence checker.

src/pyk/proof/exploration.py
```diff
@@ -0,0 +1,347 @@
...
+        self,
+        max_iterations: int | None = None,
+        execute_depth: int | None = None,
+        cut_point_rules: Iterable[str] = (),
+        terminal_rules: Iterable[str] = (),
+        implication_every_block: bool = True,
+        fail_fast: bool = False,
+    ) -> None:
+        if self.proof.admitted:
+            raise ValueError('Attempting to advance an admitted proof')
```
PetarMax, line 319: I've added this check to ensure admitted proofs cannot be advanced

src/pyk/proof/exploration.py
```diff
@@ -0,0 +1,347 @@
...
+        proof_dir: Path | None = None,
+        terminal_node_ids: Iterable[NodeIdLike] | None = None,
+        subproof_ids: Iterable[str] = (),
+        admitted: bool = False,
+    ):
+        if type(self) == ExplorationProof:
+            if not subproof_ids:
+                raise ValueError('Subproofs provided for an ExplorationProof')
+            if not admitted:
+                raise ValueError('Admitted ExplorationProof')
```
tothtamas28, line 54: Should it not be

```suggestion
        if type(self) == ExplorationProof:
            if subproof_ids:
                raise ValueError('Subproofs provided for an ExplorationProof')
            if admitted:
                raise ValueError('Admitted ExplorationProof')
```

instead?

tothtamas28: Ways for eliminating this special casing, as per our Slack discussion:
* Pull `admitted` and `subproofs` from `Proof` into subclasses that use them
* Make `ExplorationProof` abstract so that `APRProof` can implement it. Make basic implementation for `ExplorationProver` to instantiate.
* `APRProver` is not an `ExplorationProver`, just uses one. `APRPRoof` is no longer an `ExplorationProof`.

PetarMax: Yes yes, that was bizarre.

PetarMax: > Should it not be
> 
> instead?

Yes yes, that was bizarre.

src/pyk/proof/reachability.py
```diff
@@ -333,11 +277,240 @@ def write_proof_data(self) -> None:
...
+            f'    status: {self.status}',
+            f'    admitted: {self.admitted}',
+            f'    nodes: {self.nodes}',
+            f'    pending: {self.pending}',
+            f'    failing: {self.failing}',
+            f'    stuck: {self.stuck}',
+            f'    terminal: {self.terminal}',
+            f'    refuted: {self.refuted}',
+            f'Subproofs: {self.subproofs}',
+        ]
```
ehildenb, line 306: The problem with doing this code re-organization/movement at the same time as introducing a new feature is that now I have to review this by opening two windows side-by-side (where the code came from and where it was moved to) to see if there were any differences in the code:

image

Can you please re-organize the classes to be in the same order they were before the PR? You can open a separate PR afterwards with the code movement. Remember, the goal is to make easy to review PRs. Please make sure you are doing self-review of your PRs in github. Actually doing self-review, not just scanning the code change. Actively check for problems as if you are reviewing someone else changes, and make sure that the changes you have are in scope and easy to review. Please review our GitHub Guidelines for Collaborative Coding again, and try to follow all the steps there, everything there is from 5 years of trying to manage software development at an organization with 40+ active repositories with interconnected dependencies between them: https://runtimeverification.com/blog/github-guidelines-for-collaborative-coding

src/pyk/proof/exploration.py
```diff
@@ -0,0 +1,359 @@
...
+        return self.kcfg.is_leaf(node_id) and not self.is_terminal(node_id) and not self.kcfg.is_stuck(node_id)
+
+    @property
+    def pending(self) -> list[KCFG.Node]:
+        return [nd for nd in self.kcfg.leaves if self.is_pending(nd.id)]
+
+    @property
+    def failing(self) -> list[KCFG.Node]:
+        return []
+
```
ehildenb, line 88: Why is this property needed for an `ExplorationProof`? It seems to me that it's probably not making sense here.

src/pyk/proof/exploration.py
```diff
@@ -0,0 +1,359 @@
...
+        raise ValueError(f'Could not load ExplorationProof from file {id}: {proof_path}')
+
+    @property
+    def status(self) -> ProofStatus:
+        if len(self.kcfg.stuck) > 0:
+            return ProofStatus.FAILED
+        elif len(self.pending) > 0:
+            return ProofStatus.PENDING
+        else:
+            return ProofStatus.PASSED
```
ehildenb, line 125:
```suggestion
    @property
    def status(self) -> ProofStatus:
        if len(self.kcfg.stuck) > 0 or self.subproofs_status == ProofStatus.FAILED:
            return ProofStatus.FAILED
        elif len(self.pending) > 0 or self.subproofs_status == ProofStatus.PENDING:
            return ProofStatus.PENDING
        else:
            return ProofStatus.PASSED
```

Does it really make sense to have a `FAILED` result here, even for `stuck` nodes? Maybe this should only return `PENDING` or `PASSED`.

But this makes me think that this class structure just doesn't make sense. This may be better suited as just a method added to `KCFGExplore` which takes in a `KCFGSemantics` as well.

src/pyk/proof/exploration.py
```diff
@@ -0,0 +1,347 @@
...
+        fail_fast: bool = False,
+    ) -> None:
+        if self.proof.admitted:
+            raise ValueError('Attempting to advance an admitted proof')
+
+        iterations = 0
+
+        self._update_terminals()
+
+        while not (fail_fast and self.proof.status == ProofStatus.FAILED) and self.proof.pending:
```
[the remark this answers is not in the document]
ehildenb: Should be taken care of here: https://github.com/runtimeverification/pyk/pull/572

src/pyk/proof/exploration.py
```diff
@@ -0,0 +1,359 @@
...
+    from ..kcfg.kcfg import NodeIdLike
+
+    T = TypeVar('T', bound='Proof')
+
+_LOGGER: Final = logging.getLogger(__name__)
+
+
+class ExplorationProof(Proof):
+    kcfg: KCFG
+    init: int
```
ehildenb, line 34:
```suggestion
    init: NodeIdLike
```

Any reason to use `int` here?

src/pyk/proof/exploration.py
```diff
@@ -0,0 +1,359 @@
...
+        dct['type'] = 'ExplorationProof'
+        dct['cfg'] = self.kcfg.to_dict()
+        dct['init'] = self.init
+        dct['terminal_node_ids'] = self._terminal_node_ids
+        logs = {k: [l.to_dict() for l in ls] for k, ls in self.logs.items()}
+        dct['logs'] = logs
+        return dct
+
+    @property
+    def summary(self) -> CompositeSummary:
```
ehildenb, line 173:
```suggestion
    def summary(self) -> ProofSummary:
```

I guess this should match the superclass type signature?

src/pyk/proof/exploration.py
ehildenb: Why would `APRProver` check for subsumption on non-pending nodes? That wouldn't be very smart.

It still makes sense to check every node for `terminal`, because it's fundamentally an attribute of the node itself, not of the algorithm using it. So I think that we should just make it simple and check every node for `terminal` on every invocation of `advance_pending_node`. I also think `advance_pending_node` should also check fi the node is actually pending or not (which means that it won't advance a `terminal` node, for example).

src/pyk/proof/exploration.py
PetarMax: It's not needed - it brought it in briefly while I was trying to understand how to work with `fail_fast`, but when I moved to checking for `ProofStatus.FAILED` it became redundant and I neglected to remove it - fixed now.

src/pyk/proof/reachability.py
ehildenb: Not sure why this wouldn't work for the situation though. This could be a `KCFGExploration` class instead, which contains additional semantics-specifci information. Think of it as a thin wrapper around `KCFG`, which also incorporates the heuristics defined in `KCFGSemantics` into the exploration process.

Then that class `KCFGExploration` could avoid having to define things like `status` and `summary`, which don't make as much sense for it.

src/pyk/proof/exploration.py
PetarMax: If I remember correctly, in one discussion you said something along the lines of "`stuck` nodes should not be happening in a correctly-defined semantics". I am following that intuition to mark a proof with stuck nodes as failed.

`ExplorationProof`s are currently, by construction, not allowed to have `subproofs`. What would a subproof mean in an `ExplorationProof`?

ehildenb:
I'm somehow in agreement with @tothtamas28 on this one, that it doesn't make sense as a `Proof` subclass. Instead, I think it makes sense as a wrapper `KCFGExploration`, which basically stores a `KCFG` but is enriched with data from the `KCFGSemantics` that it can compute (such as `terminal` and other heuristics there about how to do explorations). `KCFGExplore` can be modified to work over a `KCFGExploration` instead of a `KCFG`, and so can `APRProof` and the `EquivalenceProof`.

The reason for this design is because we want to achieve:

- All data that is stored in the `KCFG` is data the backends could feasibly compute all on their own. This is important especially as the Maude backend comes online, and as the booster gets faster, because we'll want to ask the backends to compute an entire `KCFG` out to some iteration bound, then we post-process it, as the overhead of communication begins to dominate the overall time (ping @dwightguth ). That is why we can't store python-heuristic stuff from `KCFGSemantics` directly in this `KCFG` datastructure.
- The `KCFGExploration` datastructure can take advantage of these heuristics to enrich it with semantics-specific supplied Python heuristics for guiding/cutting the exploration down.

But `KCFGExploration` doesn't have much to do with proof state, because there isn't much notion of what it means for it to be passed/failed.

src/pyk/proof/exploration.py
PetarMax: The identifiers of KCFG nodes are `int`s. By having an `int` here, we minimise the calls to `_resolve` to only the class constructor. It also prevents malformed proofs from being constructed, for example, proofs that provide an `init` that is not in the given `kcfg`.

From what I've noticed, `NodeIdLike` is only used for input arguments of functions and is almost always immediately resolved. The only place where a function returns `NodeIdLike` is the `kcfg.prune` function, which returns `list[NodeIdLike]`, but it actually returns `list[int]`, and should probably be typed with this return type unless there is really a need for `NodeIdLike` to be there.

src/pyk/proof/exploration.py
PetarMax: I think so, if no subproofs are allowed.

PetarMax:
I think I'm in agreement with the last comment as well. How should I proceed?

Perhaps more precisely, shall I proceed to do this in this PR or discard this PR and start another? I'm ok either way, it should be quick.

PetarMax:
But if we do this, what kind of proofs will I have for the two programs in the equivalence prover? Can't be APR. I could just try to manipulate them using this new `KCFGExploration` structure...

tothtamas28:
> enriched with data from the `KCFGSemantics` that it can compute (such as `terminal` and **other heuristics there about how to do explorations**)

I discussed a very similar issue with @anvacaru: https://github.com/runtimeverification/pyk/pull/575. The design outlined by @ehildenb (https://github.com/runtimeverification/pyk/pull/559#issuecomment-1662951777) would be a solution there as well I think.

(Maybe we should rename `KCFGSemantics` to `ExplorationStrategy`.)

ehildenb:
Yes @PetarMax , I think just manipulate them using the new `KCFGExploration` datastructure.

PR #560, opened 2023-07-31
 Add class `DefinitionInfo`

tothtamas28:
Enables extracting metadata from a kompiled definition directory.


src/pyk/ktool/kompile.py
```diff
@@ -347,3 +347,31 @@ def args(self) -> list[str]:
 
 
 COMMON_ARGS: Final = frozenset(field.name for field in dataclasses.fields(KompileArgs))
+
+
+@final
+@dataclass(frozen=True)
+class DefinitionInfo:
```
ehildenb, line 354: Could we have `KPrint` use a `DefinitionInfo` (or inherit from it)? Right now I think it reads a bunch of the same information.

ehildenb (approved):

src/pyk/ktool/kompile.py
tothtamas28: Yes, I changed the implementation there.

PR #569, opened 2023-08-02
Inject into `KItem` instead of `K`

### LINKED ISSUE #565: `==K` getting encoded as an uninterpreted function despite having an smt-hook
# Steps to reproduce:
1. Create files:
`pyproject.toml`
```
[tool.poetry]
name = "test5"
version = "0.1.0"
description = ""
authors = ["Your Name <you@example.com>"]
readme = "README.md"

[tool.poetry.dependencies]
python = "^3.10"
pyk = {git = "https://github.com/runtimeverification/pyk.git"}
pytest = "^7.4.0"


[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

`test.k`
```
module TEST-SYNTAX
  imports INT-SYNTAX

  syntax KItem ::= foo(Int)
                 | success()
                 | fail()

endmodule

module TEST
  imports TEST-SYNTAX
  imports INT
  imports K-EQUAL
  imports BOOL-COMMON

  configuration <k> $PGM:KItem </k>

  rule foo(X:Int) => success() requires X <Int 2
  rule foo(_:Int) => fail() [owise]

endmodule
```

`tests-spec.k`
```
module TEST-SPEC
  imports TEST
  
  claim [claim1]: <k> foo(X) => success() ... </k>
    requires X ==K 0 orBool X ==K 1

endmodule
```

`test.py`
```
from pyk.ktool.kprove import KProve
from pyk.testing._kompiler import Kompiler
from pyk.proof.reachability import APRProof, APRProver
from pyk.kcfg.explore import KCFGExplore
from pyk.kcfg.semantics import KCFGSemantics
from pyk.kcfg.show import KCFGShow
from pyk.cterm import CTerm
from pyk.kast.inner import KInner
from pyk.proof.show import APRProofNodePrinter
from pathlib import Path

class TestSemantics(KCFGSemantics):
    def is_terminal(self, c: CTerm) -> bool:
        return False

    def extract_branches(self, c: CTerm) -> list[KInner]:
        return []

    def abstract_node(self, c: CTerm) -> CTerm:
        return c

    def same_loop(self, c1: CTerm, c2: CTerm) -> bool:
        return False


base_dir = Path('use_dir')
if not base_dir.is_dir():
    base_dir.mkdir()
use_directory = base_dir / 'kprove'
if not use_directory.is_dir():
    use_directory.mkdir()

proof_dir = Path('proof_dir')

definition_dir = Path('test-kompiled')

kprove = KProve(
    definition_dir,
    use_directory=use_directory,
)

claim = kprove.get_claims(
    Path('test-spec.k'),
    spec_module_name='TEST-SPEC',
    claim_labels=['TEST-SPEC.claim1'],
)[0]

proof = APRProof.from_claim(kprove.definition, claim, proof_dir=proof_dir, logs={})

kcfg_explore = KCFGExplore(
    kprove,
    kcfg_semantics=TestSemantics(),
)

prover = APRProver(proof, kcfg_explore=kcfg_explore)

prover.advance_proof()

kcfg_show = KCFGShow(
    kcfg_explore.kprint, node_printer=APRProofNodePrinter(proof, kcfg_explore.kprint, full_printer=True)
)

cfg_lines = kcfg_show.show(proof.kcfg)
print('\n'.join(cfg_lines))
```

2. `poetry update`
3. `kompile test.k --backend haskell --emit-json`
4. `KORE_RPC_OPTS="--solver-transcript transcript.z3" poetry run python3 test.py`
# Expected result:
Initial state is rewritten to `success()` and transcript shows that it is checking something like:
```
(assert (and (= true (or (= <0> 0) (= <0> 1) ) ) (not (= true (< <0> 2 ) ) ) ) )
```
which should not be satisfiable, leading the branch where X < 2 to be pruned.
# Actual result:
Proof branches on condition `X <Int 2` and transcript shows:
```
(declare-fun <0> () Bool )
(declare-fun <1> () Bool )

(assert (and (= true (or <0> <1> ) ) (not (= true (< <2> 2 ) ) ) ) )
```

# Notes

I'm not sure if that this is a Haskell backend thing. It could be the way pyk is encoding the kore-rpc calls.

It only seems to happen with the kore-rpc route. Proving the claim with `kprove` works fine.

nwatson22:
closes #565 
We weren't doing the kast to kore conversion in the same way as `kore-exec`, resulting in a possibly invalid kore term (per discussion on Slack involving @goodlyrottenapple @tothtamas28 and @virgil-serbanuta).
This changes the conversion so that the injection is not done directly into the `K` sort, but into a `KItem` which is embedded into a K sequence.
I don't know much about kore so I'm not sure if this is correct, but it did produce the correct expected output on #565.


src/pyk/kore/kompiled.py
```diff
@@ -75,10 +75,10 @@ def is_subsort(self, sort1: Sort, sort2: Sort) -> bool:
         if sort1 == sort2:
             return True
 
-        if sort2 == SortApp('SortK'):
+        if sort2 == SortApp('SortKItem'):
             return True
 
-        if sort1 == SortApp('SortK'):
+        if sort1 == SortApp('SortKItem'):
             return False
```
tothtamas28, line 82: `SortK` is the top sort, this shouldn't be changed.

src/pyk/kore/kompiled.py
```diff
@@ -122,4 +122,7 @@ def _inject(self, pattern: Pattern, sort: Sort) -> Pattern:
         if self.is_subsort(actual_sort, sort):
             return App('inj', (actual_sort, sort), (pattern,))
 
+        if sort == SortApp('SortK'):
```
tothtamas28, line 125: Please add a comment for this special case.

Also this branch needs to precede the subsort checking.

tothtamas28:
I wonder though, is there a reason why `SortK` doesn't admit injections? Maybe the ultimate solution would be to just generated the corresponding axioms on `inj` based on

```
syntax K ::= KItem
```

@radumereuta @virgil-serbanuta

virgil-serbanuta:
> I wonder though, is there a reason why `SortK` doesn't admit sort injections? 

I think that `K` is supposed to be a list of `KItem`, based on two constructors, `kseq` and `dotk`. This is used as such most obviously in the `<k>` cell.

If you would allow injections, that would make things ambiguous, or would make unification hard, or both.

The main question is in what way an injection is related to a single element list.

If an injection should be treated as a single element list, that would mean that `inj{KItem, K}(stuff)` should unify with `kseq(stuff, dotk)`, which would mean that `kseq` is no longer a constructor, which would make unification harder than it should be.

If an injection is not the same as a single element list, then a rule containing `<k>stuff</k>`, i.e., containing `<k>stuff ~> .K </k>`, would not unify with a configuration containing `<k>inj(stuff)</k>`, which would probably be confusing.

src/pyk/kore/kompiled.py
```diff
@@ -122,4 +122,7 @@ def _inject(self, pattern: Pattern, sort: Sort) -> Pattern:
...
+            return App('kseq', (), (App('inj', (actual_sort, SortApp('SortKItem')), (pattern,)), App('dotk')))
```
virgil-serbanuta, line 126: Would it be reasonable to use `KSequence` and `_ksequence_to_kore` here?

virgil-serbanuta:
> @virgil-serbanuta perhaps a more sensible change would be to have `==K` defined between two `KItems`? Tho this would be a very pervasive change too, so not sure what the consequences would be

I think that `==K` is currently used for actual sequences of `KItem`, so it should stay that way. Also, `==Sort`, whenever it's defined, means equality on `Sort`, so changing that for `K` would be confusing.

src/pyk/kore/kompiled.py
tothtamas28: At least I think it is the top sort. @radumereuta, can you confirm?

ehildenb: I somehow agree this sholudn't be changed either. Fro example, if `sort1` is `SortK`, then this code will return that `SortKItem` is a subsort of `SortK`.

ehildenb:
Let's add unit tests of this (or integration tests) of `kast_to_kore`.

But in general, `KItem` is not the top sort, `K` is. So we shouldn't treat `KItem` as such. we should be special casing the handful of places where we need to do this injection.

What does the original term we have look like? And what the backend expects?

src/pyk/kore/kompiled.py
ehildenb: This makes sense. Basically in `kast_to_kore`, when we see something of sort `KItem` being used in a place where sort `K` is expected, emit `_ksequence_to_kore(KSequence[KITEM])`.

Maybe then we don't need to touch anything here.

goodlyrottenapple:
@ehildenb from the linked issue:

pyk does not encode the `==K` predicates correctly into kore. Currently, for `LHS ==K RHS`, we emit:

```
Lbl'UndsEqlsEqls'K'Unds'{}(
    inj{SortInt{}, SortK{}}(LHS),
    inj{SortInt{}, SortK{}}(RHS)
)
```

however, the correct encoding should be

```
Lbl'UndsEqlsEqls'K'Unds'{}(
  kseq{}(
    inj{SortInt{}, SortKItem{}}(ConfigVarX:SortInt{}),
    dotk{}()
  ),
  kseq{}(
    inj{SortInt{}, SortKItem{}}(\dv{SortInt{}}("0")),
    dotk{}()
  )
)
```

src/pyk/kore/kompiled.py
nwatson22: This is reverted. It now adds the KSequences while it is still in Kast form.

src/pyk/kore/kompiled.py
nwatson22: This case has been removed and now the KSeqences are added earlier, when the term is still in Kast format. Added a comment for this conversion step.

src/pyk/kore/kompiled.py
nwatson22: Moved the whole conversion to before the `_kast_to_kore` conversion, so we are now using `KSequence`.

nwatson22:
> Ok, then I think we should probably special case `_==K_` in `kast_to_kore`, to insert the `KSequence` underneath it. It can be added as a pass here:
> 
> https://github.com/runtimeverification/pyk/blob/1679ce41cd9b8c281a2ddda5295e3d37a8a54f48/src/pyk/konvert.py#L90
> 
> Basically, before that line, traverse the configuration and do `add_ksequence_under_kequal` to the kast term.
> 
> But, more importantly, add a test of this directly first, for example here:
> 
> https://github.com/runtimeverification/pyk/blob/1679ce41cd9b8c281a2ddda5295e3d37a8a54f48/src/tests/integration/konvert/test_simple_proofs.py#L473

Added this as a test and changed to be a special case of `_==K_`.

ehildenb (approved):

ehildenb:
This looks good to me. Would be good to confirm that it solves the original issue as well before adding automerge.

src/pyk/kast/outer.py
```diff
@@ -1137,6 +1137,16 @@ def greatest_common_subsort(self, sort1: KSort, sort2: KSort) -> KSort | None:
             return sort2
         return None
 
+    # Sorts like Int cannot be injected directly into sort K so they are embedded in a KSequence.
+    def add_ksequence_under_kequal(self, kast: KInner) -> KInner:
+        def _add_ksequence_under_kequal(kast: KInner) -> KInner:
+            if type(kast) is KApply and kast.label.name == '_==K_':
+                return KApply('_==K_', [KSequence(arg) for arg in kast.args])
```
virgil-serbanuta, line 1144: What happens if `arg` was already a `KSequence`?

ehildenb: Yes, `KSequence` is able to flatten, but a test wolud be good.

ehildenb: I guess it would be good to also test for variables of sort `K` underneath the equals, to make sure they _don't_ get injected into the KSequence.

And to test for empty KSequence and non-empty KSequence under the equals.

nwatson22:
I confirmed this does fix the issue that originally appeared in KEVM.

PR #570, opened 2023-08-02
kompile: meaningful exception when `kompile` not found

### LINKED ISSUE #446: `kompile` returns `No such file or directory` when k is not installed
Using the `kompile` function from `pyk.ktool.kompile` returns the "No such file or directory" error when ran while k is not installed on the system https://github.com/runtimeverification/pyk/blob/master/src/pyk/ktool/kompile.py#L121. It can be a misleading error since it requires a file path input and the user might think that the path is wrong. Maybe running `which kompile` first to see if there is any path returned from the command might help to troubleshoot and display a more informative message.

h0nzZik:
Fixes #446

This makes it easier for the user to recognize from the exception that `kompile` was not found (in contrast to some other file being not found) when `ktool.kompile.kompile` fails, as suggested in https://github.com/runtimeverification/pyk/issues/446 .


src/pyk/ktool/kompile.py
```diff
@@ -114,6 +119,8 @@ def __call__(
             check_dir_path(abs_or_rel_to(include_dir, cwd or Path()))
 
         command = list(command) if command is not None else ['kompile']
+        if shutil.which(command[0]) is None:
+            raise KompileNotFoundException(command[0])
```
tothtamas28, line 123: I'd just go with a `RuntimeError` with a message:
```suggestion
            raise RuntimeError('Kompile command not found: {command[0]}')
```

src/pyk/ktool/kompile.py
```diff
@@ -114,6 +119,8 @@ def __call__(
             check_dir_path(abs_or_rel_to(include_dir, cwd or Path()))
 
         command = list(command) if command is not None else ['kompile']
+        if shutil.which(command[0]) is None:
```
tothtamas28, line 122:
```suggestion
        if not shutil.which(command[0]):
```

src/tests/integration/test_kompile.py
```diff
@@ -31,3 +33,12 @@ def test(self, definition_info: DefinitionInfo) -> None:
         assert definition_info.main_module_name == 'IMP'
         assert definition_info.syntax_module_name == 'IMP-SYNTAX'
         assert definition_info.timestamp >= 0
+
+
+def test_kompile_not_found() -> None:
+    k_file = K_FILES / 'imp-verification.k'
+    # this is supposed to NOT be found on $PATH
+    bad_kompile = 'bad-name-of-kompile'
```
tothtamas28, line 41: Consider monkeypatching the environment: https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html#monkeypatching-environment-variables

By mocking `run_process`, you can also make this a unit test if you want.

tothtamas28 (approved):

src/pyk/ktool/kompile.py
h0nzZik: I would like to test in the test that that it actually throws the right exception, without parsing the error message there. So what if I made `KompileNotFoundException` inherit from `RuntimeError`? `KompileNotFoundException.__init__` would take `command[0]` and called `RuntimeError.__init__(f'Kompile command not found: {cmd}')`.

tothtamas28: Yes, that's better.

> without parsing the error message there

You don't have to, you can just assert equivalence:

```python3
with pytest.raises(RuntimeError) as excinfo:
    ...

assert str(excinfo.value) == 'Kompile command not found: foo'
```

src/tests/integration/test_kompile.py
```diff
@@ -31,3 +33,13 @@ def test(self, definition_info: DefinitionInfo) -> None:
         assert definition_info.main_module_name == 'IMP'
         assert definition_info.syntax_module_name == 'IMP-SYNTAX'
         assert definition_info.timestamp >= 0
+
+
+def test_kompile_not_found(monkeypatch: pytest.MonkeyPatch) -> None:
+    k_file = K_FILES / 'imp-verification.k'
+    # this is supposed to NOT be found on $PATH
```
tothtamas28, line 40: Stale comment.

PR #571, opened 2023-08-02
refactoring of PrettyPrinter: extract base class `BasePrinter`

h0nzZik:


ehildenb:
@tothtamas28 was the one that suggested this architecture, so we should ask him if it meets the requirements.

ehildenb:
I personally think that things have been working fine without friction so far, so maybe this change is not needed. But if it improves something, without degrading performance, then that's good.

tothtamas28:
> @tothtamas28 was the one that suggested this architecture, so we should ask him if it meets the requirements.

What I had in mind with `BasePrinter` was just to have straightforward unparsing based on an (optional) `SymbolTable`:

```python3
class BasePrinter:
    _symbol_table: SymbolTable | None

    def __init__(self, symbol_table: SymbolTable | None = None):
        self._symbol_table = symbol_table

    def print(self, kast: KAst) -> str:
        # do simple dispatch to protected methods
        ...

    def _print_kas(self, kas: KAs) -> str:
        ...

    ...
```

> This must be done in a way that allows the `PrettyPrinter` class to: (i) only compute the symbol table a single time when calling it (memoize it), and (ii) only compute the symbol table if it's actually needed (compute it lazily)

Optional lazyness can be built in to `BasePrinter` (take a `SymbolTable | Callable[[], SymbolTable] | None`), but this is not necessary to enable `PrettyPrinter` to be lazy with the symbol table:

```python3
class PrettyPrinter:
    def __init__(self, ...):
        super().__init__()  # lazily initialize that attribute
```

PR #573, opened 2023-08-03
Factor out predicate to Subst logic into `Subst.from_pred` function

### LINKED ISSUE #525: Refactor the `KInner` predicate to `Substitution` transformation
As suggested by @ehildenb in https://github.com/runtimeverification/pyk/pull/519#discussion_r1248884597_, the logic of transforming a `KInner` predicate into a `Substitution` can be factored out as `@staticmethod Subst.from_pred(subst_pred: KInner) -> Subst` near here: https://github.com/runtimeverification/pyk/blob/49d861944bad260be137f68f4adc37b360e25771/src/pyk/kast/inner.py#L127

At the moment, this transformation is implemented in [KCFGExplore.cterm_get_model(...)](https://github.com/runtimeverification/pyk/blob/0ea352b3f084ca8afbcdb1fb665012671cc5f01f/src/pyk/kcfg/explore.py#L194) and [cterm_implies(...)](https://github.com/runtimeverification/pyk/blob/0ea352b3f084ca8afbcdb1fb665012671cc5f01f/src/pyk/kcfg/explore.py#L237) as follows:
```python
 model_subst = self.kprint.kore_to_kast(result.model)
  _subst: dict[str, KInner] = {}
  for subst_pred in flatten_label('#And', model_subst):
      subst_pattern = mlEquals(KVariable('###VAR'), KVariable('###TERM'))
      m = subst_pattern.match(subst_pred)
      if m is not None and type(m['###VAR']) is KVariable:
          _subst[m['###VAR'].name] = m['###TERM']
```

Moving this logic out to `@staticmethod Subst.from_pred(subst_pred: KInner) -> Subst` in a straightforward way several circular dependency issues.

nwatson22:
Closes #525 
Minor refactoring to #519.
As @palinatolmach pointed out, moving this logic into a static function of `Subst` creates circular dependencies, specifically with the `flatten_label` function. I'm not sure if it would be better to just re-implement the function as I've done here or to have the function `Subst.from_pred` as something like `subst_from_pred` in `pyk/kast/manip.py`.


src/pyk/kast/inner.py
```diff
@@ -123,6 +123,24 @@ def unapply(self, term: KInner) -> KInner:
...
+        def _flatten_label(label: str, kast: KInner) -> list[KInner]:
+            if type(kast) is KApply and kast.label.name == label:
+                items = (_flatten_label(label, arg) for arg in kast.args)
+                return [c for cs in items for c in cs]
+            return [kast]
+
+        _subst: dict[str, KInner] = {}
+        for subst_pred in _flatten_label('#And', substs_pred):
+            subst_pattern = KApply('#Equals', [KVariable('###VAR'), KVariable('###TERM')])
+            m = subst_pattern.match(subst_pred)
```
tothtamas28, line 137: Consider using `match-case` instead ([PEP 636](https://peps.python.org/pep-0636/)):

```python3
match subst_pred:
    case KApply(KLabel('#Equals'), KVariable(var), term):
        _subst[var] = term
    case _:
        raise ValueError(...)
```

src/pyk/kast/inner.py
```diff
@@ -123,6 +123,24 @@ def unapply(self, term: KInner) -> KInner:
...
+            if m is not None and type(m['###VAR']) is KVariable:
+                _subst[m['###VAR'].name] = m['###TERM']
+            else:
+                raise AssertionError(f'Received a non-substitution from get-model endpoint: {subst_pred}')
```
tothtamas28, line 141:
```suggestion
                raise ValueError(f'Invalid substitution predicate: {subst_pred}')
```

src/pyk/kast/inner.py
```diff
@@ -123,6 +123,24 @@ def unapply(self, term: KInner) -> KInner:
             new_term = KRewrite(lhs, rhs).replace(new_term)
         return new_term
 
+    @staticmethod
+    def from_pred(substs_pred: KInner) -> Subst:
```
tothtamas28, line 127:
```suggestion
    def from_pred(pred: KInner) -> Subst:
```

src/pyk/kast/inner.py
```diff
@@ -123,6 +123,24 @@ def unapply(self, term: KInner) -> KInner:
...
+        def _flatten_label(label: str, kast: KInner) -> list[KInner]:
+            if type(kast) is KApply and kast.label.name == label:
+                items = (_flatten_label(label, arg) for arg in kast.args)
+                return [c for cs in items for c in cs]
+            return [kast]
+
+        _subst: dict[str, KInner] = {}
+        for subst_pred in _flatten_label('#And', substs_pred):
```
tothtamas28, line 135:
```suggestion
        for conjunct in _flatten_label('#And', substs_pred):
```

src/pyk/kast/inner.py
```diff
@@ -123,6 +123,24 @@ def unapply(self, term: KInner) -> KInner:
...
 
+    @staticmethod
+    def from_pred(substs_pred: KInner) -> Subst:
+        def _flatten_label(label: str, kast: KInner) -> list[KInner]:
+            if type(kast) is KApply and kast.label.name == label:
+                items = (_flatten_label(label, arg) for arg in kast.args)
+                return [c for cs in items for c in cs]
+            return [kast]
+
+        _subst: dict[str, KInner] = {}
```
tothtamas28, line 134:
```suggestion
        subst: dict[str, KInner] = {}
```

src/pyk/kcfg/explore.py
```diff
@@ -130,15 +130,7 @@ def cterm_get_model(self, cterm: CTerm, module_name: str | None = None) -> Subst
...
-            _subst: dict[str, KInner] = {}
-            for subst_pred in flatten_label('#And', model_subst):
-                subst_pattern = mlEquals(KVariable('###VAR'), KVariable('###TERM'))
-                m = subst_pattern.match(subst_pred)
-                if m is not None and type(m['###VAR']) is KVariable:
-                    _subst[m['###VAR'].name] = m['###TERM']
-                else:
-                    raise AssertionError(f'Received a non-substitution from get-model endpoint: {subst_pred}')
-            return Subst(_subst)
+            return Subst.from_pred(model_subst)
```
tothtamas28, line 133:
```suggestion
            try:
                return Subst.from_pred(model_subst)
            except ValueError as err:
                raise AssertionError(f'Received a non-substitution from get-model endpoint: {subst_pred}') fro err
```

src/pyk/prelude/k.py
```diff
@@ -3,6 +3,6 @@
 from ..kast.inner import KSort, KToken
 
 K: Final = KSort('K')
-GENERATED_TOP_CELL: Final = KSort('GeneratedTopCell')
+GENERATED_TOP_CELL: Final[KSort] = KSort('GeneratedTopCell')
```
tothtamas28, line 6: Type checker should be able to infer that. Are there type errors related to this?

tothtamas28 (changes requested):
Please add unit tests (for both successful cases and exceptions).

src/pyk/kast/inner.py
```diff
@@ -123,6 +123,24 @@ def unapply(self, term: KInner) -> KInner:
             new_term = KRewrite(lhs, rhs).replace(new_term)
         return new_term
 
+    @staticmethod
+    def from_pred(substs_pred: KInner) -> Subst:
+        def _flatten_label(label: str, kast: KInner) -> list[KInner]:
```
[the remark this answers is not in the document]
nwatson22: This is intentionally duplicated because trying to import `flatten_label`  from `kast/manip.py` in `kast/inner.py` results in a circular import error. I was wondering if you had an opinion about the best way to deal with this.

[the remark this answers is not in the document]
tothtamas28: What do you think about the two options above?

src/pyk/kast/inner.py
nwatson22: Done.

src/pyk/kast/inner.py
nwatson22: Done.

src/pyk/kast/inner.py
nwatson22: Done.

src/pyk/kast/inner.py
nwatson22: Done.

src/pyk/kcfg/explore.py
nwatson22: Done.

src/pyk/prelude/k.py
nwatson22: I've removed it. It was giving a type error but only when the circular dependency was active.

src/pyk/kast/inner.py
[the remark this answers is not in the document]
nwatson22: The ones I mentioned? I guess I would slightly lean toward keeping the duplicate local function and keep this as a static method of `Subst`.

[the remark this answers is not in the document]
tothtamas28: I mean the two options I suggested in this thread. Specifically:

> Okay, now I see this is to avoid introducing circular imports. You can either:
> * Move `flatten_label` to `kast.inner`
> * Import locally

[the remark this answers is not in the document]
nwatson22: I changed it to a local import.

nwatson22:
Added tests for this function.

src/pyk/kast/inner.py
```diff
@@ -123,6 +123,20 @@ def unapply(self, term: KInner) -> KInner:
             new_term = KRewrite(lhs, rhs).replace(new_term)
         return new_term
 
+    @staticmethod
+    def from_pred(pred: KInner) -> Subst:
+        from .manip import flatten_label
+
+        subst: dict[str, KInner] = {}
+        for conjunct in flatten_label('#And', pred):
+            print(conjunct)
```
tothtamas28, line 132:
```suggestion
```

src/tests/unit/kast/test_subst.py
```diff
@@ -174,3 +174,52 @@ def test_propagate_subst() -> None:
...
+    ML_SUBST_FROM_PRED_TEST_DATA,
+    ids=[test_id for test_id, *_ in ML_SUBST_FROM_PRED_TEST_DATA],
+)
+def test_subst_from_pred(test_id: str, pred: KInner, expected_subst: Subst) -> None:
+    if expected_subst:
+        subst = Subst.from_pred(pred)
+        assert subst == expected_subst
+    else:
+        with pytest.raises(ValueError):
+            _ = Subst.from_pred(pred)
```
tothtamas28, line 225:
```suggestion
            Subst.from_pred(pred)
```

tothtamas28 (approved):

src/pyk/kcfg/explore.py
```diff
@@ -130,15 +130,11 @@ def cterm_get_model(self, cterm: CTerm, module_name: str | None = None) -> Subst
...
-                m = subst_pattern.match(subst_pred)
-                if m is not None and type(m['###VAR']) is KVariable:
-                    _subst[m['###VAR'].name] = m['###TERM']
-                else:
-                    raise AssertionError(f'Received a non-substitution from get-model endpoint: {subst_pred}')
-            return Subst(_subst)
+            try:
+                return Subst.from_pred(model_subst)
+            except ValueError as err:
+                raise AssertionError(f'Received a non-substitution from get-model endpoint: {model_subst}') from err
```
ehildenb, line 136: Do we need to catch and re-raise this exception?

```suggestion
            return Subst.from_pred(model_subst)
```

The stack trace + original exception should be enough information I guess?

tothtamas28: The difference is subtle, but the `AssertionError` suggests that there is a bug in the `get-model` endpoint.

PR #576, opened 2023-08-03
Changes needed to integrate with Maude Backend

dwightguth:
This PR contains the changes needed to be able to easily modify pyk into a version which can call the maude backend for its integration and unit tests rather than the Haskell backend.

The changes required are as follows:

1. We need to support http transport in KoreClient. This means that we need a Transport abstract base class that gets instantiated with either SingleSocketTransport (the current default), or HttpTransport, representing an HTTP query. We also create a DefaultTransport alias that can be used everywhere KoreClient is constructed so it's easy to switch the default transport being used.
2. We need to add a MaudeKompile class and a KompileBackend.MAUDE enum value.


src/pyk/kore/rpc.py
```diff
@@ -521,8 +616,8 @@ class KoreClient(ContextManager['KoreClient']):
 
     _client: JsonRpcClient
 
-    def __init__(self, host: str, port: int, *, timeout: int | None = None, bug_report: BugReport | None = None):
-        self._client = JsonRpcClient(host, port, timeout=timeout, bug_report=bug_report)
+    def __init__(self, transport: Transport, *, bug_report: BugReport | None = None):
+        self._client = JsonRpcClient(transport, bug_report=bug_report)
```
tothtamas28, line 620: As both transport implementations expect the same constructor parameters, `HTTPTransport` can be introduced in a backward compatible way:

```python3
def RpcTransport(Enum):
    SINGLE_SOCKET = auto()
    HTTP = auto()

    def create(self, host: str, port: int, timeout: int | None) -> Transport:
        ...

class KoreClient:
    def __init__(self, host: str, port: int, *, timeout: int | None = None, transport: RpcTransport = RpcTransport.SINGLE_SOCKET, bug_report: BugReport | None = None):
        rpc_transport = transport.create(host, port, timeout)
        self._client = JsonRpcClient(transport, bug_report=bug_report)
```

The name `RpcTransport` is a bit clumsy, but if you rename `Transport` (which is no longer public), then the enum on the API can have a nicer name.

tothtamas28: Creation of the `Transport` can even be pushed down to `JsonRpcClient`, this way you do not need to patch as many things in the unit tests.

src/tests/unit/kore/test_client.py
```diff
@@ -46,23 +47,25 @@ def __init__(self, mock: Mock):
         self.mock = mock
 
     def assume_response(self, response: Any) -> None:
-        self.mock.request.return_value = response
+        self.mock.request.return_value = json.dumps({'jsonrpc': '2.0', 'id': 1, 'result': response})
```
tothtamas28, line 50: With the suggestions above the tests should work with minimal / zero modification.

src/pyk/kore/rpc.py
```diff
@@ -93,6 +107,101 @@ def close(self) -> None:
...
+        ]
+
+    def request(self, req: str) -> str:
+        self._sock.sendall(req.encode())
+        server_addr = str(self)
+        _LOGGER.debug(f'Waiting for response from {server_addr}...')
+        return self._file.readline().rstrip()
+
+    def __str__(self) -> str:
+        return f'{self._host}:{self._port}'
```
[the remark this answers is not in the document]
dwightguth: What exactly are you suggesting? I'm happy to make a change but this comment is too vague to be actionable currently.

src/pyk/kore/rpc.py
dwightguth: I'd rather keep this as is; eventually we probably will want transport options to include unix domain sockets which definitely do not take the same parameters.

dwightguth: If you know of a way to approximate the overloading of constructors in python I'd be happy to have both ways of initializing KoreClient, but I don't know exactly how to do that.

src/pyk/kore/rpc.py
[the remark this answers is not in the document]
tothtamas28: `__str__` is supposed to represent the object.

The method is used here:

```python3
server_addr = str(self)
_LOGGER.debug(f'Waiting for response from {server_addr}...')
```

So what I mean, this overload of `__str__` is specific for this context. Consider introducing a private method instead.

src/pyk/kore/rpc.py
tothtamas28: > eventually we probably will want transport options to include unix domain sockets which definitely do not take the same parameters

If this happens we can still make this change. Until then I'd prefer to keep changes to the minimum and not break the API.

src/pyk/kore/rpc.py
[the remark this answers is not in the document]
dwightguth: Okay, I understand.

src/pyk/kore/rpc.py
dwightguth: I mean, a breaking API change will be needed at some point. The code is already written and tested. Are there other repos out there that might consume this class? If so, I can understand not wanting to make the change since it would require rolling updates elsewhere. If not, I'm not sure I really see the benefit.

tothtamas28: Yes, `KoreClient` is used extensively in `evm-semantics` / `kevm-pyk`.

dwightguth: okay. That makes sense then. We can leave off this change for when it's needed.

src/pyk/kore/rpc.py
```diff
@@ -44,23 +45,47 @@ def __init__(self, message: str, code: int, data: Any = None):
...
+        ...
+
+    @abstractmethod
+    def description(self) -> str:
+        ...
+
+
+class TransportType(Enum):
+    SINGLE_SOCKET = auto
+    HTTP = auto
```
tothtamas28, line 74:
```suggestion
    SINGLE_SOCKET = auto()
    HTTP = auto()
```

src/pyk/kore/rpc.py
```diff
@@ -82,7 +107,7 @@ def _create_connection(host: str, port: int, timeout: int | None) -> socket.sock
 
         raise RuntimeError(f'Connection timed out: {host}:{port}')
 
-    def __enter__(self) -> JsonRpcClient:
+    def __enter__(self) -> SingleSocketTransport:
         return self
 
     def __exit__(self, *args: Any) -> None:
```
tothtamas28, line 113: This is probably not necessary to override.

src/pyk/kore/rpc.py
```diff
@@ -82,7 +107,7 @@ def _create_connection(host: str, port: int, timeout: int | None) -> socket.sock
 
         raise RuntimeError(f'Connection timed out: {host}:{port}')
 
-    def __enter__(self) -> JsonRpcClient:
+    def __enter__(self) -> SingleSocketTransport:
```
tothtamas28, line 110: This isn't either, the abstract type signature is sufficient.

src/pyk/kore/rpc.py
```diff
@@ -93,6 +118,111 @@ def close(self) -> None:
...
+    def __init__(self, host: str, port: int, *, timeout: int | None = None):
+        self._host = host
+        self._port = port
+        self._timeout = timeout
+
+    def __enter__(self) -> HttpTransport:
+        return self
+
+    def __exit__(self, *args: Any) -> None:
+        pass
```
tothtamas28, line 159: Can be removed.

src/pyk/kore/rpc.py
```diff
@@ -93,6 +118,111 @@ def close(self) -> None:
...
+        timeout: int | None = None,
+        bug_report: BugReport | None = None,
+        transport: TransportType = TransportType.SINGLE_SOCKET,
+    ):
+        if transport is TransportType.SINGLE_SOCKET:
+            self._transport = SingleSocketTransport(host, port, timeout=timeout)
+        elif transport is TransportType.HTTP:
+            self._transport = HttpTransport(host, port, timeout=timeout)
+        else:
+            raise AssertionError
```
tothtamas28, line 213:
```suggestion
            raise AssertionError()
```

tothtamas28 (approved):

PR #578, opened 2023-08-04
Refactor of the `KCFGViewer`

iFrostizz:
Going in a more step-by-step fashion compared to https://github.com/runtimeverification/pyk/pull/460
Rolling out our own key bindings was messing up with the defaults of textual and adding a lot of unecessary boilerpalte. This PR also closes it.

- Dedicated classes for each widget in the `NodeView` (Term, Constraint, Info, Custom)
- Relevant ones all inherits from `ScrollableContainer` and can be focused. Adds native key bindings and can be selected using `Tab` or `Tab+Shift` to loop the other way around
- using an absolute path for the css file so that the class `APRProofViewer` inherits `KCFGViewer`'s css. `@import` is not something that works with textual, but as we may want in the future to add stuff only for the `APRProofViewer` css.


iFrostizz:
if we want to add a css file specific to the `APRProofViewer`, we could use https://textual.textualize.io/guide/CSS/#using-multiple-css-files

iFrostizz:
I'm having a hard time uploading a GIF that fits the size limit
untitled

tothtamas28 (commented):
Looks really good.

@ehildenb, please check as well.

src/pyk/kcfg/style.css
```diff
@@ -4,58 +4,66 @@
...
     height: 15fr;
     border: solid;
     overflow: auto scroll;
 }
-#info {
-    width: auto;
-    height: auto;
+#custom:focus {
+    border: solid blue;
 }
```
ehildenb, line 56: For these, are we able to facotr out a common block like:

```
.focus {
    border: solid blue;
}
```

instead? Not sure it will work, but seems intuitive based on giving that attribute to each different type of class here.

ehildenb (approved):
Looks really nice!! Thanks @iFrostizz !!

ehildenb:
I assume you tested that hiding/unhiding individual views works? Asknig because it looks like some container IDs changed, and so we need to make sure it's still able to apply the hiddent attribute.

src/pyk/kcfg/style.css
iFrostizz: That did not worked. `:focus` is a pseudo class and writing it alone like

```css
:focus {
    border: solid blue;
}
```

didn't worked either.

iFrostizz:
Yes @ehildenb that does work!

PR #579, opened 2023-08-04
make next_states field in ExecuteResult a set

dwightguth:
The Maude backend is not necessarily guaranteed to return states in the same  order as the Haskell backend or the Booster. Thus, we would prefer to maintain an unordered representation of follower states in order to prevent frivolous test failures.


ehildenb:
Hmmmm, I'm worried this will have a negative performance impact though. Tuples (being immutable) generally perform better I believe because of how the Python runtime works. Can we instead modify the testing harness to convert them to sets before comparison?

I'm not sure about the performance issue though, maybe @tothtamas28 knows more.

dwightguth:
Okay, I'll give this a try.

PR #583, opened 2023-08-07
Update `poetry.lock`

tothtamas28:


flake.nix
```diff
@@ -18,33 +18,6 @@
               groups = [ ];
               # We remove `"dev"` from `checkGroups`, so that poetry2nix does not try to resolve dev dependencies.
               checkGroups = [ ];
-              overrides = prev.poetry2nix.overrides.withDefaults
```
tothtamas28, line 21: @goodlyrottenapple Nix seems to be able to build the derivations even without these overrides now. It's safe to remove this part then, right?

goodlyrottenapple: Yes, they have likely been fixed upstream

goodlyrottenapple (approved):

PR #587, opened 2023-08-08
Add support for vacuous nodes

anvacaru:
With https://github.com/runtimeverification/haskell-backend/pull/3637, kore-rpc returns a new response with a `vacuous` stop reason and the current state before rewriting the state to `#Bottom`.

Changes in this PR:
- added a new set `vacuous` to the kcfg. The reason to add the set here and not in the `APRProof` is that the response comes from the backend, and is not semantics related.
- modified `cterm_execute` to return if the result is vacuous.
- modified `extend` to add a `vacuous` node to the kcfg if `cterm_execute` gets a `vacuous` response.
- modified `APRProof` and `APRBMCProof` to prevent marking `vacuous` nodes as `pending` or `failing`.
- modified proof summaries to include `vacuous`.

Screenshot 2023-08-08 at 16 42 48


src/pyk/kcfg/explore.py
```diff
@@ -296,7 +297,7 @@ def step(
         if len(successors) != 0 and type(successors[0]) is KCFG.Split:
             raise ValueError(f'Cannot take step from split node {self.id}: {shorten_hashes(node.id)}')
         _LOGGER.info(f'Taking {depth} steps from node {self.id}: {shorten_hashes(node.id)}')
-        actual_depth, cterm, next_cterms, next_node_logs = self.cterm_execute(
+        _, actual_depth, cterm, next_cterms, next_node_logs = self.cterm_execute(
```
iFrostizz, line 300: Why not checking and flagging the node when the returned status is vacuous here ?

src/pyk/kcfg/kcfg.py
```diff
@@ -205,6 +205,7 @@ def edges(self) -> tuple[KCFG.Edge, ...]:
     _covers: dict[int, dict[int, Cover]]
     _splits: dict[int, Split]
     _ndbranches: dict[int, NDBranch]
+    _vacuous: set[int]
```
iFrostizz, line 208: I think that vacuous nodes should be stored in the `Proof` classes https://github.com/runtimeverification/pyk/pull/528#issuecomment-1628696579

src/pyk/kore/rpc.py
```diff
@@ -450,6 +452,28 @@ def _check_reason(cls: type[ER], dct: Mapping[str, Any]) -> None:
...
+
+    @classmethod
+    def from_dict(cls: type[VacuousResult], dct: Mapping[str, Any]) -> VacuousResult:
+        cls._check_reason(dct)
+        logs = tuple(LogEntry.from_dict(l) for l in dct['logs']) if 'logs' in dct else ()
+        return VacuousResult(
+            state=State.from_dict(dct['state']),
+            depth=dct['depth'],
+            logs=logs,
+        )
```
iFrostizz, line 474: You could move this class under `TerminalResult` for consistency according to the enum

src/pyk/kcfg/kcfg.py
anvacaru: Stuck nodes are not stored in the `Proof` classes either. I stored `vacuous` nodes in the kcfg because I don't consider it part of semantics information.

src/pyk/kore/rpc.py
anvacaru: I don't think I understand. Each item in the `StopReason` enum has its own class, extending the `ExecuteResult` class.

src/pyk/kcfg/explore.py
anvacaru: Hmm, good point. I didn't consider it part of the function's scope.

src/pyk/kore/rpc.py
```diff
@@ -450,6 +452,28 @@ def _check_reason(cls: type[ER], dct: Mapping[str, Any]) -> None:
...
+
+
```
iFrostizz, line 476:
```suggestion
@final
@dataclass(frozen=True)
class VacuousResult(ExecuteResult):
    reason = StopReason.VACUOUS
    next_states = None
    rule = None

    state: State
    depth: int
    logs: tuple[LogEntry, ...]

    @classmethod
    def from_dict(cls: type[VacuousResult], dct: Mapping[str, Any]) -> VacuousResult:
        cls._check_reason(dct)
        logs = tuple(LogEntry.from_dict(l) for l in dct['logs']) if 'logs' in dct else ()
        return VacuousResult(
            state=State.from_dict(dct['state']),
            depth=dct['depth'],
            logs=logs,
        )


```

src/pyk/kore/rpc.py
iFrostizz: It's just for following the `ExecuteResult` structure, the order is
```python
StopReason.STUCK: 'StuckResult',
StopReason.DEPTH_BOUND: 'DepthBoundResult',
StopReason.BRANCHING: 'BranchingResult',
StopReason.CUT_POINT_RULE: 'CutPointResult',
StopReason.TERMINAL_RULE: 'TerminalResult',
StopReason.VACUOUS: 'VacuousResult',
 ```
 
 And the classes are following it:
 
 ```python
 class StuckResult
    ...
 class DepthBoundResult
    ...
class BranchingResult
    ...
class CutPointResult
    ...
class TerminalResult
    ...
```

You could write the `VacuousResult` class below rather than on top:

```python
class TerminalResult
    ...
class VacuousResult
    ...
```

anvacaru: Oh, my bad. Sure, I can reorder them!

src/pyk/kcfg/explore.py
```diff
@@ -88,18 +88,19 @@ def cterm_execute(
             log_successful_simplifications=self._trace_rewrites,
             log_failed_simplifications=self._trace_rewrites,
         )
+        _is_vacuous = er.reason == StopReason.VACUOUS
```
tothtamas28, line 91:
```suggestion
        _is_vacuous = er.reason is StopReason.VACUOUS
```

src/pyk/kore/rpc.py
```diff
@@ -561,6 +563,28 @@ def from_dict(cls: type[TerminalResult], dct: Mapping[str, Any]) -> TerminalResu
         return TerminalResult(state=State.from_dict(dct['state']), depth=dct['depth'], rule=dct['rule'], logs=logs)
 
 
+@final
+@dataclass(frozen=True)
+class VacuousResult(ExecuteResult):
```
tothtamas28, line 568: Please add tests for this:
* https://github.com/runtimeverification/pyk/blob/master/src/tests/unit/kore/test_client.py
* https://github.com/runtimeverification/pyk/blob/master/src/tests/integration/kore/test_kore_client.py

src/pyk/kcfg/kcfg.py
tothtamas28, line 208: Please add tests for this:
* https://github.com/runtimeverification/pyk/blob/master/src/tests/unit/test_kcfg.py

src/pyk/kore/rpc.py
```diff
@@ -561,6 +563,28 @@ def from_dict(cls: type[TerminalResult], dct: Mapping[str, Any]) -> TerminalResu
...
+    reason = StopReason.VACUOUS
```
tothtamas28, line 569: Under what condition is this returned as `reason`? I try to understand the difference between cases when the backend just prunes a branch and when this is returned.

src/pyk/kcfg/kcfg.py
anvacaru: Done.

src/pyk/kore/rpc.py
anvacaru: I've discussed this with @goodlyrottenapple: `vacuous` is returned when the backend applies rules that all go to `#Bottom` after simplification. If multiple rules apply and only some of them go to `#Bottom`,  then those are pruned and the execution continues.

iFrostizz:
@anvacaru now that pyk can handle Bottom and Top responses from the backend, a Bottom `CTerm` when building the kcfg could be marked as vacuous https://github.com/runtimeverification/pyk/pull/621

tothtamas28 (approved):

PR #593, opened 2023-08-09
add function to assert equality between ExecuteResults

dwightguth:
Better fix for #579 

We need to be able to compare next_states as if it were a set and not a tuple. This was already being done in one unit test but I wanted a more general solution. This solution is adapted from the one @tothtamas28 suggested in the linked PR.


src/tests/integration/kore/test_kore_client.py
```diff
@@ -211,6 +212,19 @@ def state(n: int) -> State:
...
 
+def assert_execute_result_equals(actual: ExecuteResult, expected: ExecuteResult) -> None:
+    assert actual.__class__ == expected.__class__
+
+    actual_dict = asdict(actual)  # type: ignore
+    expected_dict = asdict(expected)  # type: ignore
+    if 'next_states' in actual_dict and 'next_states' in expected_dict:
+        if actual_dict['next_states'] is not None and expected_dict['next_states'] is not None:
+            actual_dict['next_states'] = set(actual.next_states)  # type: ignore
+            expected_dict['next_states'] = set(expected.next_states)  # type: ignore
```
tothtamas28, line 223:
```suggestion
    if 'next_states' in actual_dict:
        actual_dict['next_states'] = set(actual.next_states) if actual.next_states is not None else None
    if 'next_states' in expected_dict:
        expected_dict['next_states'] = set(expected.next_states) if expected.next_states is not None else None
```

src/tests/integration/kore/test_kore_client.py
```diff
@@ -211,6 +212,19 @@ def state(n: int) -> State:
 )
 
 
+def assert_execute_result_equals(actual: ExecuteResult, expected: ExecuteResult) -> None:
+    assert actual.__class__ == expected.__class__
```
tothtamas28, line 216:
```suggestion
    assert type(actual) == type(expected)
```

tothtamas28 (approved):

PR #595, opened 2023-08-10
Handle empty response exception from the kore client

iFrostizz:
When the response from the `JsonRpcClient` is empty (usually happens when there is a memory leak in the backend), no handling of the response is done and the proof will crash on a JsonDecodeError that may not be very meaningful for the user.
This PR is a fix for that that returns a `JsonRpcError` with the appropriate code and tests.


src/pyk/kore/rpc.py
```diff
@@ -232,6 +232,8 @@ def request(self, method: str, **params: Any) -> dict[str, Any]:
 
         _LOGGER.debug(f'Sending request to {server_addr}: {req}')
         resp = self._transport.request(req)
+        if len(resp) == 0:
```
tothtamas28, line 235:
```suggestion
        if not resp:
```

src/pyk/kore/rpc.py
```diff
@@ -232,6 +232,8 @@ def request(self, method: str, **params: Any) -> dict[str, Any]:
...
+            raise JsonRpcError('Parse error: Empty response received', -32700)
```
tothtamas28, line 236: This indicates that the request received by the server cannot be parsed. Instead, this should raise a `RuntimeError` or similar.

src/tests/unit/kore/test_client.py
```diff
@@ -256,3 +258,66 @@ def test_add_module(
 
     # Then
     rpc_client.assert_request('add-module', **params)
+
+
+class MockTransport:
```
tothtamas28, line 263: Consider moving these tests to a new module (e.g. `test_rpc_client.py`). Then there's no conflict with old names and things can have more concise names.

src/pyk/kore/rpc.py
iFrostizz: Thanks!

src/pyk/kore/rpc.py
iFrostizz: Makes sense, thank you!

src/tests/unit/kore/test_client.py
iFrostizz: Now moved to a new module, thanks!

src/pyk/kore/rpc.py
```diff
@@ -232,6 +232,8 @@ def request(self, method: str, **params: Any) -> dict[str, Any]:
 
         _LOGGER.debug(f'Sending request to {server_addr}: {req}')
         resp = self._transport.request(req)
+        if not resp:
+            raise RuntimeError('Parse error: Empty response received')
```
tothtamas28, line 236:
```suggestion
            raise RuntimeError('Empty response received')
```

tothtamas28 (approved):
Looks good, thank you!

Issue #548, opened 2023-07-25
Profile `kore-rpc` use


tothtamas28:
Add a profile test that exercises `KoreClient` on a non-trivial proof.

ehildenb:
I've made a simple script for extracting transit times from the `--verbose` logs of a pyk execution:

```
#!/usr/bin/env bash
    
import sys
    
from datetime import datetime
    
buckets = [
    ('kast_to_kore start', 'kast_to_kore'),
    ('kore_to_kast start', 'kore_to_kast'),
    ('- simplify', 'rpc simplify'),
    ('- add-module', 'rpc add-module'),
    ('- execute', 'rpc execute'),
    ('- implies', 'rpc implies'),
]   
    
with open(sys.argv[1], 'r') as f:
    lines = f.readlines()
    
def split_line(line: str) -> tuple[int, str, str]:
    _, date, time, loc, _, *log = line.split(' ')
    time_int = datetime.strptime(f'{date} {time}', '%Y-%m-%d %H:%M:%S,%f')
    logs = ' '.join(log).strip()
    return (time_int, loc, logs)

times = {}
other_times = []
total_time = 0.0
curr_time, _, curr_log = split_line(lines[0])
for line in lines[1:]:
    new_time, _, new_log = split_line(line)
    time_delta = (new_time - curr_time).total_seconds()
    total_time += time_delta
    found = False
    for pattern, bucket in buckets:
        if curr_log.find(pattern) >= 0:
            if bucket not in times:
                times[bucket] = []
            times[bucket].append((time_delta, curr_log))
            found = True
            break
    if not found:
        if 'other' not in times:
            times['other'] = []
        times['other'].append((time_delta, curr_log))
    curr_time = new_time
    curr_log = new_log

for time in times:
    actual_times = [t for t, _ in times[time]]
    log_entries = set([le for _, le in times[time]])
    num_entries = len(actual_times)
    transit_times = sorted(actual_times)
    total = sum(transit_times)
    percent = total / total_time
    avg = total / num_entries
    median = transit_times[int(num_entries / 2)]
    max = transit_times[-1]
    min = transit_times[0]
    print()
    print(f'Transit: {time}')
    print(f'    total: {total}')
    print(f'    percent: {percent}')
    print(f'    avg: {avg}')
    print(f'    median: {median}')
    print(f'    min: {min}')
    print(f'    max: {max}')
    if time == 'other':
        print(f'    log entries: {log_entries}')
```

[...]

PR #601, opened 2023-08-11
KCFGExploration: KCFG + terminal nodes

PetarMax:
This PR introduces the `KCFGExploration` class, which encapsulates the `KCFG` class with information about `terminal` and `explorable` nodes and associated reasoning. The main changes to existing classes, in addition to the adaptation of using `KCFGExploration` instead of `KCFG`, are as follows:

- In `KCFGExplore`, the `extend` function now checks automatically whether or not newly added nodes are `terminal`. This removes the need for these checks inside the provers.
- In `APRProof` and `APRBMCProof`, the notions of `pending` and `failing` nodes have been slightly reformulated
- In `APRProver` and `APRBMCProver`, subsumption is cached and checked for`terminal` nodes at the end of every prover iteration


PetarMax:
@ehildenb @tothtamas28 I need some feedback on this, when you have a moment, please. This, I believe is also related to [this PR](https://github.com/runtimeverification/evm-semantics/pull/1978) - I believe that this PR might no longer be needed in the presence of `KCFGExploration`.

src/pyk/kcfg/exploration.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # Terminal node collector
+    @property
+    def terminals(self) -> list[KCFG.Node]:
+        return [node for node in self.kcfg.nodes if node.id in self._terminals]
+
+    # Explorable node collector
+    @property
+    def explorables(self) -> list[KCFG.Node]:
+        return [node for node in self.kcfg.leaves if self.is_explorable(node.id)]
```
ehildenb, line 46: For other node sets, we've been using non-plural, should we keep that convention? So it would be `def terminal` and `def explorable`.

Not a strong preference, but I do like consistency.

src/pyk/kcfg/explore.py
```diff
@@ -354,33 +355,52 @@ def _check_abstract(self, node: KCFG.Node, kcfg: KCFG) -> bool:
         kcfg.create_cover(node.id, new_node.id)
         return True
 
+    def _check_terminal(self, node: KCFG.Node) -> bool:
+        _LOGGER.info(f'Checking terminal: {shorten_hashes(node.id)}')
+        if self.kcfg_semantics.is_terminal(node.cterm):
+            _LOGGER.info(f'Terminal node: {shorten_hashes(node.id)}.')
+            return True
+        return False
```
ehildenb, line 363:
```suggestion
    def _check_terminal(self, node: KCFG.Node) -> bool:
        _LOGGER.info(f'Checking terminal: {node.id}')
        if self.kcfg_semantics.is_terminal(node.cterm):
            _LOGGER.info(f'Terminal node: {node.id}.')
            return True
        return False
```

Node IDs are `int` now, I can open a PR changing the rest.

src/pyk/proof/reachability.py
```diff
@@ -207,20 +192,20 @@ def from_claim(
...
+        apr_proof = APRProof(claim.label, KCFGExploration(KCFG()), init=0, target=0, logs=logs, **kwargs)
 
         kcfg_dir = apr_proof.proof_subdir / 'kcfg' if apr_proof.proof_subdir else None
 
-        cfg, init_node, target_node = KCFG.from_claim(defn, claim, cfg_dir=kcfg_dir)
-        apr_proof.kcfg = cfg
+        kcfg, init_node, target_node = KCFG.from_claim(defn, claim, cfg_dir=kcfg_dir)
+        apr_proof.kcfg_exploration.kcfg = kcfg
         apr_proof.init = init_node
         apr_proof.target = target_node
```
ehildenb, line 202: It seems more idiomatic to do this:

```
kcfg, init_node, target_node = KCFG.from_claim(defn, claim, cfg_dir=kcfg_dir)
return APRProof(claim.label, KCFGExploration(kcfg), init=init_node, target=target_node, logs=logs, **kwargs)
```

More compact too.

src/pyk/proof/reachability.py
```diff
@@ -323,36 +306,36 @@ def write_proof_data(self) -> None:
...
-        self.kcfg.write_cfg_data()
+        self.kcfg_exploration.kcfg.write_cfg_data()
 
 
 class APRBMCProof(APRProof):
     """APRBMCProof and APRBMCProver perform bounded model-checking of an all-path reachability logic claim."""
 
     bmc_depth: int
-    _bounded_nodes: set[NodeIdLike]
+    _bounded: set[int]
```
ehildenb, line 328: Why the tighter typing?

src/pyk/proof/reachability.py
```diff
@@ -592,32 +578,48 @@ def __init__(
...
 
-        self._target_is_terminal = kcfg_explore.kcfg_semantics.is_terminal(
-            self.proof.kcfg.node(self.proof.target).cterm
-        )
+        self._target = self.proof.kcfg_exploration.kcfg.node(self.proof.target)
+        self._target_is_terminal = kcfg_explore._check_terminal(self._target)
+
+        # All nodes marked as terminal are semantically terminal
+        for node in self.proof.terminals:
+            assert kcfg_explore._check_terminal(node)
```
ehildenb, line 586: Do we need to check this at load time?

src/pyk/proof/reachability.py
```diff
@@ -592,32 +578,48 @@ def __init__(
...
+
+        # Target node is not marked as terminal
+        assert self._target.id not in self.proof.kcfg_exploration._terminals
```
ehildenb, line 589: Also this? Why can we not have target node be terminal? It is actually usually the case.

src/pyk/proof/reachability.py
```diff
@@ -592,32 +578,48 @@ def __init__(
...
+
+        # All non-target leaves not marked as terminal are checked for terminalness
+        for node in [
+            node
+            for node in self.proof.kcfg_exploration.kcfg.nodes
+            if not self.proof.is_terminal(node.id) and not self.proof.is_target(node.id)
+        ]:
+            if kcfg_explore._check_terminal(node):
+                self.proof.kcfg_exploration.add_terminal(node.id)
+
```
ehildenb, line 599: Why don't we just check all the nodes in totality?

Also, if a user is constructing a given `APRProof`, based on an existing `KCFGExploration`, shouldn't it be `KCFGExploration` that takes care of these sanity checks, because that's where `terminal` notion lives? `APRProof` should just assume it's handed a valid `KCFGExploration`.

src/pyk/proof/reachability.py
```diff
@@ -592,32 +578,48 @@ def __init__(
...
+            )
+            csubst = self.kcfg_explore.cterm_implies(node.cterm, self._target.cterm)
+            if csubst is not None:
+                self.proof.kcfg_exploration.kcfg.create_cover(node.id, self._target.id, csubst=csubst)
+                _LOGGER.info(f'Subsumed into target node {self.proof.id}: {shorten_hashes((node.id, self._target.id))}')
+                self._subsumption_checks[node.id] = True
+            else:
+                self._subsumption_checks[node.id] = False
+
+        return self._subsumption_checks[node.id]
```
ehildenb, line 622: Do we really need to store the extra set `self._subsumption_checks`? I guess it's pretty easily computed as `len(self.kcfg_exploration.kcfg.covers(source=node.id, target=self.target)) > 0`.

src/pyk/kcfg/explore.py
PetarMax: Ah, I see, ok

src/pyk/kcfg/kcfg.py
```diff
@@ -742,10 +742,10 @@ def is_covered(self, node_id: NodeIdLike) -> bool:
         node_id = self._resolve(node_id)
         return node_id in self._covers
 
-    def prune(self, node_id: NodeIdLike, keep_nodes: Iterable[NodeIdLike] = ()) -> list[NodeIdLike]:
+    def prune(self, node_id: NodeIdLike, keep_nodes: Iterable[NodeIdLike] = ()) -> list[int]:
         nodes = self.reachable_nodes(node_id)
         keep_nodes = [self._resolve(nid) for nid in keep_nodes]
-        pruned_nodes: list[NodeIdLike] = []
+        pruned_nodes: list[int] = []
```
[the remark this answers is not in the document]
PetarMax: Yes, because it actually is the tighter type by construction.

PetarMax:
> I'm also wondering if `APRProof` should just inherit from `KCFGExploration` instead? We're doing a lot of `self.kcfg_exploration.kcfg` for example, which if it inherits goes instead to `self.kcfg`. Is every `APRProof` also a `KCFGExploration`? It seems so to me....

Yes, I noticed that and am happy with that - I would like to know what @tothtamas28 has to say.

src/pyk/proof/reachability.py
PetarMax: Because the intended meaning of `_terminals` here is "terminal nodes obtained during execution". This excludes `target` - the check whether or not `target` is terminal is only interesting for whether or not the subsumption check is done on every obtained node or not.

ehildenb:
Tamas is on vacation, so we should probably not wait on his feedback on this one.

src/pyk/proof/reachability.py
PetarMax: We do not need to, but I think, given that it is a one-time check, that we should. It's a well-formedness check.

src/pyk/proof/reachability.py
PetarMax: Because it was already being treated like a set of integers, and also this minimises the number of calls to `_resolve`.

src/pyk/kcfg/explore.py
```diff
@@ -354,33 +355,52 @@ def _check_abstract(self, node: KCFG.Node, kcfg: KCFG) -> bool:
...
                     f'Found {len(branches)} branches using heuristic for node {node.id}: {shorten_hashes(node.id)}: {[self.kprint.pretty_print(bc) for bc in branches]}'
                 )
+                check_terminals(splits)
                 return
 
+        nodes_to_check: list[int] = []
+
         _LOGGER.info(f'Extending KCFG from node {self.id}: {shorten_hashes(node.id)}')
         depth, cterm, next_cterms, next_node_logs = self.cterm_execute(
             node.cterm,
```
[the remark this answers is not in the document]
PetarMax: I'm not sure how that would work, this is about checking the nodes that arise from the extension.

src/pyk/proof/reachability.py
PetarMax: Yes, I actually did that for a number of tests in either some of my previous PRs or on some of the branches that never got merged.

PetarMax: But we do have to get `apr_proof.proof_subdir` manually

src/pyk/proof/reachability.py
PetarMax: Perhaps it is, but this computation is surely more expensive than a dictionary lookup?

src/pyk/proof/reachability.py
PetarMax: Addressing this tomorrow...

PetarMax:
> Tamas is on vacation, so we should probably not wait on his feedback on this one.

Ok, so then I vote for the inheritance.

src/pyk/proof/reachability.py
PetarMax: Although now `APRProof` *is* `KCFGExploration`...

src/pyk/proof/reachability.py
```diff
@@ -204,19 +179,12 @@ def from_claim(
...
+        kcfg_dir = kwargs['proof_dir'] / claim.label / 'kcfg' if 'proof_dir' in kwargs else None
 
-        cfg, init_node, target_node = KCFG.from_claim(defn, claim, cfg_dir=kcfg_dir)
-        apr_proof.kcfg = cfg
-        apr_proof.init = init_node
-        apr_proof.target = target_node
-
-        return apr_proof
+        kcfg, init_node, target_node = KCFG.from_claim(defn, claim, cfg_dir=kcfg_dir)
+        return APRProof(claim.label, kcfg, [], init=init_node, target=target_node, logs=logs, **kwargs)
```
ehildenb, line 187: I _think_ the same can be achieved by explicitely listing `proof_dir` as a `kwarg`?

```
**kwargs: Any,
) -> APRProof:
    kcfg_dir = proof_dir / claim.label / 'kcfg' if proof_dir is not None else None
    ...
```

src/pyk/proof/reachability.py
ehildenb: I'm not sure that we should. If the user hands us an ill-formed KCFG or APRProof, we should probably run with it. Either way though, well-formedness of `terminal` nodes is a check that should happen in `KCFGExploration`, not here, because that's where that data and notion is defined.

src/pyk/kcfg/explore.py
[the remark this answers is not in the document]
ehildenb: Those nodes then become `leaf` nodes, and they will be checked the next time around. So we could either check them immediately upon adding them to the `KCFG` (instead of keeping a list to check at the end), or just wait for the next time that the KCFG is extended from a node, and check it.

We could also just check every node every iteration, and maintain a list of nodes we haven't checked.

Small preference, but this makes thi sextension algorithm more stateful, which means that changes in the middel of it need to remember to update this `needs_to_check` field correctly when appropriate.

src/pyk/proof/reachability.py
ehildenb: `terminal` is a generic property that could apply to any node, not about how the node came about. It's just a mapping from nodes to bool.

We in fact _need_ `target` node to be `terminal` if we want to disable `implication_every_block` automatically.

src/pyk/proof/reachability.py
ehildenb: It just means that it's another piece of state that needs to be maintained in sync with the underlying datastructures.

Do you have any profiling data showing tha tcomputing `KCFG.covers(...)` is a performance hotspot? I suspect it's not, the biggest hotspots in pyk are `kast_to_kore` and `kore_to_kast`, by orders of magnitude, everything else in pyk is tiny in comparison.

Do not optimize things that we don't have evidence that they need to be optimized. Make the code simple and direct and easy to maintain first.

src/pyk/kcfg/exploration.py
```diff
@@ -0,0 +1,93 @@
...
+        return self.kcfg.is_leaf(node_id) and not self.is_terminal(node_id) and not self.kcfg.is_stuck(node_id)
+
+    #
+    # Collectors
+    #
+
+    # Terminal node collector
+    @property
+    def terminals(self) -> list[KCFG.Node]:
+        return [node for node in self.kcfg.nodes if node.id in self._terminal]
```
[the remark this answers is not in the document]
PetarMax: Urgh, no...

src/pyk/proof/reachability.py
PetarMax: Probably, yes.

src/pyk/proof/reachability.py
PetarMax: I wouldn't know how to check that in `KCFGExploration`, we need a `KCFGExplore` for that, or a `KCFGSemantics`. I know how to check it inside an `APRProver`.

src/pyk/proof/reachability.py
PetarMax: The `is_terminal` function of a `KCFGSemantics` can be applied to any node, and it is applied to the `target` node when this check is needed. Specifically, in line 571 of `reachability.py`:
```
self._target_is_terminal = kcfg_explore._check_terminal(self._target)
```

I will check what would happen if `target` would be included.

src/pyk/proof/reachability.py
PetarMax: I do not agree with this approach to optimisation. I do not believe I need evidence of something being a performance hotspot in order to make a very natural memoization of it, which this is. 

Furthermore, I do not understand how this additional piece of state has to be maintained in sync with the underlying datastructures - it exists only within `APRProver` and is only affected by the `APRProver` constructor and `_check_subsume`.

Having said that, this PR has gone on for long enough, so I will just remove it.

src/pyk/proof/reachability.py
ehildenb: I still think we should not check this. This is a sanity check on data the user passes in, we can assume they either get it from a serialization on disk, or they know what they're asking for.

When constructing new nodes, I think it makes sense to do such sanity checks, but when handed the data-structure the user wants, I think usually you wouldn't.

src/pyk/proof/reachability.py
PetarMax: I'm including `target` now and have removed the "correctness" checks

src/pyk/kcfg/exploration.py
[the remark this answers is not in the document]
PetarMax: It's fixed now...

src/pyk/proof/reachability.py
ehildenb: It just means that instead of a (very fast) computation that will _always_ return the correct thing, we are storing something that a later change could forget to keep in sync. Another function gets added to `reachability.py` later (which is a frequent occurrence), does it need to now also modify the `_subsumption_checks` datastructure? What if that developer (and the reviewer) did not know about that, because they didn't read this conversation?

We don't have to reach very far back in the git history to see examples of where a change was made locally to a datastructure, and it forgot to keep some other piece of state in sync, or to update another procedure, and it caused problems. This has happened at least a handful of times across `APRProof` and `APRBMCProof` themselves. The less we create cases where it can happen, the easier our code is to maintain. Having data that an optimization _must_ be made is _exactly_ the criteria to use for deciding that it's appropriate to maintain such an extra piece of state.

Here is an example of such a discussion about targeted performance (though it was unproductive discussion at times): https://github.com/runtimeverification/pyk/issues/548. The summary is: everything pyk does is super fast except kast_to_kore and kore_to_kast, and loading K definitions.

We've been working with this (absolutely massive) codebase for long enough to know that introducing _any_ complexity that isn't needed is going to bite us later. So we have to actually _measure_ if performance optimizations are needed, and apply them only in a targeted focused way, because almost always performance optimizations lead to additional complexity (more state to track, for example). This is how the Haskell backend codebase became such a mess in the first place, was speculative optimization for cases that were not actually measured.

The reason I'm trying to drive this point home so much is not to be rude, but because it won't always be me reviewing your PRs, and this tenant of writing code that is maintainable is _essential_ to keep in mind for all of our sanity. Reducing statefulness of code is one of the simplest ways to keep it sane and easy to understand for later coders who join the effort. No code is throw-away, it must all be treated with the same care, assuming that someone new is going to have to read it and understand all of its intricacy. It's worth occasionally revisiting (and occasionally contributing to!) our Github Guidelines for Collaborative Coding: https://runtimeverification.com/blog/github-guidelines-for-collaborative-coding

src/pyk/kcfg/exploration.py
```diff
@@ -0,0 +1,93 @@
...
+        return self.kcfg.is_leaf(node_id) and not self.is_terminal(node_id) and not self.kcfg.is_stuck(node_id)
+
+    #
+    # Collectors
+    #
+
+    # Terminal node collector
+    @property
+    def terminal(self) -> list[KCFG.Node]:
+        return [node for node in self.kcfg.nodes if node.id in self._terminal]
```
ehildenb, line 41:
```suggestion
    # Terminal node collector
    @property
    def terminal(self) -> list[KCFG.Node]:
        return [node for node in self.kcfg.nodes if self.is_terminal(node.id)]
```

This allows us to change th edefinition of `is_terminal` in only one place, and have its effects propogate correctly everywhere.

src/pyk/proof/reachability.py
PetarMax: I do not feel like I'm doing anything that hasn't already been done. For bounded nodes in the `APRBMCProver`, there is the set `_checked_nodes: list[int]`, is that not effectively the same? In one of your other comments on terminal nodes, you said that one of the solutions would be to "check every node every iteration, and maintain a list of nodes we haven't checked." and I am doing that now. I am not understanding why keeping an extra dictionary with a clear purpose and encapsulated behaviour is different.

src/pyk/proof/reachability.py
PetarMax: This leads to the `test_fail_fast` tests failing fast, because they are assuming the meaning that I originally intended:
```
 >       assert len(proof._terminal) == 2
E       assert 3 == 2
E        +  where 3 = len({2, 6, 7})
E        +    where {2, 6, 7} = <pyk.proof.reachability.APRProof object at 0x7fe088a4fac0>._terminal
```
where 7 is the target node. 

Shall I adjust these tests?

src/pyk/kcfg/exploration.py
PetarMax: Yes, done

src/pyk/proof/reachability.py
PetarMax: @ehildenb I think this is the only remaining bit of the PR to be corrected - I can either correct the passing tests to include the target node as terminal or exclude target from terminal nodes, as was previously done. Let me know which you'd prefer.

ehildenb:
It appears there is an integration test failing. 

image

Locally, you can run:

```
make test-integration TEST_ARGS='-k test_fail_fast'
```

To try and reproduce that specific failure.

src/pyk/proof/reachability.py
ehildenb: I think adjust the test, yes.

src/pyk/kcfg/exploration.py
```diff
@@ -0,0 +1,93 @@
...
+    def __init__(self, kcfg: KCFG, terminal: Iterable[NodeIdLike] | None = None) -> None:
+        self.kcfg = kcfg
+        self._terminal = {kcfg._resolve(node_id) for node_id in terminal} if terminal is not None else set()
+
+    #
+    # Recognisers
+    #
+
+    # Terminal node recogniser
+    def is_terminal(self, node_id: NodeIdLike) -> bool:
```
ehildenb, line 27: FYI, for Python, it's good practice to use "docstrings" for comments: https://peps.python.org/pep-0257/

We don't do this now, but we probably will eventually, but these docstrings can be used to generate really nice documentation about the python code automatically. For example, see [the python pathlib docstring](https://github.com/python/cpython/blob/a071ecb4d13595f3580cf82061dcd7b39cd475c5/Lib/pathlib.py#L779), and generated documentation: https://docs.python.org/3/library/pathlib.html#module-pathlib

We haven't really been following that convention, but if you want to start improving our comments/documentation on code you touch, following this convention will yield better results long term.

src/pyk/proof/reachability.py
```diff
@@ -246,10 +217,10 @@ def path_constraints(self, final_node_id: NodeIdLike) -> KInner:
...
         dct = super().dict
         dct['type'] = 'APRProof'
-        dct['cfg'] = self.kcfg.to_dict()
+        dct['kcfg'] = self.kcfg.to_dict()
+        dct['terminal'] = sorted(self._terminal)
         dct['init'] = self.init
         dct['target'] = self.target
-        dct['terminal_nodes'] = list(self._terminal_nodes)
         dct['node_refutations'] = {node_id: proof.id for (node_id, proof) in self.node_refutations.items()}
         dct['circularity'] = self.circularity
```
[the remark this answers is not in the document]
PetarMax: We have to be explicit, yes, I will check throughout.

src/pyk/proof/reachability.py
PetarMax: Very well

src/pyk/kcfg/exploration.py
PetarMax: Yes, great, I will look into that.

src/pyk/proof/reachability.py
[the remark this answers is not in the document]
PetarMax: I am having some `mypy` problems when trying to deal with super, it thinks both this code:
```
dct = Proof.dict.fget(self)
```
and this code
```
dct = super(Proof, self).dict
```
are wrong with the respective error messages:
```
src/pyk/proof/reachability.py:218: error: "Callable[[Proof], dict[str, Any]]" has no attribute "fget"  [attr-defined]
Found 1 error in 1 file (checked 166 source files)
```
and
```
src/pyk/proof/reachability.py:218: error: "dict" undefined in superclass  [misc]
```
but these seem to be the ways to precisely specify from which superclass the property comes…

The former is a known issue in mypy, see [here](https://github.com/python/mypy/issues/8085), and the latter I don’t know about.

In any case, just having `super().dict` there will work because Python first checks `Proof`, which has the attribute, and if it didn't then it'd check `KCFGExploration`...

Issue #592, opened 2023-08-08
Support running subprocesses in GDB


ehildenb:
In an effort to remove bash runner script in KEVM and have pyk handle more functionality directly, it became apparent that one missing functionality is the ability to launch `gdb` on the resulting process.

- Migrating KEVM over to pyk instead fo bash: https://github.com/runtimeverification/evm-semantics/pull/1999
- Corresponding PR in pyk: https://github.com/runtimeverification/pyk/pull/584

It would be nice if `KRun.run_kore` and `KRun.run_kore_term` had ways to handle `debugger: bool` flag, which when set to `True` would launch GDB on the resulting call to `krun` instead of just invoking it directly. The naive way of doing this, by passing `--debugger` flag straight through to `krun` (by adding `--debugger` flag handling here: https://github.com/runtimeverification/pyk/blob/2d993285d981f13d96b78a398ec92f21c0601ec6/src/pyk/ktool/krun.py#L294), seemed to not work and just hang. My feeling is that what's happening is we need to forward stdin/stdout/stderr to the child process more carefully there, but not sure. There also are some GDB python libraries which maybe we need to look into.

PR #603, opened 2023-08-11
Support `--debugger` flag for krun

gtrepta:
fixes: #592


src/pyk/utils.py
```diff
@@ -400,6 +401,15 @@ def run_process(
...
 
     logger.info(f'Running: {command}')
+
+    if exec_process:
+        sys.stdout.flush()
+        sys.stderr.flush()
+        if type(args) is str:
+            args = tuple(shlex.split(args))
+        argstup = cast('tuple', args)
+        os.execvp(argstup[0], argstup)
```
ehildenb, line 411:
```suggestion
    if exec_process:
        sys.stdout.flush()
        sys.stderr.flush()
        if type(args) is str:
            argslist = shlex.split(args)
        else:
            argslist = list(args)
        os.execvp(argslist[0], argslist)
```

ehildenb (approved):

PR #604, opened 2023-08-12
Port model generation code from KEVM into APRFailureInfo

nwatson22:
I realized we still weren't using the new `APRFailureInfo` class to handle the printing of failure info in KEVM. And since that has been added @palinatolmach added counterexample/model printing to the failure info code in KEVM. Since this code is not KEVM-specific, I thought we should port that code over to the APRFailureInfo class to support switching over to using this class in KEVM.
Also refactors the existing fields of `APRFailureInfo` a bit to reduce large nested types.


nwatson22:
https://github.com/runtimeverification/evm-semantics/issues/1955 suggests to keep the model info as its own data structure, this current implementation ties it pretty tightly to the existing APRFailureInfo class rather than giving it its own class. I'm not sure if this is the best approach.

palinatolmach (approved):

ehildenb (approved):

PR #608, opened 2023-08-17
Store a digest of the proof in the proof directory

iFrostizz:
Advances towards fixing: https://github.com/runtimeverification/kontrol/issues/16
Adding the implementation and tests to write a digest of the proof once it's written.
The aim is to move one level of hashing by moving the proof id hash in a `digest` file located in the proof folder. This will allow us more versatility with the proof id to be able to use some more human readable version.
Also fixing a bug in `up_to_date` which was always returning `False` since it has not been updated with the new proof file system.


iFrostizz:
Closing for a more lightweight and backwards-compatible solution

PR #612, opened 2023-08-18
add some missing cases to ml_pred_to_bool

dwightguth:
This PR adds some missing cases to ml_pred_to_bool. These missing cases were triggering an exception in runtimeverification/evm-semantics#2017, which is now blocked on this PR.

It's a pretty simple change; equality is commutative so it should be straightforward to see why this is correct.


ehildenb:
Please add tests: https://github.com/runtimeverification/pyk/blob/e3511aa174020b7b106d32b74af3aa6150e91d77/src/tests/unit/kast/test_manip.py#L74

ehildenb (approved):

PR #613, opened 2023-08-21
Move the proof digest from the `Proof.id` to `Proof.proof_digest`

iFrostizz:
Adds a convenience field to proofs for https://github.com/runtimeverification/evm-semantics/pull/2026
`Proof.id` name choice is fully up to the code that implements pyk and used to be used as a digest, which didn't let us having a human readable id.
We now add a new `proof_digest` field in the `Proof` (and childrens) that is specifically dedicated to keeping track of the frontend chosen digest.


src/pyk/proof/proof.py
```diff
@@ -32,24 +32,21 @@ class Proof(ABC):
     _PROOF_TYPES: Final = {'APRProof', 'APRBMCProof', 'EqualityProof', 'RefutationProof'}
 
     id: str
+    proof_digest: str
```
ehildenb, line 35: Let's just call this `digest`, not `proof_digest`. We already are worknig witha  proof object, so we'll be accessing it as `proof.digest` anyway, so `proof.proof_digest` would be annoying.

src/pyk/proof/reachability.py
```diff
@@ -499,6 +515,7 @@ def from_dict(cls: type[APRBMCProof], dct: Mapping[str, Any], proof_dir: Path |
     def dict(self) -> dict[str, Any]:
         dct = super().dict
         dct['type'] = 'APRBMCProof'
+        dct['proof_digest'] = self.proof_digest
```
ehildenb, line 518: `super().dict` should already be taking care of setting the `'digest'` field, it sohuldn't be needed in subclasses.

ehildenb: Thsi probably needs to be changed in all the subclasses.

nwatson22:
I find it a bit weird to have the digest be its own field that we are passing in on creation, and saving and loading it to disk. I think of a digest as being able to be computed by the object based on its existing fields. Is this required for what you are doing with the keeping track of different versions of proofs with a counter? Maybe we could call it something other than digest if it is needed.

ehildenb:
I agree we should compute it as a `@cached_property`, so we're not recomputing it on subsequent calls, and it should be based on the contents, not passed in.

ehildenb:
Yo ucan implement it generically as:

```
from ..utils import hash_str


...
@cached_property
def digest(self) -> str:
    return hash_str(self.json)
```

src/pyk/proof/proof.py
iFrostizz: The reason I called it that way was because there is already a `digest` https://github.com/runtimeverification/pyk/blob/master/src/pyk/proof/proof.py#L95-L97
Though, it's not likely that we will need it outside, other than querying `up_to_date` so is that reasonable to make `_digest` private and return this new `proof.digest` as a getter ?

iFrostizz:
@ehildenb on kontrol, it would be good to be able to have multiple proofs to have the same digest, the reason is that we'll be able to archive proofs more easily. Also hashing the `proof.json` won't work because we won't be able to know if the user's project should create new proofs or not since they depend on themselves. This is how on Kontrol I was planning to pass the `foundry_digest`, see: https://github.com/runtimeverification/evm-semantics/pull/2026/files#diff-5591772bfb878d2c54f9c44714a5c1abea9ae5e83a7ebd2edd69bad35610b78cR654-R676

iFrostizz:
@nwatson22 Thanks for the suggestion,
It is probably required because we need to be able on the frontend to react on code changes so that we know when to create / resume a proof. I'm open to better ideas of that

ehildenb:
Then a digest isn't the correct way to control that, digests should be a hash of the contents, not a user-supplied identifier. User-supplied identifier isn't a digest, so that would be the `Proof.id` field, not the `Proof.digest` field.

iFrostizz:
@ehildenb what about creating a `index.json` in the `apr_proofs/` dir that shows the foundry_digest of each proof ?

```json
{
    "MyContract.testName:0": {
        "digest": "...",
        ...
    },
    "MyContract.testName:1": {
        "digest": "...",
        ...
    }
}
```

iFrostizz:
Isn't this already a proof digest https://github.com/runtimeverification/pyk/blob/master/src/pyk/proof/proof.py#L95-L97 ?

ehildenb:
Yes, I guess we don't need this PR then, just need to make changes to KEVM to better track proofs across revisions/changess.

PR #615, opened 2023-08-21
Support \left-assoc and \right-assoc on \and and \or

dwightguth:
This PR updates the kore parser to support using \or and \and as children of \left-assoc and \right-assoc.

I chose to also implement \and because the llvm backend parser already supports this, and I am about to implement it in the k frontend, which will mean that it will be completely supported for the purposes of the llvm and maude backends if I later need to make use of it.


dwightguth:
Dammit. I wrote tests and forgot to git add them.

tothtamas28:
Related: https://github.com/runtimeverification/haskell-backend/issues/3542

PR #616, opened 2023-08-22
Check for ml connectives in ml_pred_to_bool

anvacaru:
Addressing post-merge comments in #610


src/pyk/kast/manip.py
```diff
@@ -74,11 +74,21 @@ def bool_to_ml_pred(kast: KInner) -> KInner:
...
+            '#Top',
+            '#Bottom',
+            '#Implies',
+            '#Not',
+        }:
             return False
         if hasattr(kast_term, 'args'):
-            return all(_are_all_element_vars(arg) for arg in kast_term.args)
+            return all(_check_safe_term(arg) for arg in kast_term.args)
         return True
```
ehildenb, line 92: This should be implemented using a `collect` instead I think, instead of manually doing the recursion. For example, this won't recurse into `KSequence` as is, because it doesn't use `args`, it uses `items`. What I'd suggest is moving `def _check_safe_term(...)` to instead it's own dedicated function at the top of `manip.py` called `def is_term_like(...)`, and implement it as such:

```
def is_term_like(kast: KInner) -> bool:
    non_term_found = False
    def _is_term_like(_kast: KInner) -> None:
        match _kast:
            case KVariable(name, _):
                if name.startswith('@'):
                    non_term_found = True
            case KApply(KLabel(name, _), _):
                if name in {'#Equals', '#And', ...}:
                     non_term_found = True

    collect(_is_term_like, kast)
    return not non_term_found
```

Then we can add some unit tests for that direct routines as well.

anvacaru: @ehildenb  Should I also include `#Ceil` to the ml-connectives list?

ehildenb (approved):

ehildenb:
Excellent, thank you for the high quality change and tests.

PR #621, opened 2023-08-23
Handle `CTerm` bottom and top gracefully

iFrostizz:
closes https://github.com/runtimeverification/kontrol/issues/22
closes https://github.com/runtimeverification/pyk/issues/13

We now check when instantiating a `CTerm` `from_kast` if the pattern is bottom or top and shortcut from the split early to return either a `CTermBottom` or a `CTermTop`.


iFrostizz:
Actually shouldn't we also do the check in the `__init__()` ? I was scared that it would duplicate code

ehildenb:
This looks good, but we should adjust the various methods in `KCFGExplore` to also handle the new cases correctly. I think in particular, `KCFGExplore.cterm_simplify` should return a `CTerm` now, not a `KInner`: https://github.com/runtimeverification/pyk/blob/c292c4610183acec2ae44d8f68973ebf3cb4cd1e/src/pyk/kcfg/explore.py#L104

ehildenb:
This is also a big change, so you need to check that it will work with KEVM locally:

- Make a new branch of KEVM, link it to this branch of pyk.
- Try running the proof that was triggering the original bug with this branch of pyk.
- Add any fixes needed to KEVM to make it work with this branch.
- Link to that new branch here.

src/pyk/ktool/kprove.py
```diff
@@ -306,10 +305,12 @@ def prove_cterm(
             allow_zero_step=allow_zero_step,
             depth=depth,
         )
-        next_states = list(unique(var_map(ns) for ns in flatten_label('#Or', next_state) if not is_top(ns)))
+        next_states = list(unique(CTerm.from_kast(var_map(ns.kast)) for ns in next_state if not CTerm._is_top(ns.kast)))
```
ehildenb, line 308:
```suggestion
        next_states = list(unique(CSubst(var_map)(ns)) for ns in next_states)
```

I guess we can do something like this instead.

iFrostizz: Looks good, maybe that we should add this `__call__` method or is `apply` more explicit?

iFrostizz: (see https://github.com/runtimeverification/pyk/pull/621/commits/450e4e55609e283ed2bff416982e231ef273ca71)

iFrostizz: Actually it requires us to split the config for the `CSubst` so it would look uglier:

```python
next_states = set()
for ns in next_state:
    config, constraints = split_config_and_constraints(ns.kast)
    next_states.add(CSubst(var_map, flatten_label('#And', constraints)).apply(CTerm.from_kast(config)))
```

iFrostizz:
Trying to run the problematic code on this branch !

ehildenb (approved):

ehildenb:
Looks great, thank you for the clean focused change! Once you have confirmed that this will work with KEVM, fele free to add `automerge`.

iFrostizz:
@ehildenb I had to move away from the inheritance `CTerm` type matching because it wasn't playing well with instantiating the `CTerm` using the `__init__` function. Also added a `try_cell` because top and bottom `CTerm` don't have those, so the interfacing code can just try to fetch the cell if it's not `None`. And it's now fixed on kevm! Asking again for review please

iFrostizz:
https://github.com/runtimeverification/evm-semantics/compare/master...kevm-graceful-res#diff-8ef99db40b2010c0b478f54603fda2b277c332347b8945df0e83bb686a6eafc1

As a bottom or top node does not have any cell, in the kcfg it looks like so:
image

src/pyk/cterm.py
```diff
@@ -40,23 +40,43 @@ class CTerm:
...
         return CTerm(config, constraints)
 
+    @staticmethod
+    def cterm_top() -> CTerm:
+        return CTerm(mlTop(), ())
+
+    @staticmethod
+    def cterm_bottom() -> CTerm:
+        return CTerm(mlBottom(), ())
+
```
ehildenb, line 79:
```suggestion
    @staticmethod
    def top() -> CTerm:
        return CTerm(mlTop(), ())

    @staticmethod
    def bottom() -> CTerm:
        return CTerm(mlBottom(), ())

```

I guess this will let us write `CTerm.top()` instead of `CTerm.cterm_top()`, which feels redundant.

ehildenb (approved):

PR #624, opened 2023-08-25
Support for dispatching different commands to different servers

dwightguth:
Here we add support for a facade between the client and the actual server that it communicates with. Essentially, we can have multiple servers and pass different commands to different servers. For example, we might pass `execute` commands to one server and `check-implication` commands to another.

This is implemented via a `dispatch` dictionary that directs commands. that is passed to the constructor of KoreClient. The constructor takes a mapping from RPC endpoint names to a list of tuples containing the host, port, and TransportType used to communicate with that server.

If multiple servers are in the list, we send the same request to all the servers and return the first request that reports an error, or the final request if they all succeed. If there is nothing specified for an endpoint, we use the default `host` and `port` passed to KoreClient. For the purposes of preserving completeness, if a dispatch for a command does exist, the default server is not dispatched to; to dispatch to it you would have to add it to the list for that command.


dwightguth:
Yes. This is for the Maude backend, which is substantially faster than either Haskell backend, but only currently supports the execute and add-module endpoints.

nwatson22 (approved):

PR #627, opened 2023-08-30
Preserve path condition information in `anti_unify`

nwatson22:
@lucasmt suggested this as a way to preserve more information after a merge. Essentially, we take whatever path conditions are different between the two merged terms, and they are added to the conjunction corresponding to their respective term, so for:
```
    function testMinMax(uint256 x, uint256 y) external {
        uint256 min;
        uint256 max;

        if (x <= y) {
            min = x;
            max = y;
        } else {
            min = y;
            max = x;
        }

        // merge nodes here

        assertLe(min, max);
    }
```
we would get something like:
```
<config>
...
</config>
...
(min == x andBool max == y andBool x <=Int y) orBool (min == y andBool max == x andBool x >Int y )
```
where before it was:
```
<config>
...
</config>
...
(min == x andBool max == y) orBool (min == y andBool max == x )
```
It sort of makes sense to me but I need someone with more knowledge to confirm this works logically.


ehildenb:
What I would expect to see on this is something like:

```
(X <=Int Y impliesBool (MIN ==Int X andBool MAX ==Int Y)) orBool (Y <Int X impliesBool (MIN ==Int Y andBool MAX ==Int Y))
```

ehildenb (approved):

PR #628, opened 2023-08-31
Hotfix for the handling of bottom and top for the `CTerm` class

iFrostizz:
https://github.com/runtimeverification/pyk/pull/621 introduced a bug, it was falsely simplifying to `Bottom` because it was returning the disjunct of all states  when using the `prove_cterm` endpoint. This got uncovered as part of https://github.com/runtimeverification/evm-semantics/pull/2041

### DIFF
--- a/src/pyk/__main__.py
+++ b/src/pyk/__main__.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import json
 import logging
 import sys
 from argparse import ArgumentParser, FileType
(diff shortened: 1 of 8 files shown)


nwatson22 (changes requested):
I looked at this and it sounds like you were correct about what the problem is. We just need to make sure we aren't calling CTerm.from_kast on a term that could still possibly be a disjunction of terms, so run the `flatten_label('#Or'...` before it's converted to a CTerm.

nwatson22 (approved):

tothtamas28 (approved):

PR #638, opened 2023-09-06
Add parameter to `advance_proof` to split into subproofs after enough branches have been created

nwatson22:
In #524 @ehildenb suggested that we would want to have an option to split proofs off as subproofs if too much branching (too many pending nodes) happens in one proof. This would also be used to facilitate parallelization of a single proof, by allowing it to split itself into subproofs which could be executed independently.


src/pyk/proof/reachability.py
```diff
@@ -661,6 +671,28 @@ def advance_proof(
 
         self.proof.write_proof_data()
 
+    def delegate_to_subproof(self, node: KCFG.Node) -> None:
+        _LOGGER.info(f'Delegating node {node.id} to a new proof.')
+        subproof = self.construct_node_subproof(node)
+        self.proof.add_subproof(subproof)
+        self.proof.kcfg.add_stuck(node.id)
```
ehildenb, line 678: Hmmmm, adding `stuck` to this node will make it marked as failing, won't it? Maybe instead ew should treat it the same way we treat refutations, by tracking a list of `subproof_nodes` or something, and linknig that node to the subproof. That way we are correctly reporting pass/fail state of the overall proof.

ehildenb:
The main purpose of splitting to subproofs was to make parallelism easier. So we need to see if that will actually accomplish that goal. If not, do we need to do this? I guess we also get benefit of focusing on the subproblem, and being able to pass that around with less data?

nwatson22:
@ehildenb What do we need to see with regard to whether or not it will make parallelism easier?
I implemented this side-by-side with a branch in KEVM: https://github.com/runtimeverification/evm-semantics/compare/master...noah/branches-parallelization which uses the `max_branches` parameter and creates new threads for any new subproofs of a proof once it's finished. It still needs work but I'm hoping this makes more clear what my idea for this change was.

As you say, adding `delegate_to_subproof` would also allow the user to call it and focus in on subsections of the proof, but I'm less sure about the practical usefulness of this, so it's more of a secondary benefit.

ehildenb:
Ohhhh, nice. That looks good. Please discuss with Tamas in yoru meeting, see if it's an appropriate solution. Thanks for making an example with KEVM!

src/pyk/kcfg/show.py
```diff
@@ -282,6 +282,7 @@ def _sorted_init_nodes() -> tuple[list[KCFG.Node], list[KCFG.Node]]:
             return (init_nodes + init_leaf_nodes, remaining_nodes)
 
         init, _ = _sorted_init_nodes()
+
```
nwatson22, line 285:
```suggestion
```

src/pyk/kcfg/kcfg.py
```diff
@@ -459,6 +459,8 @@ def create_node(self, cterm: CTerm) -> Node:
         term = cterm.kast
         term = remove_source_attributes(term)
         cterm = CTerm.from_kast(term)
+        while self._node_id in self._nodes:
+            self._node_id += 1
```
tothtamas28, line 463: In which situation can `self._node_id in self._nodes` be `True`?

src/pyk/proof/reachability.py
```diff
@@ -635,13 +669,23 @@ def advance_proof(
...
         while self.proof.pending:
             self.proof.write_proof_data()
+
+            if max_branches is not None and len(self.proof.pending) > max_branches:
+                _LOGGER.info(
+                    f'Reached max_branches={max_branches} bound on number of pending nodes. Nodes {self.proof.pending} will be turned into subproofs.'
+                )
+                for pending_node in self.proof.pending:
+                    self.delegate_to_subproof(pending_node)
+                break
```
tothtamas28, line 687: Can `delegate_to_subproof` result in new pending nodes for `self.proof`?

src/pyk/kcfg/kcfg.py
nwatson22: I added this logic because `KCFG.add_node` can add a node of arbitrary ID to the KCFG, so if one gets added higher than the existing ones, `self._node_id` could be incremented to that level and it would end up overwriting this manually added node. However, I think the specific situation where this was happening that originally prompted me to add this is no longer happening, because instead of copying the 2 relevant nodes from the base proof as I was originally doing, it constructs new nodes, just with the same CTerms.

src/pyk/proof/reachability.py
nwatson22: It shouldn't create any new nodes in `self.proof`, no.

src/pyk/kcfg/kcfg.py
tothtamas28: > `self._node_id` could be incremented to that level and it would end up overwriting this manually added node

Good catch!

Is there a reason for `add_node` to be public though? I think users should just let the `KCFG` manage its IDs.

nwatson22: Good point. I've checked to make sure it wasn't being used externally and renamed it to reflect its private status, and removed the while loop.

nwatson22:
Blocked on https://github.com/runtimeverification/evm-semantics/pull/2094

PR #639, opened 2023-09-06
Add ksequence under all K arguments under all productions

### LINKED ISSUE #630: Incorrect kore emitted for KItem
Related to #565, pyk still does not emit correct kore for term `foo(X:KItem)` See the generated implication request from pyk:

```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "method": "implies",
  "params": {
    "antecedent": {
      "format": "KORE",
      "version": 1,
      "term": {
        "tag": "App",
        "name": "Lbl'-LT-'generatedTop'-GT-'",
        "sorts": [],
        "args": [
          {
            "tag": "App",
            "name": "Lbl'-LT-'k'-GT-'",
            "sorts": [],
            "args": [
              {
                "tag": "App",
                "name": "kseq",
                "sorts": [],
                "args": [
                  {
                    "tag": "App",
                    "name": "Lblfoo'LParUndsRParUnds'TEST-SYNTAX'Unds'KItem'Unds'K",
                    "sorts": [],
                    "args": [
                      {
                        "tag": "App",
                        "name": "inj",
                        "sorts": [
                          {
                            "tag": "SortApp",
                            "name": "SortKItem",
                            "args": []
                          },
                          {
                            "tag": "SortApp",
                            "name": "SortK",
                            "args": []
                          }
                        ],
                        "args": [
                          {
                            "tag": "EVar",
                            "name": "VarX",
                            "sort": {
                              "tag": "SortApp",
                              "name": "SortKItem",
                              "args": []
                            }
                          }
                        ]
                      }
                    ]
                  },
                  ...
                ]
              }
            ]
          },
          ...
        ]
      }
    },
    ...
  }
}
```

vs kore generated via `kprove`:

```
claim{} \implies{SortGeneratedTopCell{}} (
      \and{SortGeneratedTopCell{}} (
      \top{SortGeneratedTopCell{}}(), Lbl'-LT-'generatedTop'-GT-'{}(Lbl'-LT-'k'-GT-'{}(kseq{}(Lblfoo'LParUndsRParUnds'TEST-SYNTAX'Unds'KItem'Unds'K{}(kseq{}(VarX:SortKItem{},dotk{}())),Var'Unds'DotVar1:SortK{})),Lbl'-LT-'generatedCounter'-GT-'{}(Var'Unds'Gen0:SortInt{}))), weakAlwaysFinally{SortGeneratedTopCell{}} (
       \exists{SortGeneratedTopCell{}} (Var'QuesUnds'Gen1:SortInt{}, 
      \and{SortGeneratedTopCell{}} (
      Lbl'-LT-'generatedTop'-GT-'{}(Lbl'-LT-'k'-GT-'{}(kseq{}(Lblfail'LParRParUnds'TEST-SYNTAX'Unds'KItem{}(),Var'Unds'DotVar1:SortK{})),Lbl'-LT-'generatedCounter'-GT-'{}(Var'QuesUnds'Gen1:SortInt{})), \top{SortGeneratedTopCell{}}()))))
```

notice we have 

```
Lblfoo'LParUndsRParUnds'TEST-SYNTAX'Unds'KItem'Unds'K{}(inj{SortKItem{}, SortK{}}(VarX:SortKItem{}))
```
in pyk, vs

```
Lblfoo'LParUndsRParUnds'TEST-SYNTAX'Unds'KItem'Unds'K{}(kseq{}(VarX:SortKItem{},dotk{}()))
```
in kprove.

nwatson22:
Closes #630
Issue originally discovered in #565 and fixed in #569, but this turned out to be a special case of the more general case of kast-to-kore transformations on productions with `K`-sort arguments. (`==K` is the only built-in production like this, but the user can define their own. This PR handles the more general case.)
Future work should be done via #653 to find and eliminate any other potential problems with `kast_to_kore`.


src/pyk/kast/outer.py
```diff
@@ -1154,14 +1154,27 @@ def greatest_common_subsort(self, sort1: KSort, sort2: KSort) -> KSort | None:
...
+                    for i, sort in enumerate(prod.argument_sorts)
+                ],
+            )
+
+        #
+        #              if type(kast) is KApply and kast.label.name == '_==K_':
+        #                  return KApply('_==K_', [KSequence(arg) for arg in kast.args])
+        #              else:
+        #                  return kast
+
```
nwatson22, line 1176:
```suggestion
```

src/tests/integration/konvert/test_simple_proofs.py
```diff
@@ -644,6 +655,9 @@ def test_kast_to_kore(
         # When
         actual_kore = kast_to_kore(definition, kompiled_kore, kast, sort=sort)
 
+        print(kore)
+        print(actual_kore)
+
```
nwatson22, line 660:
```suggestion
```

src/pyk/kast/outer.py
```diff
@@ -1154,14 +1154,21 @@ def greatest_common_subsort(self, sort1: KSort, sort2: KSort) -> KSort | None:
...
+            prod = self.production_for_klabel(kast.label)
+            return KApply(
+                kast.label,
+                [
+                    KSequence(kast.args[i]) if sort.name == 'K' else kast.args[i]
+                    for i, sort in enumerate(prod.argument_sorts)
+                ],
+            )
+
+        return top_down(_add_ksequence_under_k_productions, kast)
```
ehildenb, line 1171: What if the sort of th ething underneath is already sort `K`? Should we not add the `KSequence` then? I don't think you're allowed to put a `K` directly into a `KSequence`.

Maybe we should add a test for this case, like `K1:K ==K K2:K`? What would `kast --output kore ...` give for this term? We should make sure to give the same thing probably.

nwatson22: It does add the KSequence for this example, but it passes the `kast_to_kore` test with the example you gave, because it looks like `ksequence_to_kore` handles this case later on. I change the `add_ksequence_under_k_productions` function to handle it anyway though, so that function would be correct by itself.

src/pyk/kast/outer.py
```diff
@@ -1154,14 +1154,30 @@ def greatest_common_subsort(self, sort1: KSort, sort2: KSort) -> KSort | None:
...
+    def add_ksequence_under_k_productions(self, kast: KInner) -> KInner:
+        def _add_ksequence_under_k_productions(kast: KInner) -> KInner:
+            def _is_k(kast: KInner) -> bool:
+                if type(kast) is not KVariable:
+                    return False
+                if kast.sort is None:
+                    return True
+                if kast.sort.name == 'K':
+                    return True
+                return False
```
ehildenb, line 1166: Can we just use `self.sort(kast)` instead? So we could say:

```suggestion
            def _is_k(kast: KInner) -> bool:
                return self.sort(kast) == KSort('K')
```

In which case, we don't need the helper probably?

nwatson22: Changed to use `self.sort(kast.args[i]) == KSort('K')`

ehildenb:
So, it seems to me that basically what's happening here is that every time we have an injection from sort `KItem` into sort `K`, instead we're using a `KSequence` for that rather than an explicit injection?

If it's the case, maybe this could be implemented as a pass _after_ `add_sort_injection`, which terns `inj(SortK{},SortKItem{})(...)` into `kseq(...,dotk())` directly at the Kore level? Is there a benefit to that approach?

This seems to work though, so I guess it doesn't really matter that much. But it seems to me that using something of sort `KItem` where a `K` is expected is valid KAST, but not valid in Kore for the same reasons we need to add injections in Kore. So we could just leave it alone at the KAST level, and do this transformation of injections into kseq at the Kore level instead.

src/pyk/kore/kompiled.py
```diff
@@ -105,6 +105,31 @@ def meet_all_sorts(self, sorts: Iterable[Sort]) -> Sort:
         unit: Sort = SortApp('SortK')
         return reduce(self.meet_sorts, sorts, unit)
 
+    def replace_k_injections_with_ksequence(self, pattern: Pattern) -> Pattern:
+        print(pattern)
+        print('')
+
```
nwatson22, line 111:
```suggestion
```

src/pyk/konvert.py
```diff
@@ -87,12 +87,14 @@ def kast_to_kore(
 ) -> Pattern:
     if sort is None:
         sort = K
-    kast = kast_defn.add_ksequence_under_kequal(kast)
+    #      kast = kast_defn.add_ksequence_under_k_productions(kast)
```
nwatson22, line 90:
```suggestion
```

nwatson22:
> So, it seems to me that basically what's happening here is that every time we have an injection from sort `KItem` into sort `K`, instead we're using a `KSequence` for that rather than an explicit injection?
> 
> If it's the case, maybe this could be implemented as a pass _after_ `add_sort_injection`, which terns `inj(SortK{},SortKItem{})(...)` into `kseq(...,dotk())` directly at the Kore level? Is there a benefit to that approach?
> 
> This seems to work though, so I guess it doesn't really matter that much. But it seems to me that using something of sort `KItem` where a `K` is expected is valid KAST, but not valid in Kore for the same reasons we need to add injections in Kore. So we could just leave it alone at the KAST level, and do this transformation of injections into kseq at the Kore level instead.

@ehildenb I agree that makes more sense. I've added a function that does this `replace_k_injections_with_ksequence(self, pattern: Pattern) -> Pattern`. It handles 2 separate cases where the `kseq` is inserted, one where the item under the kseq needs to be injected into `KItem`, and one where it is already `KItem`.

src/pyk/kast/outer.py
```diff
@@ -1154,14 +1154,23 @@ def greatest_common_subsort(self, sort1: KSort, sort2: KSort) -> KSort | None:
...
+                kast.label,
+                [
+                    KSequence(kast.args[i])
+                    if sort.name == 'K' and not self.sort(kast.args[i]) == KSort('K')
+                    else kast.args[i]
+                    for i, sort in enumerate(prod.argument_sorts)
+                ],
+            )
+
+        return top_down(_add_ksequence_under_k_productions, kast)
```
ehildenb, line 1173: Is this method used anymore? Maybe it can be deleted then?

src/pyk/kore/kompiled.py
```diff
@@ -105,6 +105,32 @@ def meet_all_sorts(self, sorts: Iterable[Sort]) -> Sort:
         unit: Sort = SortApp('SortK')
         return reduce(self.meet_sorts, sorts, unit)
 
+    def replace_k_injections_with_ksequence(self, pattern: Pattern) -> Pattern:
+        def _replace_k_injections(pattern: Pattern) -> Pattern:
+            if (
+                type(pattern) is App
+                and pattern.symbol == 'inj'
+                and pattern.sorts[0] != SortApp('SortK')
+                and pattern.sorts[1] == SortApp('SortK')
```
ehildenb, line 114: This check is confusing to me? Should we instead be saying:

```suggestion
                and pattern.sorts[0] == SortApp('SortKItem')
                and pattern.sorts[1] == SortApp('SortK')
```

Basically very specifically calling out injections from `KItem -> K`, and replacing them?

Or is the thinking here that anything injected into K must go through `KItem` first? 

Maybe we need to ask @dwightguth the design decision behind the sorts `K` and `KItem`. Whenever we have something of sort `B` lower than `K` in place where a `K` is expected, are we expected that we have a `kseq` at the top, then an injection from `B -> KItem` under that?

If so, I guess I would propose doing this pass as:

- A pass that turns injections like `B -> K` into a double injection `B -> KItem -> K` (if `B != KItem` and `B != K`), and then
- A pass that takes injections `KItem -> K` and turns them into `kseq(...)` instead.

dwightguth: Okay, I can give you some insight into how KORE is generated by the Java frontend.

The frontend considers KItem a subsort of K, but the backends do not. From the perspective of the backend, `K` as a sort has two constructors and one function. The constructors are `dotk{}`, the empty K sequence, and `kseq{}`, which is the `cons` constructor for the list, and has the following signature:

```
symbol kseq(KItem, K) : K
```

There is also the following symbol:

```
symbol append(K, K) : K
```

This is a fairly standard cons-list-append function.

The frontend converts a list of terms into the backend representation by using `kseq` for each term that is a subsort of KItem and `append` if the term on the left is sort K. This then ultimately gets normalized by the implementation of `append` into a list of `KItem`.

It is never correct to have an injection directly into `K`. The behavior in this case is to create a singleton K sequence whose element is an injection into KItem (unless the term itself is of sort KItem, in which case no injection is needed).

dwightguth: The intuition of the K sort is that it's literally just a list of KItems, so it stands to reason that you would need a singleton list if you have a KItem in a place where a K is expected. I don't anticipate that that will change unless the behavior of the K sort is fundamentally changed, which seems unlikely to me since that would break a lot of assumptions semantics are making. If you don't want to support 0 or 2+ element lists, why are you using sort K? And if you do, you need the explicit list representation even if the list is currently of length 1. So I don't expect this behavior will change anytime soon.

ehildenb: Hmmmm, OK. In that case, what I would propose is this:

- Run the pass that inserts `kseq` _before_ adding sort injections. Basically look for places where you have a production that expects a `K` under it (like `_==K_`), but the argument is not already of sort `K`, then you add the `KSequence` under that.
- Then run the sort injection pass (which already will correctly infer `KItem` for that position).

This is basically what we were doing before, we were inserting `KSequence` at the KAST level, and letting sort injections happen at Kore level. So I'd say lets go back to that solution. Thanks for the insight @dwightguth !

dwightguth: Even if you do this, you still need to use `append` instead of `kseq` if a term in a KSequence other than the last is of sort K when converting from KAST to KORE. Not sure if you're already doing that.

dwightguth: You also need to make sure that you flatten any K sequences that are nested /after/ splitting the left and right hand side of a rule if you are creating a new rule or claim.

ehildenb: Hmmmm, right now we are not doing any splitting of rules and such, so I don't think that is the concern (this is for individual terms). But we are not emitting `append(...)` in any cases, so that's probably something we need to look at.

I'm wondering how we would even really test that though....

I guess the test which has:

```
KSequence([KVariable('X', 'K'), KVariable('Y', 'K')])
```

Should be emitting something like:

```
append(X:K, Y:K)
```

rather than:

```
kseq(X:K, kseq(Y:K, dotk))
```

And the test that has:

```
KSequence([KVariable('X'), KVariable('Y')])
```

Should be emitting:

```
kseq(X:KItem, Y:K)
```

Does that sound right?

nwatson22: Is it important that `kast_to_kore` always produce the same KAST as `kast` (k binary)? If so, could we just build this requirement so it tries both and compares them? I have done some experimenting with this and it seems like there may be other discrepancies. Not sure if they are necessarily important though. For example, `kast_to_kore` outputs
```
App(symbol='Lblbarholder2', sorts=(), args=(App(symbol='inj', sorts=(SortApp(name='SortBaz', sorts=()), SortApp(name='SortBar', sorts=())), args=(EVar(name='VarX', sort=SortApp(name='SortBaz', sorts=())),)), EVar(name='VarX', sort=SortApp(name='SortBaz', sorts=()))))
```
while using `KPrint.kast_to_kore` modified to always use the `kast` binary outputs:
```
App(symbol='Lblbarholder2', sorts=(), args=(App(symbol='inj', sorts=(SortApp(name='SortBaz', sorts=()), SortApp(name='SortBar', sorts=())), args=(EVar(name='VarX', sort=SortApp(name='SortBaz', sorts=())),)), App(symbol='inj', sorts=(SortApp(name='SortBar', sorts=()), SortApp(name='SortBaz', sorts=())), args=(EVar(name='VarX', sort=SortApp(name='SortBar', sorts=())),))))
```
for test `variable-with-multi-sort`.

ehildenb: Hmmmm, adding that as a test seems like a good idea! I would first just get this PR finished up (with the ksequence changes, basically going back to how you were doing it at first), then we can make a separate PR which tests all teh `kast_to_kore` conversion data with both techniques and compares. We can just skip tests with `pytest.skip()` for ones that don't work at the moment, and make an issue about it.

src/pyk/kast/outer.py
```diff
@@ -1154,14 +1154,23 @@ def greatest_common_subsort(self, sort1: KSort, sort2: KSort) -> KSort | None:
...
-        return top_down(_add_ksequence_under_kequal, kast)
+            prod = self.production_for_klabel(kast.label)
+            return KApply(
+                kast.label,
+                [
+                    KSequence(kast.args[i])
+                    if sort.name == 'K' and not self.sort(kast.args[i]) == KSort('K')
+                    else kast.args[i]
+                    for i, sort in enumerate(prod.argument_sorts)
+                ],
```
ehildenb, line 1170:
```suggestion
                [
                    KSequence(arg)
                    if sort.name == 'K' and not self.sort(arg) == KSort('K')
                    else arg
                    for arg, sort in zip(kast.args, prod.argument_sorts, strict=True)
                ],
```

ehildenb (approved):

PR #641, opened 2023-09-07
Change `kbuild` to produce targets locally

### LINKED ISSUE #586: `kbuild` should put definitions in a local folder by default
Currently artifacts built by `kbuild` are placed in a folder structure like this:

```
~/.kbuild
├── hoon-semantics
│   └── d2f5ce0
│       ├── include
│       │   └── hoon-semantics
│       └── target
│           └── v5.6.110-0-gf2f25ce798
│               ├── haskell
│               └── llvm
...
```

The idea is that the `~/.kbuild` folder serves as a user-local repository for kompiled definitions, arranged by source of installation (`d2f5ce0`) and K version (`v5.6.110-0-gf2f25ce798`).

However, for semantics projects, this seems to be more of an inconvenience than a feature. There, `.kbuild` is only used to produce the definitions but they are distributed in some other way. But due to the complex folder structure, the artiacts can only be accessed through `kbuild which`.

So by default `kbuild` just should output targets in a local folder, e.g.:

```
./kdist
└── hoon-semantics
    ├── include
    │   └── hoon-semantics
    ├── haskell
    └── llvm
```

Maybe the repository thing can even be dropped altogether.

tothtamas28:
Fixes #586

The tool now produces the distribution under a directory `kdist` in the project folder by default, or at a location specified by the user (using `--output`).

Further changes:
* Since there is no global repository, hashes have also been dropped from the distribution hierarchy
* Command `kbuild clean` has been removed


tothtamas28:
Current distribution layout:

```
kdist
├── evm-semantics
│   ├── include
│   │   └── evm-semantics
│   │       └── lemmas
│   └── target
│       └── v6.0.74-0-g6cbfd9428c
│           └── haskell
└── plugin
    ├── include
    │   └── plugin
    └── resource
        ├── include
        │   └── cryptopp
        ├── lib
        └── src
```

Maybe this should be further reduced to

```
kdist
└── haskell
```

This requires two things:
* Rekompile on K version change: https://github.com/runtimeverification/pyk/issues/586#issuecomment-1671854395
* Do not copy source files into the distribution. Instead, do kompilation in `/tmp`, `$XDG_CACHE_HOME`, etc.

ehildenb:
I think having source files in the distribution is a feature actually. Eventually, we want other semantics to be able to improt this semantics directly without having to clone the soruces, just going through the distribution packages.

tothtamas28:
> I think having source files in the distribution is a feature actually. Eventually, we want other semantics to be able to improt this semantics directly without having to clone the soruces, just going through the distribution packages.

The sources (similarly to the kompiled resources) could be distributed with the Python package, e.g.

```
evm-semantics
├── pyproject.toml
└── src
    └── evm-semantics
        ├── kdist
        └── ksrc
            └── kbuild.toml
```

In this case, the projects should probably define some simple hooks for `kbuild` though (`kbuild.py`?), I'm not sure at this point.

tothtamas28:
> the projects should probably define some simple hooks for `kbuild` though

Actually, this is not necessary. We can add a feature that allows dependencies to be (installed) Python modules.

For example, given a package for

```
plugin-semantics
├── pyproject.toml
└── src
    └── plugin-semantics
        └── ksrc
            └── kbuild.toml
```

is installed, `evm-semantics` could just define

```toml
[dependencies]
plugin = { module = "plugin-semantics.ksrc" }
```

and then `kbuild` can easily resolve the actual file system path. This enables distributing `kbuild` projects as Python packages.

radumereuta:
Why did you remove `kbuild clean`?
I still find that useful.

tothtamas28:
> Why did you remove `kbuild clean`? I still find that useful.

Since the folder to clean comes from the user now, the command no longer has an added value:

```
kbuild clean kdist
```

is just a less flexible version of

```
rm -rf kdist
```

radumereuta:
> is just a less flexible version of
> 
> ```
> rm -rf kdist
> ```

Great. I want that.
Think of the fact that someone wants to share the project; they don't care about how it is structured.
They just know that they have `make clean`, `mvn clean` and they all do the same thing. Clean the generated files in order to build everything from scratch.
It abstracts some information that I don't really need to know.
It also makes dependent tools more flexible. For example, the VSCode extension can just call `kbuild clean` and everything is going to happen externally. W/o that option I will have to look into the toml file, find the output directory and call `rm`.

tothtamas28:
> W/o that option I will have to look into the toml file, find the output directory and call `rm`.

The output directory is not specified in  the `.toml` file, it is up to the user to give that input when invoking the `kbuild` command (or, if it is not provided, then it's `kdist` in the project root).

radumereuta:
> The output directory is not specified in the `.toml` file, it is up to the user to give that input when invoking the `kbuild` command (or, if it is not provided, then it's `kdist` in the project root).

I see. Then all the more reasons to have a clean command.
This should function like the rest of the build tools (make, mvn). The clean command is an important part.

tothtamas28:
> I see. Then all the more reasons to have a clean command. This should function like the rest of the build tools (make, mvn). The clean command is an important part.

In `make` it's up to the user do define a `clean` rule. In `mvn` there is a global cache. `kbuild` does not have a global resource like that right now (that's the point of this PR), so a `clean` command could only look like:

```
$ kbuild clean           # removes kdist from the project root
$ kbuild clean some-dir  # removes some-dir
```

If you find this feature useful, I can add it, but ultimately, it is still up to the user to manage the artifacts produced by `kbuild` based on their input.

radumereuta:
Then, the output dir should be specified in the toml file.
Why would I want to specify it each time I want to build my project?
There should be a default. `kdist` is fine. `kbuild clean` should remove that dir.
You should be able to override it in the toml file. `kbuild clean` should read that path and clean it.
This is also important because of `.gitignore`. You set it once and forget.

If you want to do a one time `kbuild --output <dir>` then sure, you can do that. But it's going to show up in gid diff, and the default clean won't know about it.

ehildenb:
I agree that specifying the output directory (but having a default at `kdist`) is a good thing to have, and then having `kbuild clean ...` handle it automatically is good too. Eventually, we probably will have caches and such, and it would be nice if users didn't have to learn a new command at that time. Just my 2c.

ehildenb (approved):

PR #646, opened 2023-09-08
Kompile definitions in a temporary build environment

tothtamas28:
Extracts a class `KBuildEnv` that encapsulates the logic of the build environment. Source files are no longer copied into the
`kbuild` distribution.


src/pyk/kbuild/project.py
```diff
@@ -232,6 +232,16 @@ def resource_file_names(self) -> dict[str, list[str]]:
...
+    def all_files(self) -> list[Path]:
+        res: list[Path] = []
+        for sub_project in self.sub_projects:
+            res.append(sub_project.project_file)
+            res.extend(sub_project.source_files)
+            for resource_name in sub_project.resources:
+                res.extend(sub_project.resource_files[resource_name])
+        return res
+
     def get_target(self, target_name: str) -> Target:
```
ehildenb, line 245: Should this also be including things like `self.resource_files`, `self.source_files`, etc...?

ehildenb (approved):

src/pyk/kbuild/project.py
tothtamas28: Yes, `project.sub_projects` includes `project`.

PR #652, opened 2023-09-13
Make branch extraction more intelligent

### LINKED ISSUE #632: Branch handling improvements
I have observed one problem and one detail with the branch extraction mechanism in `pyk` while working on a recent engagement:

**Problem**: the results of the `extract_branches` function (as used [here](https://github.com/runtimeverification/pyk/blob/a18d76df9419194acf1f9d159d8fa697e0471bd9/src/pyk/kcfg/explore.py#L376)) are not simplified or checked for satisfiability. This means that by using the mechanism we could end up with more complex terms in the path constraint or explore vacuous branches. My experience is that simplifications being applied is essential for tractability of our verification. For this reason, I am currently commenting out the `extract_branches` of KEVM while working with client code. I see two solutions:

1. Let the backend handle the branching fully - this will ensure the simplifications are done
2. Simplify manually in `pyk` (which does mean calling the backend) while checking feasibility

**Detail**: It is not always the case that the path constraints of next states syntactically extend the path constraint of the initial state. For example, if we learn that two symbolic variables are equal, then the backend will perform the substitution. This means that [this check](https://github.com/runtimeverification/pyk/blob/a18d76df9419194acf1f9d159d8fa697e0471bd9/src/pyk/kcfg/explore.py#L419) will not always work (as @lucasmt noticed). I have also [flagged this substitution behaviour](https://github.com/runtimeverification/haskell-backend/issues/3648) to the Haskell backend team so that it can be confirmed that it is desired.

nwatson22:
Closes #632.

Before introducing a branch, simplify each of the would-be new nodes and see if they simplify to bottom
If there are more than 1 non-bottom node, add all these simplified nodes to the CFG, creating a split/branch. If not, simply continue straight-line execution.

This should reduce eliminate some infeasible branching by simplifying earlier in the process.

Also should eliminate repeated branching encountered in https://github.com/runtimeverification/evm-semantics/pull/2065 when creating new proofs from branches.


src/pyk/kcfg/explore.py
```diff
@@ -377,14 +377,24 @@ def extend(
...
+            if not cterm.is_bottom:
+                branch_conditions.append(constraint)
+                branch_cterms.append(cterm)
+
+        if len(branch_cterms) > 1 and branches:
+            kcfg.branch(node.id, zip(branch_cterms, branch_conditions, strict=True))
+            _LOGGER.info(
+                f'Found {len(branches)} branches using heuristic for node {node.id}: {shorten_hashes(node.id)}: {[self.kprint.pretty_print(bc) for bc in branches]}'
+            )
+            return
```
ehildenb, line 398:
```suggestion
        _branches = self.kcfg_semantics.extract_branches(node.cterm)
        branches = []
        for constraint in _branches:
            cterm = node.cterm.add_constraint(constraint)
            cterm, _ = self.cterm_simplify(cterm)
            if not cterm.is_bottom:
                branches.append(constraint)
        if len(branches) > 1:
            kcfg.split_on_constraints(node.id, branches)
            _LOGGER.info(
                f'Found {len(branches)} branches using heuristic for node {node.id}: {shorten_hashes(node.id)}: {[self.kprint.pretty_print(bc) for bc in branches]}'
            )
            return

```

Does this work? Seems a bit simpler and doesn't require creating a zip list or modifynig `KCFG`.

ehildenb: Then we can get rid of the modification down lower too.

src/pyk/kcfg/explore.py
```diff
@@ -377,14 +377,24 @@ def extend(
         if self._check_abstract(node, kcfg):
             return
 
-        if not kcfg.splits(target_id=node.id):
```
PetarMax, line 380: This check is no longer relevant?

src/pyk/kcfg/explore.py
```diff
@@ -377,14 +377,24 @@ def extend(
...
+        branch_cterms = []
+
+        for constraint in branches:
+            cterm = node.cterm.add_constraint(constraint)
+            cterm, _ = self.cterm_simplify(cterm)
+            if not cterm.is_bottom:
+                branch_conditions.append(constraint)
+                branch_cterms.append(cterm)
+
+        if len(branch_cterms) > 1 and branches:
```
PetarMax, line 392: Can it be that `len(branch_cterms) > 1` and `branches` is empty?

src/pyk/kcfg/explore.py
```diff
@@ -377,14 +377,24 @@ def extend(
...
-                    f'Found {len(branches)} branches using heuristic for node {node.id}: {shorten_hashes(node.id)}: {[self.kprint.pretty_print(bc) for bc in branches]}'
-                )
-                return
+        branches = self.kcfg_semantics.extract_branches(node.cterm)
+
+        branch_conditions = []
+        branch_cterms = []
+
+        for constraint in branches:
+            cterm = node.cterm.add_constraint(constraint)
```
PetarMax, line 386: Do we need to carry the configuration around, could we not use just the `constraints` part of the `CTerm` in question? The configuration could be quite complex...

src/pyk/kcfg/explore.py
```diff
@@ -377,14 +377,24 @@ def extend(
...
+            cterm, _ = self.cterm_simplify(cterm)
```
PetarMax, line 387: ...and here use `kast_simplify`?

PetarMax: and maybe put the constraint back into the `cterm` if that's what's easier for `branch` later on...

src/pyk/kcfg/explore.py
nwatson22: That would work, that's basically what I did at first. It's just that the new nodes wouldn't have the simplification applied. Maybe we don't care if it immediately gets applied, just that the nodes don't get created if they *would* simplify to bottom?

src/pyk/kcfg/explore.py
nwatson22: Yes, as I understand, it was the check preventing us from branching over and over again. But by simplifying before branching when it tries to branch a second time without having made an execution step, one of the branches will now simplify to bottom and it won't branch.

src/pyk/kcfg/explore.py
nwatson22: It no longer checks `branches` now.

src/pyk/kcfg/explore.py
nwatson22: You're right, I changed it here so we can just use the constraints and use `kast_simplify`.

src/pyk/kcfg/explore.py
nwatson22: Using `kast_simplify`. I removed for now the adding of the already simplified term as the new branch nodes.

src/pyk/kcfg/explore.py
nwatson22: I've changed it to the simpler way for now.

src/pyk/kcfg/explore.py
ehildenb: Why `kast_simplify` instead of `cterm_simplify`?

ehildenb: I guess the argument is performance. Have we checked on KEVM that it solves the problems we had there?

nwatson22: I don't myself know if it makes an appreciable difference to performance, but if it doesn't make any difference to have the configuration in there, it's just simpler.
I have checked and it does solve my issue with repeated branching. I'm not sure if there was a specific problem that prompted #632 or just a general observation. @PetarMax, do you have a KEVM example where this was causing problems that I could test this against?

ehildenb (approved):

src/pyk/kcfg/explore.py
PetarMax: What do you mean by 'over and over again', when can that happen? As part of looping code?

src/pyk/kcfg/explore.py
```diff
@@ -377,14 +377,19 @@ def extend(
...
+            kast = mlAnd(list(node.cterm.constraints) + [constraint])
+            kast, _ = self.kast_simplify(kast)
+            if not CTerm._is_bottom(kast):
+                branches.append(constraint)
+        if len(branches) > 1:
+            kcfg.split_on_constraints(node.id, branches)
+            _LOGGER.info(
+                f'Found {len(branches)} branches using heuristic for node {node.id}: {shorten_hashes(node.id)}: {[self.kprint.pretty_print(bc) for bc in branches]}'
+            )
+            return
```
PetarMax, line 392: In this case, will the unsimplified constraint be added to the configuration or will the Haskell backend simplify it? The problem that we had was that the booster was not able to simplify a given constraint. Perhaps just checking for `#Bottom` will be enough.

I will try to run this on an eppropriate test from the engagement in question.

ehildenb: The unsimplified contsraint will be added, but that should be enough to make it only take the one branch on subsequent execute call.

PR #667, opened 2023-09-28
Re-enable LLVM backend Unicode tests

Scott-Guest:
The LLVM backend no longer conflates Bytes and Unicode strings, so we can re-enable the string integration tests


tothtamas28 (approved):
Nice!

PR #670, opened 2023-09-29
Accept multiary input syntax for `\and` and `\or`

tothtamas28:
* Generalize textual syntax: `\and{S}(a{}(), b{}(), c{}())`
* Add alternative JSON syntax: `{"tag": "And", "patterns": [...], ...}`

Input is still parsed into a binary form (left-associative). The new syntax does not propagate to the output.


dwightguth:
This seems like it should be sufficient for the purposes of compatibility with a backend that emits this notation. I note you implemented the version with 2+ arguments. See my comment on the issue.

tothtamas28:
> I note you implemented the version with 2+ arguments.

I'll adjust in an upcoming PR that changes internal representation and allows emitting multiary output (but for that we need backend support first). Would that work for you?

dwightguth (approved):

PR #671, opened 2023-09-29
Decouple next-state computation from `KCFG` updates

tothtamas28:
Extracts three steps from `KCFGExplore.extend`:
* `check_extendable`: check if extending the node makes sense
* `extend_cterm`: compute the next states for a `CTerm`
* `extend_kcfg`: apply the state update to a `KCFG`

This enables parallelization of next-state computation with the current `KCFG` data structure: submit multiple `extend_cterm` calls to a thread pool, and upon a result call `extend_kcfg` from the main thread.


src/pyk/kcfg/explore.py
```diff
@@ -402,63 +410,85 @@ def extend(
...
-                kcfg.split_on_constraints(node.id, branches)
+            case Step(cterm, depth, next_node_logs, cut):
+                next_node = kcfg.create_node(cterm)
+                logs[next_node.id] = next_node_logs
+                kcfg.create_edge(node.id, next_node.id, depth)
+                cut_str = 'cut-rule ' if cut else ''
                 _LOGGER.info(
-                    f'Found {len(branches)} branches for node {self.id}: {shorten_hashes(node.id)}: {[self.kprint.pretty_print(bc) for bc in branches]}'
+                    f'Found {cut_str}basic block at depth {depth} for {self.id}: {shorten_hashes((node.id, next_node.id))}.'
                 )
```
ehildenb, line 473: I wonder if the logging message can be made a part of the `ExtendResult` class, so that we just say something like `_LOGGER.info(f'Extend result for {self.id}: {result.log_message}')` or something.

src/pyk/kcfg/explore.py
```diff
@@ -402,63 +410,85 @@ def extend(
...
-        elif len(next_cterms) == 1:
-            next_node = kcfg.create_node(next_cterms[0])
-            logs[next_node.id] = next_node_logs
-            kcfg.create_edge(node.id, next_node.id, 1)
-            _LOGGER.info(
-                f'Inserted cut-rule basic block at depth 1 for {self.id}: {shorten_hashes((node.id, next_node.id))}'
-            )
+            case Abstract(cterm):
+                new_node = kcfg.create_node(cterm)
+                kcfg.create_cover(node.id, new_node.id)
```
ehildenb, line 463: I guess we're missing a log message here, which would be fixed by making it part of the `ExtendResult` class and just emitting the log event at the beginning of the `extend_kcfg(...)` method.

ehildenb:
I'm wondering about locking? When we do `check_extendable`, should we acquire a lock perhaps? Or is that something to handle in the user/outside application?

src/pyk/kcfg/explore.py
tothtamas28: This is difficult as these log messages have all sorts of dependencies, e.g.:

```python3
f'Found ... for {self.id}: {node.id} -> {[self.kprint.pretty_print(bc) for bc in constraints]}'
```

src/pyk/kcfg/explore.py
tothtamas28: Organized logging a bit.

tothtamas28:
> I'm wondering about locking? When we do `check_extendable`, should we acquire a lock perhaps? Or is that something to handle in the user/outside application?

I don't think that's necessary, as long as the main thread does not try to extend a frontier twice.

tothtamas28:
Tested with Foundry proofs.

ehildenb (approved):

PR #676, opened 2023-10-06
Extend `pyk` CLI with Kore RPC requst/response pretty-printer

geo2a:
This PR extends the `pyk` CLI with `pyk rpc-print` --- a command that allows pretty-printing of Kore RPC request and responses.

Current capabilities include:
* automatic detection of request/response type
* pretty-printing of the following request/responses:
   - `"execute"` request/response
   - `"simplify"` request/response
   - `"add-module"` request/response

I've added a number of integration tests based on IMP. The test only check that the pretty-printer exited with 0, since comparing its output to golden files will likely prove a major headache.


dwightguth:
Why are we not comparing the output of the tests for correctness? If there's a reason not to, that's fine, but if so, we need to share the outputs on the pr for review.

ehildenb:
We do have other tests which exercise the endpoints more directly, eg all the tests in here: https://github.com/runtimeverification/pyk/tree/41706ef5081585a4c71f7fccb4dea2d7a3796421/src/tests/integration/kcfg

So I think if the goal is to only test the pretty-printing, it's OK to not check responses. Not sure if that's quite the goal.

dwightguth:
I mean, my concern is mainly that right now it's impossible for a reviewer to easily judge what the output of this command actually looks like, which is an important part of reviewing a PR like this. If we feel that pretty printing generally and the endpoints themselves are sufficiently tested that we don't need to test the combination, fine, but then we need to share some example outputs to this PR for each type of request/response that this PR is supposed to be able to handle.

geo2a:
Makes sense @dwightguth, I'll add golden files.

geo2a:
@dwightguth I've added golden files.

dwightguth (approved):

PR #700, opened 2023-10-27
`pyk rpc-kast` --- convert responses to new requests

geo2a:
This PR extends `pyk` with `pyk rpc-kast` --- and utility to convert Kore JSON RPC responses to new requests. Thiss functionality is useful when doing rpc server-level debugging.

Supported conversions:
* "execute" response to "execute" request
* "execute" response to "simplify" request

The tool also expects a reference request, which will be used to detect the type of the new request and fill-in the parameters. This way we can avoid a ton of command-line arguments to the tool that would st various request parameters. A good reference request is almost always around, and the parameters could be fine-tuned manually or with tools like `jq`.


src/pyk/__main__.py
```diff
@@ -188,6 +188,23 @@ def pretty_print_execute_response(execute_result: ExecuteResult) -> list[str]:
...
+    """
+    reference_request = json.loads(args.reference_request_file.read())
+    input_dict = json.loads(args.response_file.read())
+    execute_result = ExecuteResult.from_dict(input_dict['result'])
+    non_state_keys = set(reference_request['params'].keys()).difference(['state'])
+    request_params = {}
+    for key in non_state_keys:
+        request_params[key] = reference_request['params'][key]
+    request_params['state'] = {'format': 'KORE', 'version': 1, 'term': execute_result.state.kore.dict}
+    request = {'jsonrpc': '2.0', 'id': 42, 'method': reference_request['method'], 'params': request_params}
```
ehildenb, line 204: Why do we hardcode `id: 42`, could we just use th eoriginal ID? Not sure it matters.

ehildenb (approved):

src/pyk/__main__.py
geo2a: Yes, we should definitely use the original `id` , thanks!

PR #709, opened 2023-10-31
Parallel prover interface and tests

nwatson22:
Adds module `pyk.proof.parallel` which contains abstract class definitions `Prover`, `Proof`, `ProofStep`, and `ProofResult`. Intended for use with `APRProof`, `EqualityProof`, etc, to have an interface for defining subtasks for parallel execution which could then be done in a loop supporting all types of proofs which implement this interface.
Adds example implementers `TreeExploreProof`, etc. and tests them for some expected properties, as per https://github.com/runtimeverification/kontrol/issues/131


src/pyk/proof/equality.py
```diff
@@ -326,7 +326,7 @@ def lines(self) -> list[str]:
         ]
 
 
-class ImpliesProver(Prover):
+class ImpliesProver(Prover[ImpliesProof, None, None]):
```
tothtamas28, line 329: Do not modify these classes yet. Instead, let's create a separate module, e.g. `pyk.proof.parallel`, and put adapters that wrap these classes there.

Then can be sure that all changes are non-breaking, and add a comprehensive test suite.

src/tests/integration/proof/test_parallel_prove.py
```diff
@@ -0,0 +1,134 @@
...
+
+from pyk.proof.proof import Proof, ProofStatus, Prover
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+    from concurrent.futures import Executor, Future
+    from pathlib import Path
+
+
+class TreeExploreProof(Proof):
```
tothtamas28, line 16: Let's simplify this. Consider something like

```python3
class ReachProof(Proof):
    init: int
    target: int
    edges: dict[int, set[int]]
    reached: set[int]
```

Also, `Proof` does not have to be the same class as the one that e.g. `APRProof` extends, so that you only have to implement `status`.

src/pyk/proof/equality.py
nwatson22: Fixed.

src/tests/integration/proof/test_parallel_prove.py
nwatson22: Done.

src/pyk/proof/parallel.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class ProofResult(ABC):
+    """
+    Should be a description of how to make a small update to a `Proof` based on the results of a computation specified by a `ProofStep`.
+    Must be picklable.
+    Must be frozen dataclass.
+    Should be small.
+    """
+
```
nwatson22, line 84:
```suggestion
```

src/tests/integration/proof/test_parallel_prove.py
```diff
@@ -0,0 +1,211 @@
...
+
+@dataclass(frozen=True)
+class TreeExploreProofStep(ProofStep[TreeExploreProofResult]):
+    node: int
+
+    def __hash__(self) -> int:
+        return self.node.__hash__()
+
+    def exec(self) -> TreeExploreProofResult:
+#          print(f'Advancing node {self.node}\n', file=sys.stderr)
```
nwatson22, line 73:
```suggestion
```

src/tests/integration/proof/test_parallel_prove.py
```diff
@@ -0,0 +1,210 @@
...
+@dataclass(frozen=True)
+class TreeExploreProofStep(ProofStep[TreeExploreProofResult]):
+    node: int
+
+    def __hash__(self) -> int:
+        return self.node.__hash__()
+
+    def exec(self) -> TreeExploreProofResult:
+        time.sleep(5)
+#          print(f'Done advancing node {self.node}\n', file=sys.stderr)
```
nwatson22, line 74:
```suggestion
        time.sleep(1)
```

src/pyk/proof/parallel.py
```diff
@@ -0,0 +1,85 @@
...
+
+class Prover(ABC, Generic[P, S, U]):
+    """
+    Should contain all data needed to make progress on a `P` (proof).
+    May be specific to a single `P` (proof) or may be able to handle multiple.
+    """
+
+    @abstractmethod
+    def steps(self, proof: P) -> Iterable[S]:
```
tothtamas28, line 24:
```suggestion
class Prover(ABC, Generic[P, U]):
    """
    Should contain all data needed to make progress on a `P` (proof).
    May be specific to a single `P` (proof) or may be able to handle multiple.
    """

    @abstractmethod
    def steps(self, proof: P) -> Iterable[ProofStep[U]]:
```

src/pyk/proof/parallel.py
```diff
@@ -0,0 +1,85 @@
...
+        ...
+
+
+class ProofResult(ABC):
+    """
+    Should be a description of how to make a small update to a `Proof` based on the results of a computation specified by a `ProofStep`.
+    Must be picklable.
+    Must be frozen dataclass.
+    Should be small.
+    """
```
tothtamas28, line 83: This class defines no method, is thus superfluous.

```suggestion
```

src/tests/integration/proof/test_parallel_prove.py
```diff
@@ -0,0 +1,208 @@
...
+        self.edges[0] = {1, 2}
+        self.edges[1] = set()
+        self.edges[2] = {3, 4}
+        self.edges[3] = {5, 6}
+        self.edges[4] = {7}
+        self.edges[5] = set()
+        self.edges[6] = set()
+        self.edges[7] = {8, 9}
+        self.edges[8] = set()
+        self.edges[9] = set()
```
tothtamas28, line 49: This structure should not be hardcoded.

src/tests/integration/proof/test_parallel_prove.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def commit(self, proof: TreeExploreProof, update: TreeExploreProofResult) -> None:
+        if proof in TreeExploreProver.proofs:
+            assert TreeExploreProver.proofs[proof] == self
+        else:
+            TreeExploreProver.proofs[proof] = self
+        proof.reached.add(update.node)
+
+
+def prove_parallel(
```
tothtamas28, line 110: This function should be moved to `parallel.py`, and implemented in generic terms.

src/pyk/proof/parallel.py
nwatson22: Done.

src/pyk/proof/parallel.py
nwatson22: I removed it.

src/tests/integration/proof/test_parallel_prove.py
nwatson22: Tree structure is now passed in as constructor arguments.

src/tests/integration/proof/test_parallel_prove.py
nwatson22: Done.

src/tests/integration/proof/test_parallel_prove.py
```diff
@@ -0,0 +1,161 @@
...
+            return ProofStatus.PENDING
+
+
+@dataclass(frozen=True)
+class TreeExploreProofStep(ProofStep[int]):
+    node: int
+
+    def __hash__(self) -> int:
+        return self.node.__hash__()
+
```
tothtamas28, line 43: `dataclass` generates `__hash__` by default.

src/tests/integration/proof/test_parallel_prove.py
```diff
@@ -0,0 +1,161 @@
...
+    def exec(self) -> int:
+        time.sleep(1)
+        return self.node
+
+
+class TreeExploreProver(Prover[TreeExploreProof, int]):
+    proofs: dict[TreeExploreProof, TreeExploreProver] = {}
```
tothtamas28, line 50: What's the purpose of this attribute?

src/tests/integration/proof/test_parallel_prove.py
```diff
@@ -0,0 +1,161 @@
...
+    edges[4] = {7}
+    edges[5] = set()
+    edges[6] = set()
+    edges[7] = {8, 9}
+    edges[8] = set()
+    edges[9] = set()
+    return edges
+
+
+def test_multiple_provers_fails() -> None:
```
tothtamas28, line 107: What property does this test evaluate?

src/tests/integration/proof/test_parallel_prove.py
```diff
@@ -0,0 +1,161 @@
...
+        steps = prover.steps(proof)
+        if len(list(steps)) == 0:
+            break
+        for step in steps:
+            result = step.exec()
+            results.append(result)
+            prover.commit(proof, result)
+            prover.commit(proof, result)
+    for result in results:
+        prover.commit(proof, result)
```
tothtamas28, line 152: It's fine not to explicitly test for these I think.

nwatson22: Removed.

src/tests/integration/proof/test_parallel_prove.py
nwatson22: > a single P must only ever interact through this interface with one prover object.
Perhaps there is no need to test this, though.

src/tests/integration/proof/test_parallel_prove.py
nwatson22: I was trying to translate the requirement "a single P must only ever interact through this interface with one prover object" into code, but I guess this doesn't need to be enforced here. I could see this not being very useful since I don't know of a way to enforce it on the interface level. I can remove if you think so.

src/tests/integration/proof/test_parallel_prove.py
nwatson22: It must generate `__hash__` after it checks to see whether the class is concrete. Since `ProofStep` inherits `Hashable`, when I remove the explicit `__hash__`, I get the error:
```
Cannot instantiate abstract class "TreeExploreProofStep" with abstract attribute "__hash__"
```

src/tests/integration/proof/test_parallel_prove.py
```diff
@@ -0,0 +1,161 @@
...
+    edges[1] = set()
+    edges[2] = {3, 4}
+    edges[3] = {5, 6}
+    edges[4] = {7}
+    edges[5] = set()
+    edges[6] = set()
+    edges[7] = {8, 9}
+    edges[8] = set()
+    edges[9] = set()
+    return edges
```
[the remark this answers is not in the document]
nwatson22: Fixed.

src/pyk/proof/parallel.py
```diff
@@ -0,0 +1,126 @@
...
+            explored.add(step)
+            future = pool.submit(step.exec)  # <-- schedule steps for execution
+            pending[future] = proof
+
+    with ProcessPoolExecutor(max_workers=2) as pool:
+        for proof in proofs:
+            submit(proof, pool)
+
+        while pending:
+            future = list(wait(pending).done)[0]
```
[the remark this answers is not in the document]
nwatson22: Yes, but I see now it could be written more clearly as
```suggestion
            future = wait(pending).done.pop()
```
Does this look better? I guess another awkward spot with this is waiting on the dictionary with `Future`s as keys rather than a list of `Future`s, etc., but as long as this works I think this is probably the best way to write that.

src/pyk/proof/parallel.py
```diff
@@ -0,0 +1,126 @@
...
+        Allowed to be nondeterministic.
+        Able to be called on any `ProofStep` returned by `prover.steps(proof)`.
+        """
+        ...
+
+
+def prove_parallel(
+    proofs: list[Proof],
+    # We need a way to map proofs to provers, but for simplicity, I'll assume it as a given
+    provers: dict[Proof, Prover],
```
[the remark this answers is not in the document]
nwatson22: I've changed it to use `str` IDs. By a better way do you mean not passing in the mapping to `prove_parallel` and just passing in the proofs? Right now the contract isn't really specific enough for me to be able to see a way for this to be set up automatically. All we know is that there is some sort of one-to-many mapping of proofs to provers.

src/pyk/proof/parallel.py
```diff
@@ -0,0 +1,126 @@
...
+
+from pyk.proof.proof import ProofStatus
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+    from concurrent.futures import Executor, Future
+
+
+P = TypeVar('P', bound='Proof')
+U = TypeVar('U', bound='Any')
```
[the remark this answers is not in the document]
nwatson22: Fixed.

src/tests/integration/proof/test_parallel_prove.py
tothtamas28: That's unfortunate. Let's drop the `Hashable` superclass then, and add a line in the docstring with the requirement. Hash functions are a bit tricky, it's best not to force ourselves to implement them in all cases.

src/tests/integration/proof/test_parallel_prove.py
tothtamas28: Yes, let's remove it.

src/tests/integration/proof/test_parallel_prove.py
tothtamas28: Let's remove this as well then.

src/pyk/proof/parallel.py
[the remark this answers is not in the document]
tothtamas28: Yes, `pop` looks good.

By default, wait has `return_when=ALL_COMPLETED`, so let's fix that. Then let's split this into two lines:

```python3
done, _ = wait(pending, return_when=FIRST_COMPLETED)
future = done.pop()
```

src/tests/integration/proof/test_parallel_prove.py
nwatson22: Removed.

src/tests/integration/proof/test_parallel_prove.py
nwatson22: Removed.

src/tests/integration/proof/test_parallel_prove.py
nwatson22: Fixed.

src/pyk/proof/parallel.py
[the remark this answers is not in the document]
nwatson22: Fixed.

src/pyk/proof/parallel.py
```diff
@@ -0,0 +1,128 @@
...
+        prover = provers[proof_id]
+        for step in prover.steps(proof):  # <-- get next steps (represented by e.g. pending nodes, ...)
+            if step in explored:
+                continue
+            explored.add(step)
+            future = pool.submit(step.exec)  # <-- schedule steps for execution
+            pending[future] = proof_id
+
+    with ProcessPoolExecutor(max_workers=2) as pool:
+        for proof_id in proofs.keys():
```
tothtamas28, line 103:
```suggestion
        for proof_id in proofs:
```

src/tests/integration/proof/test_parallel_prove.py
```diff
@@ -0,0 +1,85 @@
...
+    def exec(self) -> int:
+        time.sleep(1)
+        return self.node
+
+
+class TreeExploreProver(Prover[TreeExploreProof, int]):
+    def __init__(self) -> None:
+        return
+
+    def steps(self, proof: TreeExploreProof) -> Iterable[TreeExploreProofStep]:
```
tothtamas28, line 47:
```suggestion
    def steps(self, proof: TreeExploreProof) -> list[TreeExploreProofStep]:
```

src/tests/integration/proof/test_parallel_prove.py
```diff
@@ -0,0 +1,85 @@
...
+    #         0
+    #        / \
+    #       1   2
+    #          / \
+    #         3   4
+    #        / \   \
+    #       5   6   7
+    #              / \
+    #             8   9
+    return {0: {1, 2}, 1: set(), 2: {3, 4}, 3: {5, 6}, 4: {7}, 5: set(), 6: set(), 7: {8, 9}, 8: set(), 9: set()}
```
tothtamas28, line 76:
```suggestion
    #         0
    #        / \
    #       1   2
    #          / \
    #         3   4
    #        / \   \
    #       5   6   7
    #              / \
    #             8   9
    SIMPLE_TREE = {0: {1, 2}, 1: set(), 2: {3, 4}, 3: {5, 6}, 4: {7}, 5: set(), 6: set(), 7: {8, 9}, 8: set(), 9: set()}
```

src/pyk/proof/parallel.py
nwatson22: Fixed.

src/tests/integration/proof/test_parallel_prove.py
nwatson22: Fixed.

src/tests/integration/proof/test_parallel_prove.py
nwatson22: Fixed.

nwatson22:
We have an option for reachability proofs, `fail_fast`, which I now realize doesn't make sense in the context of our requirement that there be nothing returned by `steps()` whenever `status` is `FAILED` or `PASSED`, since the whole notion of `fail_fast` is that proof failure can happen before all of the proof has been explored. If we relax that requirement and say that a proof could fail but still have steps to advance, we could handle `fail_fast` in this generic way by just treating it as "if `fail_fast` is enabled, as soon as the status is `FAILED` for a proof, don't submit any more `commits` to that proof".

tothtamas28:
> We have an option for reachability proofs, `fail_fast`, which I now realize doesn't make sense in the context of our requirement that there be nothing returned by `steps()` whenever `status` is `FAILED` or `PASSED`

I agree,`steps` should not be restricted this way for a `FAILED` test: https://github.com/runtimeverification/kontrol/issues/131#issuecomment-1792119914

src/pyk/proof/parallel.py
```diff
@@ -0,0 +1,143 @@
...
+                continue
+
+            prover.commit(proof, update)  # <-- update the proof (can be in-memory, access disk with locking, ...)
+
+            iterations[proof_id] += 1
+
+            match proof.status:
+                # terminate on first failure, yield partial results, etc.
+                case ProofStatus.FAILED:
+                    if fail_fast:
```
tothtamas28, line 133: Shouldn't this terminate `prove_parallel`? Because if not, then `fail_fast` can be simply emulated by any `prover` where

```
IF proof.status == ProofStatus.FAILED THEN len(list(prover.steps(proof))) == 0
```

src/pyk/proof/parallel.py
```diff
@@ -0,0 +1,143 @@
...
+            done, _ = wait(pending, return_when='FIRST_COMPLETED')
+            future = done.pop()
+            proof_id = pending[future]
+            proof = proofs[proof_id]
+            prover = provers[proof_id]
+            update = future.result()
+
+            if max_iterations is not None and iterations[proof_id] >= max_iterations:
+                pending.pop(future)
+                continue
```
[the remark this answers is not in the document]
nwatson22: Ok, maybe I was thinking about this wrong. I assumed we should offer the feature for free if it makes sense in general.

src/pyk/proof/parallel.py
nwatson22: `fail_fast` only terminates individual proofs when that proof fails, so yes, it could be implemented at the prover level too.

src/pyk/proof/parallel.py
```diff
@@ -0,0 +1,143 @@
...
+            iterations[proof_id] += 1
+
+            match proof.status:
+                # terminate on first failure, yield partial results, etc.
+                case ProofStatus.FAILED:
+                    if fail_fast:
+                        pending.pop(future)
+                        continue
+                case ProofStatus.PENDING:
+                    assert len(list(prover.steps(proof))) > 0
```
anvacaru, line 137: I'm not sure if this applies here, but `assert` is ommited if the code is executed with optimizations. Also, you could add some error messages.

src/pyk/proof/parallel.py
```diff
@@ -0,0 +1,143 @@
...
+            submit(proof_id, pool)
+            iterations[proof_id] = 0
+
+        while pending:
+            done, _ = wait(pending, return_when='FIRST_COMPLETED')
+            future = done.pop()
+            proof_id = pending[future]
+            proof = proofs[proof_id]
+            prover = provers[proof_id]
+            update = future.result()
```
anvacaru, line 120: Maybe insert this in a `try`.. `except` block to handle any possible errors?

src/pyk/proof/parallel.py
nwatson22: Fixed.

src/pyk/proof/parallel.py
nwatson22: I'm not sure what the philosophy about catching/rethrowing errors is here. I've added something to catch errors which the `result()` function can throw. However most of these except the generic `Exception` which I guess is used to propagate the exception from the task code are things that should not happen by design (the task should never be cancelled because we don't cancel it and the timeout shouldn't happen because the future is already returned by `wait` as done. Is this about what you meant?

src/pyk/proof/parallel.py
```diff
@@ -0,0 +1,143 @@
...
+                raise RuntimeError('Exception was raised in ProofStep.exec() for proof {proof_id}.') from err
+
+            prover.commit(proof, update)  # <-- update the proof (can be in-memory, access disk with locking, ...)
+
+            match proof.status:
+                # terminate on first failure, yield partial results, etc.
+                case ProofStatus.FAILED:
+                    ...
+                case ProofStatus.PENDING:
+                    if len(list(prover.steps(proof))) == 0:
```
tothtamas28, line 135:
```suggestion
                    if not list(prover.steps(proof)):
```

src/pyk/proof/parallel.py
```diff
@@ -0,0 +1,143 @@
...
+                        raise ValueError('Prover violated expectation. status is pending with no further steps.')
+                case ProofStatus.PASSED:
+                    if len(list(prover.steps(proof))) > 0:
```
tothtamas28, line 138:
```suggestion
                    if list(prover.steps(proof)):
```

src/pyk/proof/parallel.py
tothtamas28: @anvacaru is right that this is the statement that should be surrounded with `try` to handle errors more gracefully. In my opinion, this can happen in a follow-up PR (this is related to our discussion about isolating provers with failed steps).

ehildenb (approved):
Looks good overall, let's get to testing it.

PR #767, opened 2023-12-08
Add parameter `type_inference_mode` to `kompile`

tothtamas28:
And use `--type-inference-mode checked` when kompiling using the test fixtures.


radumereuta (commented):
Can you do the same for calling kprove?
The option should be available once this merges: https://github.com/runtimeverification/k/pull/3855

tothtamas28:
> Will wait for the equivalent change to `kprove` to review, but looks good to me in principle.

I wanted to open a separate PR for that, In my opinion, this is a self-contained change.

Baltoli (approved):
Makes sense; thanks @tothtamas28!

PR #770, opened 2023-12-08
BugReport reporter

gtrepta:
fixes: runtimeverification/k#4189


src/pyk/utils.py
```diff
@@ -476,26 +477,61 @@ def abs_or_rel_to(path: Path, base: Path) -> Path:
 
 class BugReport:
     _bug_report: Path
+    _executor: ThreadPoolExecutor | None = None
```
tothtamas28, line 480: I don't think this class requires a thread pool. But it probably should have a lock to ensure mutual exclusion on parallel file access.

src/pyk/utils.py
```diff
@@ -476,25 +477,59 @@ def abs_or_rel_to(path: Path, base: Path) -> Path:
...
+        self._check_bug_report()
+
+    def add_file(self, finput: Path, arcname: Path) -> None:
+        pass
+
+    def add_file_contents(self, input: str, arcname: Path) -> None:
+        pass
+
+    def add_command(self, args: Iterable[str]) -> None:
+        pass
```
tothtamas28, line 498: By organizing the code this way, this will not break:

```python3
br = BugReporter(Path('foo'))
br.add_file(Path('bar'), Path('baz'))
```

yet the bug report entry will not be written.

`BugReporter` should not be a `BugReport`, and these methods should be removed from `BugReport`.

src/pyk/utils.py
```diff
@@ -476,25 +477,59 @@ def abs_or_rel_to(path: Path, base: Path) -> Path:
...
+        if self._lock is None:
+            self._lock = Lock()
+        return BugReporter(self._bug_report, prefix, self._lock)
+
+
+class BugReporter(BugReport):
     _command_id: int
     _defn_id: int
     _file_remap: dict[str, str]
+    _lock: Lock
```
tothtamas28, line 510: What properties does this lock ensure?

src/pyk/utils.py
```diff
@@ -476,25 +477,59 @@ def abs_or_rel_to(path: Path, base: Path) -> Path:
...
+    def add_file(self, finput: Path, arcname: Path) -> None:
+        pass
+
+    def add_file_contents(self, input: str, arcname: Path) -> None:
+        pass
+
+    def add_command(self, args: Iterable[str]) -> None:
+        pass
+
+    def reporter(self, prefix: Path) -> BugReporter:
```
tothtamas28, line 500:
```suggestion
    def reporter(self, prefix: str) -> BugReporter:
```

tothtamas28: What happens if two threads request a reporter to the same `prefix`?

src/pyk/utils.py
tothtamas28: Then spurious methods like `_check_bug_report` can also be removed from `BugReporter`.

gtrepta: How should users instantiate a bug report? It looks to me that currently the `bug_report_arg` parameter from `cli` is the primary use case, and I've ensured that the reporter coming from there works. The intent was to have `BugReporter` only be instantiated by a call to `BugReport.reporter`, and have `BugReport` hold the base interface, file paths, and synchronization primitive to spawn reporters from. Something like your example that doesn't write to a file can certainly be a problem if it isn't discouraged in some way, like renaming the class to `_BugReporter` so it doesn't get imported or used as easily.

If we want to allow users to directly create a `BugReport` or a `BugReporter` then I'll have to think a bit more about what this should look like.

I made `BugReporter` an inherited class because `BugReport` is expected in many places like KRun and Kompile, and wanted to keep those interfaces stable.

src/pyk/utils.py
gtrepta: Assuming the usage pattern expected in my above comment, this Lock should be shared by all BugReporters that were created off of the same BugReport, so that only one of them can write to the file at a time.

src/pyk/utils.py
gtrepta: They will both be writing to the same prefix in the bug report archive. The Lock should ensure that they don't write at the same time, but it is possible that they write to the same path, which in this implementation I believe will work like an append. It should probably be up to the user whether that's an issue or not.

src/pyk/utils.py
tothtamas28: > How should users instantiate a bug report?

In the final design, users should be able to instantiate a `BugReport` (i.e. the `BugReporter` factory) directly. On the other hand, `BugReporter` (i.e. the replacement of the old `BugReport`) should only be instantiated through `BugReporter`.

(We can also consider a composite design where a `BugReporter` can instantiate further `BugReporter` instances, e.g. a `BugReporter` for a proof process creates a `BugReporter` for one of the `KoreClient` instances it manages.)

> `BugReport` is expected in many places like `KRun` and `Kompile`, and wanted to keep those interfaces stable.

I see why this is a good approach, but ultimately the goal is to get rid of the old interface, and instead

* Let the user decide where to write the archive: `BugReport`.
* Let the application (e.g. `kontrol`) decide how to structure the archive: `BugReporter`.
* Let the library (e.g. `KoreServer`) be responsible only for what's written to the archive.

 One way to achieve this:

1) Turn `BugReport` into an abstract class
2) Have two implementations: 1) old behavior 2) `BugReporter`
3) Gradually replace old behavior by `BugReporter`
4) Do the renamings 

All this doesn't have to be achieved in a single PR.

src/pyk/utils.py
tothtamas28: What happens in a multiprocessing scenario, i.e. when the `Lock` gets pickled and sent to another Python process? Will it still ensure mutual exclusion?

src/pyk/utils.py
tothtamas28: I think we can safely introduce this restriction (i.e. let the `BugReporter` own its prefix) as the way we use the archive is always adding new things to it.

If the same prefix is requested twice, the call can either fail, or the target directory can be renamed to make it unique:

```
foo
foo_1
foo_2
...
```

What do you think?

src/pyk/utils.py
gtrepta: By the looks of it, pickling Lock objects isn't supported. Would this be an issue?

src/pyk/utils.py
gtrepta: Renaming while logging a message about it sounds reasonable.

src/pyk/utils.py
tothtamas28: Yes, we use multiprocessing quite intensively in `kontrol`.

An alternative would be a file lock ([`filelock`](https://py-filelock.readthedocs.io/en/latest/index.html), cross-platform) or OS level locking ([`os.lockf`](https://docs.python.org/3/library/os.html#os.lockf), [`fcntl`](https://docs.python.org/3/library/fcntl.html), only on Unix). Also, please check whether [`tarfile`](https://docs.python.org/3/library/tarfile.html) does any form of locking, maybe it already does.

src/pyk/utils.py
gtrepta: I've force pushed this branch with a change that implements 1 and 2. There's no thread safety over the BugReporter yet, if you're happy with deferring that to a later PR then we should be able to merge this.

src/pyk/utils.py
```diff
@@ -524,3 +521,26 @@ def _remap_arg(_a: str) -> str:
         shebang = '#!/usr/bin/env bash\nset -euxo pipefail\n'
         self.add_file_contents(shebang + ' '.join(remapped_args) + '\n', arcname)
         self._command_id += 1
+
+
+class BugReport(BugReportBase):
```
tothtamas28, line 526: Consider this hierarchy instead:

* `BugReportBase` - abstract class without any actual implementation.
* `BugReport` - same as before, but now implements `BugReportBase`.
* `BugReport2` - something that can instantiate a `BugReporter` for a prefix.
* `BugReporter` - a `BugReportBase` that writes into a folder of an archive.

This is the easy part as so far, nothing breaks. The hard part is eliminating `BugReport` in favor of the new design, which will be a breaking change in `pyk` and in downstream projects.

We should only merge the PR with the new classes once we have a plan on how to manage the rest. To get the intuition of what needs to be changed, you can rename `BugReport` and see where the type checker complains.

PR #773, opened 2023-12-11
Refactor `cterm_execute`

tothtamas28:
* Extract class `CTermExecute` to represent the return value
* Assert some properties on the RPC response and simplify logic


ehildenb:
What is the purpose of this change?

`cterm_execute` currently loses a lot of information that the RPC call actually returns, re-inferring it based on things like the number of next states. If anything, we should be propagating more of that logic up, and dealing directly with the `ExecuteResult` that we already have: https://github.com/runtimeverification/pyk/blob/b5c15a28a350a7ad1e353f68a023310d4639b441/src/pyk/kore/rpc.py#L502.

tothtamas28:
> What is the purpose of this change?

It's a clean-up:

1) Replaces a five-tuple return value by a class
2) Eliminates some unreachable branches and asserts the corresponding conditions (tested in `evm-semantics` and `kontrol`)

> If anything, we should be propagating more of that logic up, and dealing directly with the `ExecuteResult`

I agree, the more complexity we can eliminate the better, we could replace the return type by a wrapper on `ExecuteResult` for example. But I don't see how this PR would prevent that, in fact I think we can be more confident in the correctness of such a change after this PR has been merged, as it replaces

```python3
if len(next_states) == 1 and len(next_states) < len(_next_states):
    return ...
elif len(next_states) == 1:
    if er.reason == StopReason.CUT_POINT_RULE:
        return ...
    else:
        next_states = []
return ...
```

by

```python3
return ...
```

ehildenb (approved):

ehildenb:
Sounds good, thank you for the explanation!

PR #789, opened 2023-12-18
Fix `_kore_to_kast`

tothtamas28:
Fixes a bug introduced in #775


Baltoli (approved):
Looks good to me; perhaps consider adding a regression test if there's one that can be conveniently derived from the original discovery of the issue?

tothtamas28:
> consider adding a regression test if there's one that can be conveniently derived from the original discovery of the issue

Added a profile test for the problematic term.

Issue #677, opened 2023-10-10
Start documenting public API


tothtamas28:
- [ ] Look into docstring formats and tools for generating documentation
- [ ] Set up infrastructure for generating documentation from docstrings
- [ ] Set up infrastructure for deploying generated documentation
- [ ] Start adding docstrings for newly added / touched public modules, functions and classes

ehildenb:
Related: https://github.com/runtimeverification/erc20-verification/issues/313

tothtamas28:
> Look into docstring formats and tools for generating documentation

We'll use [Sphinx](https://www.sphinx-doc.org) with its own [docstring format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html). This is probably the most common setup by far.

PR #811, opened 2024-01-12
Add Sphinx setup

tothtamas28:
Related: #677

Run `make docs` to generate HTML documentation under `docs/build`. (Right now the build emits a lot of warnings.)

Run `make clean-docs` to clean generated docs files.


ehildenb:
Could we also have it publish the sphinx docs to the GH pages branch? We can do that in a follow-on PR, but having that automated publishing would be really nice!

tothtamas28:
> Could we also have it publish the sphinx docs to the GH pages branch?

Definitely, this is the upcoming item on the [list](https://github.com/runtimeverification/pyk/issues/677).

ehildenb (approved):

PR #812, opened 2024-01-15
Refine `KoreClientError`

### LINKED ISSUE #807: Refine `KoreClientError` to include additional details returned by the server
When a request fails, an instance of `KoreClientError` is thrown by `KoreClinet` which will contain the error infomation returned by the backend. In the upstream projects, i.e. `kevm_pyk`, this exception will be [shown](https://github.com/runtimeverification/evm-semantics/blob/5ab752002fdc669e5d3736f8350676ec9aec76d6/kevm-pyk/src/kevm_pyk/utils.py#L147C72-L147C72) to the used as part of an ERROR-level log message:

```
    except Exception as e:
        _LOGGER.error(f'Proof crashed: {proof.id}\n{e}', exc_info=True)
        return False
```

As I understand, the code above will internally call `str(e)` to convert the exception to a string, which will actually only include `e.message`. However, the `err.data` field contains additional error information that may be worth displaying too: https://github.com/runtimeverification/haskell-backend/blob/master/docs/2022-07-18-JSON-RPC-Server-API.md#error-response

Perhaps we could refine the `KoreClientError` exception class and override its `str` representation to include the additional error information from the "data"->"error" field if it is present?

@tothtamas28

tothtamas28:
Closes #807

Adds subclasses `ParseError`, `PatternError`, `ModuleError`, `ImplicationError` and `DefaultError` for `KoreClientError`. `KoreClientError` becomes abstract, and is replaced by `DefaultError` to represent any, so far not refined application error.


geo2a (approved):
LGTM, thanks @tothtamas28!

PR #832, opened 2024-01-22
Better handling of subproof dependencies when executing superproofs

ehildenb:
~Blocked on: https://github.com/runtimeverification/pyk/pull/835~

In this PR, we setup for CSE (compositional symbolic execution) in Kontrol. This PR enables using the existing subproof mechanism for setting up that possibility. In particular:

- The method `KCFG.to_rules` is factored out of `KCFG.to_module`, to produce the list of rules associated with the basic blocks of the KCFG.
- The injection of dependencies for a given subproof is modified to use this `KCFG.to_rules` instead, so that it is injecting the basic blocks of the subproof instead of the created transitino from init to target term.
- A proof that was marked as `PENDING` before when having an un-admitted dependency is now marked as `FAILING` because of the more careful way we handle un-admitted dependencies when adding dependencies modules.
- Improves the module name munging to produce Modules from KCFGs.


ehildenb:
I've confirmed that this enables a simple CSE on this branch of Kontrol: https://github.com/runtimeverification/kontrol/pull/284

anvacaru (approved):

PR #834, opened 2024-01-23
Add new rpc error types

goodlyrottenapple:


src/pyk/kore/rpc.py
```diff
@@ -368,13 +368,35 @@ def __init__(self, error: str, context: Iterable[str]):
...
+    def __init__(self, error: str, context: Iterable[str] | None):
+        self.error = error
+        self.context = tuple(context) if context else None
+        context_str = ' Context: ' + ' ;; '.join(self.context) if self.context else ''
+        super().__init__(f'Could not verify module: {self.error}{context_str}')
+
+
+@final
+@dataclass
+class DuplicateModuleNameError(KoreClientError):
```
tothtamas28, line 393: Just for succinctness (and symmetry with the other module errors) I'd call this:

```suggestion
class DuplicateModuleError(KoreClientError):
```

src/pyk/kore/rpc.py
```diff
@@ -368,13 +368,35 @@ def __init__(self, error: str, context: Iterable[str]):
...
 
     def __init__(self, module_name: str):
         self.module_name = module_name
         super().__init__(f'Could not find module: {self.module_name}')
 
+@final
+@dataclass
+class InvalidModuleError(KoreClientError):
+    error: str
+    context: tuple[str, ...] | None
```
tothtamas28, line 382: Are there cases where `context` a) is not provided b) is the empty list? If so, are these cases semantically different?

goodlyrottenapple: Sometimes context is `None`, this is because the specific errors are slightly different between the two back-ends. I'm not sure the context is ever an empty list

tothtamas28: Looks good then.

PR #835, opened 2024-01-23
Implement first past of KFlatModule conversion to Kore

ehildenb:
This PR is to begin a refactoring needed for CSE, and also begin work on implementing the skeleton of ModuleToKORE in pyk.

- A new routine is added to `konvert` for `kflatmodule_to_kore`, which converts modules which contain only rules and imports to Kore.
- This routine is used for building up the dependencies and circularities modules in initializing the `APRProver`, and then putting them into the `add_module` endpoint directly.
- The code `KCFG.as_claim` is turned into a simpler `KCFG.as_rule`, which uses existing routines instead of making a new one. This uses the `cterm` modules `build_rule` functionality, which is adjusted to not strip out information needed for making backend-compatible rules (previously this routine was used for display purposes only).
- Old code for handling the dependencies/circularities modules in `APRProver` is removed.


nwatson22 (approved):
I think I am just missing some context for this. How/for what is `ModuleToKORE` intended to be used?

ehildenb:
`ModuleToKORE` is implemented in the Java frontend [here](https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java). It implements the frontends translation of KAST to Kore.

This takes us one step closer to fully implementing that in pyk instead of in the Java frontend.

src/pyk/proof/reachability.py
```diff
@@ -659,21 +653,29 @@ def __init__(
 
         apr_subproofs: list[APRProof] = [pf for pf in subproofs if isinstance(pf, APRProof)]
 
-        dependencies_as_claims: list[KClaim] = [d.as_claim(self.kcfg_explore.kprint) for d in apr_subproofs]
+        dependencies_as_rules: list[KRule] = [d.as_rule(priority=20) for d in apr_subproofs]
```
tothtamas28, line 656: Consider extracting a helper method for this initialization logic (i.e. constructing the modules, then sending the requests).

tothtamas28 (approved):

src/pyk/proof/reachability.py
ehildenb: Done.

PR #842, opened 2024-01-26
Fully qualify claim labels/dependencies, include dependencies in output

ehildenb:
We were having trouble in https://github.com/runtimeverification/evm-semantics/pull/2272 selecting the correct claims because they can be unqualified by the user. This PR:

- Fully qualifies all claim labels and dependency labels returned by `KProve.get_claim_modules`.
- Makes sure that we return not only the user-selected claims, but their dependencies (unless the user turns that off), from `KProve.get_claims`.
- Adds a test-suite for the claim selection functionality.

There is a bug in this code, and a test added (and skipped using `pytest.skip()`) which demonstrates a (probably minor) issue where we are not checking that claim dependencies are in a module imported by the current module.


src/pyk/ktool/kprove.py
```diff
@@ -355,16 +377,40 @@ def get_claims(
...
+            claim_label = claim_labels.pop(0)
+            if claim_label in final_claim_labels:
+                continue
+            elif claim_label in exclude_claim_labels:
+                _LOGGER.warning(f'Including claim that is also excluded: {claim_label}')
+            else:
+                final_claim_labels.append(claim_label)
+                if include_dependencies:
+                    claim_labels.extend(
+                        [d for d in all_claims[claim_label].dependencies if d not in final_claim_labels]
```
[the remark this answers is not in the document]
tothtamas28: Here, maintaining uniqueness of elements is more expensive than doing the filtering at the end.

src/pyk/ktool/kprove.py
```diff
@@ -355,16 +377,40 @@ def get_claims(
             md_selector=md_selector,
         )
 
-        all_claims = {c.label: c for m in flat_module_list.modules for c in m.claims}
-        unfound_labels = []
-        claim_labels = list(all_claims.keys()) if claim_labels is None else claim_labels
-        exclude_claim_labels = [] if exclude_claim_labels is None else exclude_claim_labels
-        unfound_labels.extend([cl for cl in claim_labels if cl not in all_claims])
-        unfound_labels.extend([cl for cl in exclude_claim_labels if cl not in all_claims])
+        all_claims = {
```
tothtamas28, line 380: Just to check my understanding, let's consider the following example

```k
module FOO
    claim [foo]: true
endmodule

module BAR
    claim [bar]: true
endmodule
```

Suppose the main module is `FOO`.

Before this change, the result was `foo` and `bar`, and filtering was based on these same labels, i.e. `foo` and `bar`.

After the change, the result is `FOO.foo` and `FOO.BAR.bar` (it might be possible to construct a label clash between claims in different modules), and filtering is based on labels obtained after the first renaming, i.e. `FOO.foo` and `BAR.bar`. In that case, it is not ensured that the result labels and the filter labels match.

In order to make this more straightforward, would the following work?

* In `qualify_name`, always qualify the claim label with the module name. This ensures that there is no label clash between claims of a module in `get_claim_modules`. Also, when using this API, it will be straightforward to refer to claim labels without calling `KProve.qualify_name` (thus it can become private).
* Do not do a second round of qualification in `get_claims`. Since module names are unique, and claim labels within a module are also unique, the resulting claim list will have no label clashes. Also, the filters come from the same domain as the claim labels of the result.

src/pyk/ktool/kprove.py
```diff
@@ -355,16 +377,40 @@ def get_claims(
...
+        unfound_labels: list[str] = [cl for cl in claim_labels + exclude_claim_labels if cl not in all_claims]
         if len(unfound_labels) > 0:
             raise ValueError(f'Claim labels not found: {unfound_labels}')
 
-        return [all_claims[cl] for cl in all_claims if cl in claim_labels and cl not in exclude_claim_labels]
+        final_claim_labels: list[str] = []
+        while len(claim_labels) > 0:
+            claim_label = claim_labels.pop(0)
+            if claim_label in final_claim_labels:
+                continue
```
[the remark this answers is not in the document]
ehildenb: No, it's because multiple claims could have the same dependencies.

src/pyk/ktool/kprove.py
```diff
@@ -355,16 +377,40 @@ def get_claims(
...
+            elif claim_label in exclude_claim_labels:
+                _LOGGER.warning(f'Including claim that is also excluded: {claim_label}')
```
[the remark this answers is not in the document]
ehildenb: It's not an error though, it's just a warning.

src/pyk/ktool/kprove.py
ehildenb: No, it should be `FOO.foo` and `BAR.bar`. I can add a test to make sure. But name qualification happens within the given module in `KProve.get_claim_modules`.

ehildenb:
I'm redoing the logic here completely, I agree that the original way was very fragile.

src/pyk/ktool/kprove.py
```diff
@@ -321,6 +321,10 @@ def prove_cterm(
         ]
         return next_states if len(next_states) > 0 else [CTerm.top()]
 
+    @staticmethod
+    def qualify_name(_module_name: str, _claim_label: str) -> str:
+        return _claim_label if _claim_label.startswith(_module_name) else f'{_module_name}.{_claim_label}'
```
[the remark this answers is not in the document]
ehildenb: No longer have this issue.

src/pyk/ktool/kprove.py
```diff
@@ -337,7 +341,24 @@ def get_claim_modules(
...
+                        _att = sentence.att.update({'label': _label})
+                        if len(sentence.dependencies) > 0:
+                            _dependencies = [KProve.qualify_name(module.name, dep) for dep in sentence.dependencies]
+                            _att = _att.update({'depends': ','.join(_dependencies)})
+                        sentences.append(sentence.let(att=_att))
+                    else:
+                        sentences.append(sentence)
+                modules.append(module.let(sentences=sentences))
+
+            return flat_module_list.let(modules=modules)
```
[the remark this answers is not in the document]
ehildenb: This code doesn't exist anymore.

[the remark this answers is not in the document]
ehildenb: And tests are added of the overall routine.

src/pyk/ktool/kprove.py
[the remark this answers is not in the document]
ehildenb: We now throw out claims that are both included and excluded.

src/pyk/ktool/kprove.py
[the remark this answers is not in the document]
ehildenb: Removed.

src/pyk/ktool/kprove.py
ehildenb: This should be fixed now.

src/pyk/ktool/kprove.py
```diff
@@ -355,16 +356,49 @@ def get_claims(
...
 
-        all_claims = {c.label: c for m in flat_module_list.modules for c in m.claims}
-        unfound_labels = []
-        claim_labels = list(all_claims.keys()) if claim_labels is None else claim_labels
-        exclude_claim_labels = [] if exclude_claim_labels is None else exclude_claim_labels
-        unfound_labels.extend([cl for cl in claim_labels if cl not in all_claims])
-        unfound_labels.extend([cl for cl in exclude_claim_labels if cl not in all_claims])
+        all_claims = {
+            claim.label: (claim, module.name) for module in flat_module_list.modules for claim in module.claims
+        }
```
tothtamas28, line 361: `claim.label` is not necessarily unique across modules. This `dict` will only store the last claim with the given label.

src/pyk/ktool/kprove.py
```diff
@@ -355,16 +356,49 @@ def get_claims(
...
+
+        claim_labels = list(all_claims.keys()) if claim_labels is None else list(claim_labels)
+        exclude_claim_labels = [] if exclude_claim_labels is None else list(exclude_claim_labels)
+
+        unfound_labels: list[str] = [
+            cl
+            for cl in list(claim_labels) + list(exclude_claim_labels)
+            if cl not in all_claims and f'{flat_module_list.main_module}.{cl}' not in all_claims
```
tothtamas28, line 369: What's the purpose of the second conjunct? So a claim can be referred to as `foo`, and also as `MAIN-MODULE.foo`?

tothtamas28 (commented):
Please add a docstring for the method that specifies how the module names and claim labels in `spec_file` relate to `claim_labels` and the claim labels in the result.

src/pyk/ktool/kprove.py
ehildenb: Yeah, that's the intention. What do you think?

src/pyk/ktool/kprove.py
ehildenb: Hmmmm, I wonder how we should handle this then. Let me make an example test and see if I can make it work as well.

ehildenb: Ok, it turns out the frontend always qualifies labels that it returns on the claims anyway. So I've simplified the code with that assumption.

src/pyk/ktool/kprove.py
ehildenb: I've simplified this based on the assumption that the frontend always returns qualified label names.

ehildenb:
@tothtamas28 I'm going to get this merged to unblock KEVM stuffs, we can discuss more improvements as follow-up.

nwatson22 (approved):
Is adding a `depends` with an unqualified label supposed to only work if a claim with that label exists in the current module, or should you be allowed to specify as a dependency an unqualified claim label from an imported module?

ehildenb:
It should only work in the current module I suppose. But I guess overall it's ill-defined, we don't have a reference or spec for it.

src/pyk/ktool/kprove.py
tothtamas28: I guess we do not really have integration tests on `--emit-json-spec`. Please open an issue for testing this assumption and other structural features of the resulting definition.

PR #844, opened 2024-01-26
Implement KAST-to-KORE conversion of sort declarations

tothtamas28:
Adds a function `module_to_kore` that converts a given module of a KAST `KDefinition` to a KORE `Module`.

So far only `syntax` sentences are supported. The function is tested on `IMP`.

## How it works

First, do the following KAST-to-KAST transformations (`simplified_module`).
1. Flatten module (`_flatten_module`).
2. Make sort declarations explicit (`_add_syntax_sorts`).
3. Add attributes `concat`, `element` and `unit` for built-in collection sort declarations (`_add_collection_atts`).
4. Add attribute `hasDomainValue` to token sort declarations (`_add_domain_value_atts`).

Then, do a straightforward conversion of the declarations in the flat KAST module into KORE.


src/pyk/konvert.py
```diff
@@ -30,26 +31,298 @@
...
+
+def _add_collection_atts(module: KFlatModule) -> KFlatModule:
+    """Return a module where concat, element and unit attributes are added to collection sort declarations"""
+
+    # Example: syntax Map ::= Map Map [..., klabel(_Map_), element(_|->_), unit(.Map), ...]
+    concat_prods = {prod.sort: prod for prod in module.productions if KAtt.ELEMENT in prod.att}
+
+    assert all(
+        KAtt.UNIT in prod.att for _, prod in concat_prods.items()
+    )  # TODO Could be saved with a different attribute structure: concat(Element, Unit)
```
Baltoli, line 259: Yes, agreed - maybe open an issue and we can think about migrating this? Cleaner than having separate attributes.

Baltoli (approved):
A few small questions / comments, but this looks like a really good start to the process!

src/pyk/konvert.py
```diff
@@ -30,26 +31,298 @@
...
+}
+
+
+def module_to_kore(definition: KDefinition) -> Module:
+    """Convert the main module of a kompiled KAST definition to KORE format."""
+
+    module = simplified_module(definition)
+
+    name = name_to_kore(module.name)
+    attrs = atts_to_kore({key: value for key, value in module.att.items() if key != 'digest'})  # filter digest
```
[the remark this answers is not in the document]
tothtamas28: I'm not sure about the underlying reason, but that's how `ModuleToKORE.java` is implemented: https://github.com/runtimeverification/k/blob/7d5882aa7a28139cb733bfadedf36794a360a60b/kernel/src/main/java/org/kframework/backend/kore/ModuleToKORE.java#L381

src/pyk/konvert.py
tothtamas28: Done: https://github.com/runtimeverification/k/issues/3933

PR #850, opened 2024-01-30
new module to load bindings from llvm backend distro

dwightguth:
I spoke with Bruce and he suggested this approach for dealing with the `_kllvm` bindings module in Pi^2. Essentially, the LLVM backend builds the bindings module with a particular pre-configured version of python, and then this code makes the assumption that that version of python is ABI-compatible with the version of python used by the code importing this module. It may not be true in general, but in cases where we can control both the version of K and the version of pyk that are installed, it is possible to guarantee this. This then becomes part of the requirements for a successful installation of the tool in question. I tested that locally, this seems to cause the unit tests to pass. I can also confirm that this causes a significant improvement in the time to run a short-lived python invocation that imports this module, on the order of about 6 seconds.


Baltoli:
> It may not be true in general, but in cases where we can control both the version of K and the version of pyk that are installed, it is possible to guarantee this.

Yep, that's correct - just for the sake of writing down what I said earlier, in my experience this tends to only be an issue when running in a local dev environment. When built into a production / packaged release of a tool we can better control dependency versions and make sure there isn't a mismatch.

Baltoli (approved):
Marked as a draft so I won't automerge, but the code here is what I'd expect to see based on our conversation earlier.

PR #857, opened 2024-02-03
Print parseable rules from KCFGShow

ehildenb:
In this PR: https://github.com/runtimeverification/pyk/pull/835, the function `build_rule` is modified to maintain all structure of a rule. But that routine is also used for creating pretty-printing rules for parsing via the frontend. Rules that need to be parsed via the frontend must not contain the `<generated*>` cells.

This PR makes sure that `KCFGShow(...)` still minimizes the rules that are printed via `to_module`. This should allow undoing most of the output changes here: https://github.com/runtimeverification/kontrol/pull/325.


src/pyk/kcfg/show.py
```diff
@@ -302,8 +293,23 @@ def pretty(
...
+        parseable_output: bool = True,
+    ) -> KFlatModule:
+        module = cfg.to_module(module_name)
+        _sentences = []
+        for sent in module.sentences:
+            if type(sent) is KRule:
+                sent = sent.let(body=KCFGShow.hide_cells(sent.body, omit_cells))
+                if parseable_output:
+                    sent = sent.let(body=remove_generated_cells(sent.body))
+                    sent = minimize_rule(sent)
```
anvacaru, line 310: Perhaps move this into an inner `_process_sentence(sent: KSentence) -> KRule` and have just
```
    _sentences = [_process_sentence(sent) for sent in module.sentences]
``` 
instead of the `for` loop.

anvacaru (approved):
Approved with a single suggestion :+1:

src/pyk/kcfg/show.py
ehildenb: Done.

PR #861, opened 2024-02-05
Implement KAST-to-KORE conversion of symbol declarations

tothtamas28:
Process attributes on symbol productions (i.e. productions with at least one non-terminal that are not `bracket` productions), then transform these productions to KORE symbol declarations.

Attributes that are addressed:
* `hook`
* `macro`
* `anywhere`
* `function`
* `injective`
* `constructor`

Attributes that are not addressed:
* `colors`, `format`, `left`, `priorities`, `right`, `strict`, `terminals` is ignored by the test
* `production` is discarded by the transformation


src/pyk/konvert.py
```diff
@@ -410,6 +411,27 @@ def _rules_by_klabel(module: KFlatModule) -> dict[KLabel, list[KRule]]:
     return res
 
 
+def _add_functional_atts(module: KFlatModule) -> KFlatModule:
+    """Add the macro attribute to all symbol productions that are not function or total"""
```
Baltoli: typo here?

src/pyk/konvert.py
```diff
@@ -432,6 +433,25 @@ def update(sentence: KSentence) -> KSentence:
...
+    """Add the injective attribute to all symbol productions that satisfy the criteria"""
+
+    def update(sentence: KSentence) -> KSentence:
+        if not isinstance(sentence, KProduction):
+            return sentence
+
+        if not sentence.klabel:
+            return sentence
+
+        if not any(att in sentence.att for att in [KAtt.FUNCTION, KAtt.ASSOC, KAtt.COMM, KAtt.IDEM]):
```
Baltoli: I think we dropped this usage of the `comm` attribute; it now generates symmetric cases for simplification rules. Given that perhaps we should look at how `assoc` and `idem` are used?

src/pyk/konvert.py
```diff
@@ -452,6 +453,29 @@ def update(sentence: KSentence) -> KSentence:
     return module.let(sentences=sentences)
 
 
+def _add_constructor_atts(module: KFlatModule) -> KFlatModule:
+    """Add the constructor attribute to all symbol productions that satisfy the citeria.
+
+    Depends on the macro, anywhere and injective attributes already being sorted out.
```
Baltoli: It's interesting to me that this is how we identify constructor symbols; are there cases where something that's not a K constructor is marked as a KORE constructor? Is there a mismatch between the concepts?

src/pyk/konvert.py
```diff
@@ -476,6 +500,35 @@ def update(sentence: KSentence) -> KSentence:
     return module.let(sentences=sentences)
 
 
+def _remove_hook_atts(module: KFlatModule, *, hook_namespaces: Iterable[str] = ()) -> KFlatModule:
+    """Remove hooks attributes from symbol productions that are either 1) array hooks or 2) not built in and not activated."""
```
Baltoli, line 554: Should there be a warning emitted somewhere if we have to filter out hooks here? What's the rationale for _explicitly_ having to state the namespaces that apply?

Baltoli (approved):
Code looks good; a few queries on the stuff you've turned over here but we can discuss these in the K meeting. Merge once you get CI passing if there aren't any huge changes needed!

src/pyk/konvert.py
tothtamas28: Right now `constructor` is attached to a symbol production that does not have any of the following attributes:

```
function
assoc
comm
idem
unit
macro
anywhere
```

Source: https://github.com/runtimeverification/pyk/blob/e68db42142d1e03d23cabd4cfff9082a08798f12/src/pyk/konvert.py#L298

Is this reasonable?

src/pyk/konvert.py
tothtamas28: Opened an issue: https://github.com/runtimeverification/k/issues/3976

PR #865, opened 2024-02-07
Add `module_name` method for method name processing

### LINKED ISSUE #337: Refactor the module name generation
**Problem:**
Running a proof which has array types in its definition leads to a crash like 
```sh
 _kore_module = kflatmodule_to_kore(                                                                                                                                                                     
  File "/nix/store/s5dgc4ryqifwbdv9bwjfvxmq8kpr0l76-python3.10-pyk-0.1.613/lib/python3.10/site-packages/pyk/konvert.py", line 424, in kflatmodule_to_kore                                                   
    return Module(name=kflatmodule.name, sentences=(imports + kore_axioms))                                                                                                                                 
  File "/nix/store/s5dgc4ryqifwbdv9bwjfvxmq8kpr0l76-python3.10-pyk-0.1.613/lib/python3.10/site-packages/pyk/kore/syntax.py", line 2138, in __init__                                                         
    name = Id(name)                                                                                                                                                                                         
  File "/nix/store/s5dgc4ryqifwbdv9bwjfvxmq8kpr0l76-python3.10-pyk-0.1.613/lib/python3.10/site-packages/pyk/kore/syntax.py", line 36, in __init__                                                           
    self._check(value)                                                                                                                                                                                      
  File "/nix/store/s5dgc4ryqifwbdv9bwjfvxmq8kpr0l76-python3.10-pyk-0.1.613/lib/python3.10/site-packages/pyk/kore/syntax.py", line 42, in _check                                                             
    raise ValueError(f'Expected identifier, got: {value}')                                                                                                                                                  
ValueError: Expected identifier, got: M-TEST-OPTIMISMPORTALKONTROL-TEST-FINALIZEWITHDRAWALTRANSACTION-PAUSED-UINT256-ADDRESS-ADDRESS-UINT256-UINT256-BYTES-UINT256-BYTES32-BYTES32-BYTES32-BYTES32-BYTES[]-0
-DEPENDS-MODULE 
```
This happens because the brackets `[]` are not accepted in the module name pattern in pyk.
```python
_PATTERN_STR: ClassVar = "[a-zA-Z][0-9a-zA-Z'-]*"
```
In addition, the module name is generated by taking the name of the function and the name of all the arguments, which in case of dynamic arrays, can get pretty big.

A few suggestions:
- module names are generated here: https://github.com/runtimeverification/pyk/blob/272bc8cbcb61b82e1e0ad4fc3669812de213a936/src/pyk/proof/reachability.py#L668
- write a new function in pyk that generates a valid module name given some inputs
- refactor the module name to be `M-CONTRACT-NAME-FUNCTION-NAME-HASH-OF-ARGS` or `M-CONTRACT-NAME-FUNCTION-SELECTOR`, where `FUNCTION-SELECTOR` is the first 4 bytes of the signature hash.

palinatolmach:
Addresses https://github.com/runtimeverification/kontrol/issues/337. 

This PR adds property `module_name` to `APRProof`, which transforms the `proof.id` into a valid module name by escaping illegal characters. It also adds substitution of `[` and `]` with `bkp`.

~This PR adds static method `make_module_name` to `KFlatModule`, which transforms the `proof.id` into a valid module name by escaping illegal characters. It also adds substitution of `[` and `]` with `-bkp-`. It also adds a unit test for `make_module_name`.~


src/pyk/kast/outer.py
```diff
@@ -814,6 +815,12 @@ def is_not_actually_function(label: str) -> bool:
             prod.klabel and is_not_actually_function(prod.klabel.name)
         )
 
+    @staticmethod
+    def make_module_name(proof_id: str) -> str:
+        return 'M-' + re.sub(
+            r'[\[\]]|[_%().:,]+', lambda match: '-bkt-' if match.group(0) in ['[', ']'] else '-', proof_id.upper()
+        )
```
tothtamas28, line 822: Valid KAST module names are those that match the following regex:

```regex
#?[a-zA-Z0-9_]+(-[a-zA-Z0-9_]+)*
```

In particular, a dash is not allowed to be followed by another dash. However, e.g. for input `[`, the output of this function is `M--bkt-`, which is not valid.

src/pyk/proof/reachability.py
```diff
@@ -665,7 +664,7 @@ def _inject_module(module_name: str, import_name: str, sentences: list[KRuleLike
                 dependencies_as_rules.append(apr_subproof.as_rule(priority=20))
         circularity_rule = proof.as_rule(priority=20)
 
-        module_name = 'M-' + re.sub(r'[_%().:,]+', '-', self.proof.id.upper())
+        module_name = KFlatModule.make_module_name(self.proof.id)
```
tothtamas28, line 667: This is a KORE module name, which is validated by a different grammar:

```regex
[a-zA-Z][a-zA-Z0-9'-]*
```

For example, `a'` is valid KORE but not valid KAST. On the other hand, `_` is valid KAST but not valid KORE.

palinatolmach: Oh I see, thanks! Should I move this function out of `KFlatModule` (e.g., into `APRProof`)? Would it make it more consistent with it being a KORE module name?

src/pyk/kast/outer.py
palinatolmach: But is it valid KORE? I haven't observed any errors due to the presence of `--` (but I can of course remove it nevertheless).

src/pyk/proof/reachability.py
palinatolmach: Understood, thank you @tothtamas28 for the explanation! I don't think it's used (nor is intended to be used) anywhere else, so I extracted it into a property in `APRProof` in https://github.com/runtimeverification/pyk/pull/865/commits/ee69304a4788af1fe587c8390dcdb60ab25cb227. I opted for a property instead of a static function, as the module name is calculated based on the same proof's `id`, but if you'd recommend to make it a private function, please let me know and I'll do that instead.

anvacaru (commented):
This looks good to me but I would like someone else to review as well.

ehildenb (approved):
I'm approving to unblock, but please add some unit tests for this one (inspired by real uses ideally) so that we can add to the testing harness later as we discover more problems.

That may require making it so that this is actually a `staticmethod` of `KFlatModule` (which does the regex substitution and the `_.upper()`), and then is re-exposed as this `property` here.

palinatolmach:
Thanks @ehildenb and @anvacaru! I added (but then removed) a unit test for the static method testing the same regex: https://github.com/runtimeverification/pyk/pull/865/commits/3c478d71d75002c6971ac3664e7e083234508bd3. I'll test this PR locally on the Optimism proof to make sure it works and will merge it so we can unblock the dynamic types support in Kontrol for the engagement; and I'll open a separate PR with a unit test so we can discuss where to place it etc.

anvacaru (approved):

src/tests/unit/test_proof.py
```diff
@@ -219,6 +220,25 @@ def test_apr_proof_from_dict_heterogeneous_subproofs(proof_dir: Path) -> None:
     assert proof.dict == proof_from_disk.dict
 
 
+MODULE_NAME_TEST_DATA: Final = (
+    ('sq-bracket', 'TEST-KONTROL-TEST-UINT256-BYTES[]-0', 'M-TEST-KONTROL-TEST-UINT256-BYTESbktbkt-0'),
+    ('underscore', 'TEST_KONTROL_%)_UINT256-1', 'M-TEST-KONTROL-UINT256-1'),
```
tothtamas28, line 225: One more thing to consider: in the ideal case your mapping should be injective, otherwise name clashes might occur.

You can also weaken this requirement and say it is injective w.r.t. a reasonable subset of inputs, and the user has to come up with weird inputs to make a clash.

palinatolmach: Created an issue: https://github.com/runtimeverification/pyk/issues/867

PR #866, opened 2024-02-07
Fix claim selection process for unlabeled claims

ehildenb:
Here: https://github.com/runtimeverification/pyk/pull/842, a mistake was made where claims without labels were filtered out of the claims to be processed and returned by `KProve.get_claims`. This PR addresses that and adds tests. In particular:

- Two claims in the `multi-claim-spec.k` which were ill-formed because of dependencies  (but not tested for it) are removed.
- An ill-formed (because of dependencies) claim is added in a separate module, and a new test added which exercises that a `ValueError` is thrown when loading this module.
- A new test is added which asserts that the correct number of claims is loaded from the main module. On `master`, this test is failing because it skips claims without labels, but is fixed on this PR.
- Adjustments to how claims are filtered are made, so that it's simpler and includes all claims, even ones without explicit module names in the label.


src/tests/integration/kprove/test_get_claims.py
```diff
@@ -95,3 +95,11 @@ def test_get_claims(
             assert set(expected_graph.keys()) == set(actual_graph.keys())
             for claim_label, deps in expected_graph.items():
                 assert set(deps) == set(actual_graph[claim_label])
+
+    def test_get_claims_unlabeled(self, kprove: KProve) -> None:
+        claims = kprove.get_claims(self.KOMPILE_MAIN_FILE, self.SPEC_MODULE_NAME)
+        assert len(claims) == 9
```
tothtamas28, line 101: Please add an additional claim without a label.

src/pyk/ktool/kprove.py
```diff
@@ -364,10 +364,7 @@ def _get_claim_module(_label: str) -> str | None:
             return None
 
         all_claims = {
-            claim.label: (claim, _get_claim_module(claim.label))
-            for module in flat_module_list.modules
-            for claim in module.claims
-            if _get_claim_module(claim.label) is not None
+            claim.label: (claim, module.name) for module in flat_module_list.modules for claim in module.claims
```
tothtamas28, line 367: This I think only admits a single claim such that `claim.label is None`.

src/tests/integration/kprove/test_get_claims.py
ehildenb: There is a claim without a label, do you want 2?

tothtamas28: Yes, at least two to test my assumption below.

ehildenb: Done.

src/pyk/ktool/kprove.py
ehildenb: The issue wasn't that `claim.label` was `None`, because the frontend inserts claim labels. It was that I was using `_get_claim_module(...)` to extract the module name rather than just `module.name`, and `_get_claim_module(...)` sometimes returned `None` instead of the actual module name.

tothtamas28 (approved):

PR #870, opened 2024-02-10
Integrate the APR and APRBMC proof/prover classes into one

ehildenb:
Currently we have a completely separate class hierarchy for `APRProof` and `APRBMCProof` classes, which requires duplicates in the `*Prover` classes and in the `*NodePrinter` classes and in the `*Show` classes. This unifies the classes by making the `bmc_depth` an optional parameter of `APRProof` and tracking the `_bounded` nodes directly in `APRProof`.

- `bmc_depth: int | None` is added as a field to `APRProof`, as is the `_bounded: set[int]` set. The various utilities for manipulating the class are updated to account for this.
- The `APRProver` class is updated to check if the `bmc_depth` field is present, and if alters its `advance_pending_node` logic to check for loops and take that into account.
- The various display classes like `APRNodePrinter` and `APRSummary` are updated.


ehildenb:
- KEVM PR which contains commits needed to make this mergeable there: https://github.com/runtimeverification/evm-semantics/pull/2294
- Kontrol PR which contains commits needed to make this mergeable there: https://github.com/runtimeverification/kontrol/pull/359
- KMIR requires no changes for this to be mergeable there.

tothtamas28 (approved):
Great simplification!

nwatson22 (approved):

PR #872, opened 2024-02-12
Optimize KCFG memory usage

virgil-serbanuta:


virgil-serbanuta:
> Please add a description that roughly explains how the optimization works, and what its impacts are on performance (including of course savings in memory consumption). Thanks!

Done, I hope it's reasonable. If you have a standardized way of measuring performance, I can also run that.

ehildenb:
@virgil-serbanuta I'm curious what is taking up so much memory in teh KCFG in general. Can you guys store fewer nodes as well (increase `execute_depth` on your requests from 1k default to 10k, for example) as well?

This seems like a good move overall, I'm worried about the added complexity though. Also worried about thread safety, since we are working on a parallel prover implementation.

virgil-serbanuta:
@ehildenb Well, most of the kcfg memory is taken by the parsed wasm code, and the largest part of that is a table of functions with their type and code. This changes rarely, i.e. when there are contract calls, and for the large proof+kcfg I mentioned above there are ~10 calls between contracts. Also, the pair contract, for which I'm doing property verification right now, is rather large.

FWIW, there are 252 nodes in that large kcfg (~200M/node), most of which have 1000 steps between them. On average, the contract takes ~7 minutes between nodes, but it can take significantly longer.

I thought about using more steps between nodes, and I mentioned this in my initial discussion with @tothtamas28 . That is doable, but that would make debugging harder. The problem is that when there is a backend error (or when I stop execution, say, because I want to figure out why a split happens), I have to restart from the last valid node. With 1000 steps between nodes, I usually lose up to 7 minutes to reach the error point (plus the time needed for loading the configuration from the disk). With 10000 steps, I would lose up to 70 minutes.

ehildenb:
Thanks for the detailed explanation. I wonder if we can somehow make the storage of the KCFG more compact via something like CSE for this, by executing chunks of the code and storing that separately as a KCFG (with smaller code attached to it, just of that function), and then re-using that KCFG for the larger executions. It may be worth investigating.

The biggest concern I have here then is thread-safety. Can we make it so that this caching mechanism is thread-safe? We'd like to have multiple worker threads extending the KCFG in various places at the same time.

virgil-serbanuta:
> The biggest concern I have here then is thread-safety. Can we make it so that this caching mechanism is thread-safe? We'd like to have multiple worker threads extending the KCFG in various places at the same time.

@ehildenb I'm assuming that the kcfg object itself is/will be thread safe.

I'm not fully sure how this will be used. The simplest way of solving the problem is to add a simple lock around the `bottom_up_with_summary` call in `optimize()`, and I did exactly that.

However, there are other solutions with higher performance, and I implemented one here: https://github.com/runtimeverification/pyk/blob/optimize-kcfg-fancy/src/pyk/kast/optimizer.py#L17-L76 . It is not fully finished, but I could try to move it to the current PR if you think that kcfgs need to handle non-trivial amounts of concurrent updates.

src/pyk/kcfg/kcfg.py
```diff
@@ -230,6 +233,16 @@ def __init__(self, cfg_dir: Path | None = None) -> None:
...
+        self._optimizer = KInnerOptimizer()
+        if optimize_memory:
+
+            def optimize_cterm(cterm: CTerm) -> CTerm:
+                new_config = self._optimizer.optimize(cterm.config)
+                return CTerm(new_config, cterm.constraints)
+
+            self._optimize_cterm = optimize_cterm
+        else:
+            self._optimize_cterm = lambda x: x
```
tothtamas28, line 245: Consider the following refactoring (sketch).

```python3
_nodes: MutableMapping[int, Node]

def __init__(self, optimize_memory: bool):
    if optimize_memory:
        self._nodes = OptimizedNodeStore()
    else:
        self._nodes = {}
```

where the optimization logic is abstracted as

```python3
class OptimizedNodeStore(MutableMapping[int, Node]):
    ...
```

Probably most of `optimizer.py` will become part of this class.

Then `_optimize_cterm` does not have to be called explicitly for each node addition, and it'll be easier to test / compare / switch between optimizations.

virgil-serbanuta: I added an OptimizedNodeStore in the KCFG class, and the code does, indeed, look nicer. I also moved most of the optimizer.py code to the `OptimizedNodeStore` class.

However, I'm not sure what to do about the lock I added to the KInnerOptimizer, because the KCFG class seems to have its own lock, and that may (or may not) make my lock redundant, depending on how it is supposed to be used in multithreaded code. Should I rely on all multithreaded KCFG usage occurring in `with my_kcfg: ...` blocks?

src/pyk/kcfg/kcfg.py
```diff
@@ -200,8 +202,101 @@ def with_single_target(self, target: KCFG.Node) -> KCFG.NDBranch:
         def edges(self) -> tuple[KCFG.Edge, ...]:
             return tuple(KCFG.Edge(self.source, target, 1, ()) for target in self.targets)
 
+    class OptimizedNodeStore(MutableMapping[int, Node]):
```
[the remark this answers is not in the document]
virgil-serbanuta: Done

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,133 @@
...
+        id = self.value_to_id.get(value)
+        if id is not None:
+            return id
+        id = len(self.values)
+        self.value_to_id[value] = id
+        self.values.append(value)
+        return id
+
+
+@dataclass(eq=True, frozen=True)
```
tothtamas28, line 37: `eq` is `True` by default:

```suggestion
@dataclass(frozen=True)
```

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def build(self, klabels: list[KLabel], terms: list[KInner]) -> KInner:
+        return KSequence(tuple(terms[child] for child in self.children))
+
+
+class OptimizedNodeStore(MutableMapping[int, KCFG.Node]):
+    __nodes: dict[int, KCFG.Node]
+    __optimized_terms: CachedValues[OptimizedKInner]
+    __klabels: CachedValues[KLabel]
+    __terms: list[KInner]
```
tothtamas28, line 76:
```suggestion
    _nodes: dict[int, KCFG.Node]
    _optimized_terms: CachedValues[OptimizedKInner]
    _klabels: CachedValues[KLabel]
    _terms: list[KInner]
```

(Also for methods.)

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,133 @@
...
+@final
+@dataclass(eq=True, frozen=True)
+class OptimizedKSequence(OptimizedKInner):
+    children: tuple[int, ...]
+
+    def build(self, klabels: list[KLabel], terms: list[KInner]) -> KInner:
+        return KSequence(tuple(terms[child] for child in self.children))
+
+
+class OptimizedNodeStore(MutableMapping[int, KCFG.Node]):
```
tothtamas28, line 72: Please move this to the top of the module.

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,133 @@
...
+from ..kast.inner import KApply, KSequence, KToken, KVariable, bottom_up_with_summary
+from .kcfg import KCFG
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from ..kast.inner import KInner, KLabel
+
+
+A = TypeVar('A')
```
tothtamas28, line 19:
```python3
from collections.abc import Hashable
```

then

```suggestion
A = TypeVar('A', bound=Hashable)
```

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+@dataclass
+class CachedValues(Generic[A]):
```
tothtamas28, line 23: Consider a singular class name. Also, let's make this private. Then it can even be called something as concise as `_Cache` or  `_Index` (as the class looks like an index).

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,133 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+    from ..kast.inner import KInner, KLabel
+
+
+A = TypeVar('A')
+
+
+@dataclass
```
tothtamas28, line 22: `dataclass` does not seem to add much for this class.

```suggestion
```

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,133 @@
...
+class CachedValues(Generic[A]):
+    value_to_id: dict[A, int] = field(default_factory=dict)
+    values: list[A] = field(default_factory=list)
```
tothtamas28, line 25: Consider making these private:

```suggestion
    _value_to_id: dict[A, int] = field(default_factory=dict)
    _values: list[A] = field(default_factory=list)
```

Then it can have two methods, say, `put` (currently `cache`) and `get`. The index can then be passed to `OptimizedKInner.build`.

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,133 @@
...
+        if id is not None:
+            return id
+        id = len(self.values)
+        self.value_to_id[value] = id
+        self.values.append(value)
+        return id
+
+
+@dataclass(eq=True, frozen=True)
+class OptimizedKInner:
```
tothtamas28, line 38: Consider something more concise, e.g. `OptInner` (and similarly for subclasses).

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,133 @@
...
+    @abstractmethod
+    def build(self, klabels: list[KLabel], terms: list[KInner]) -> KInner:
+        ...
+
+
+@final
+@dataclass(eq=True, frozen=True)
+class SimpleOptimizedKInner(OptimizedKInner):
```
tothtamas28, line 46: Maybe `OptSimple` or `OptBasic`?

ehildenb:
I think if we'll have an optimized presentation of KCFGs, we should just have it on by default so it's used everywhere rather than letting the user decide to use it or not.

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,133 @@
...
+        return KApply(klabels[self.label], tuple(terms[child] for child in self.children))
+
+
+@final
+@dataclass(eq=True, frozen=True)
+class OptimizedKSequence(OptimizedKInner):
+    children: tuple[int, ...]
+
+    def build(self, klabels: list[KLabel], terms: list[KInner]) -> KInner:
+        return KSequence(tuple(terms[child] for child in self.children))
```
ehildenb, line 69: Where is `klabels` used here?

ehildenb: Also, is there a reason we are using `list` instead of `dict` here?

virgil-serbanuta: It's not used, but it's part of the interface (see the base class). Should I do anything about it (e.g. put an underscore before, or something similar)?

src/pyk/kcfg/store.py
virgil-serbanuta: Done

src/pyk/kcfg/store.py
virgil-serbanuta: Why is this a good idea?

virgil-serbanuta:
> I think if we'll have an optimized presentation of KCFGs, we should just have it on by default so it's used everywhere rather than letting the user decide to use it or not.

I think it's on by default (the default value for the `optimize_memory` argument should be true for the KCFG constructor and the static methods that build KCFGs)

ehildenb:
@virgil-serbanuta can you mark the conversations as resolved that are taken care of? Then we can look again with fresh eyes.

src/pyk/kcfg/store.py
virgil-serbanuta: I added a 'get', but didn't change the `cache` name.

virgil-serbanuta:
> @virgil-serbanuta can you mark the conversations as resolved that are taken care of? Then we can look again with fresh eyes.

Done

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,136 @@
...
+class OptimizedNodeStore(MutableMapping[int, KCFG.Node]):
+    _nodes: dict[int, KCFG.Node]
+    _optimized_terms: _Cache[_OptInner]
+    _klabels: _Cache[KLabel]
+    _terms: list[KInner]
+
+    _lock: threading.Lock
+
+    def __init__(self) -> None:
+        super().__init__()
```
tothtamas28, line 31:
```suggestion
```

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,136 @@
...
+        new_config = self._optimize(old_cterm.config)
+        new_constraints = tuple(self._optimize(c) for c in old_cterm.constraints)
+        new_node = KCFG.Node(node.id, CTerm(new_config, new_constraints))
+        self._nodes[key] = new_node
+
+    def __delitem__(self, key: int) -> None:
+        del self._nodes[key]
+
+    def __iter__(self) -> Iterator[int]:
+        return self._nodes.__iter__()
```
tothtamas28, line 53:
```suggestion
        return iter(self._nodes)
```

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def _cache(self, term: _OptInner) -> int:
+        id = self._optimized_terms.cache(term)
+        assert id <= len(self._terms)
+        if id == len(self._terms):
+            self._terms.append(term.build(self._klabels, self._terms))
+        return id
+
+    def _cache_klabel(self, label: KLabel) -> int:
+        return self._klabels.cache(label)
```
tothtamas28, line 83: Consider inlining this.

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+class _Cache(Generic[A]):
+    def __init__(self) -> None:
+        self._value_to_id: dict[A, int] = {}
+        self._values: list[A] = []
```
tothtamas28, line 89:
```suggestion
class _Cache(Generic[A]):
    _value_to_id: dict[A, int]
    _values: list[A] 

    def __init__(self) -> None:
        self._value_to_id = {}
        self._values = []
```

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def cache(self, value: A) -> int:
+        id = self._value_to_id.get(value)
+        if id is not None:
+            return id
+        id = len(self._values)
+        self._value_to_id[value] = id
+        self._values.append(value)
+        return id
```
tothtamas28, line 98: For consistency with `get`:

```suggestion
    def cache(self, value: A) -> int:
        idx = self._value_to_id.get(value)
        if idx is not None:
            return idx
        idx = len(self._values)
        self._value_to_id[value] = idx
        self._values.append(value)
        return idx
```

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def get(self, idx: int) -> A:
+        return self._values[idx]
+
+
+@dataclass(frozen=True)
+class _OptInner:
+    @abstractmethod
+    def build(self, klabels: _Cache[KLabel], terms: list[KInner]) -> KInner:
+        ...
```
tothtamas28, line 108:
```suggestion
class _OptInner(ABC):
    @abstractmethod
    def build(self, klabels: _Cache[KLabel], terms: list[KInner]) -> KInner:
        ...
```

src/tests/unit/kcfg/test_store.py
```diff
@@ -0,0 +1,83 @@
...
+)
+
+
+@pytest.mark.parametrize('term1,term2', EQUAL_TEST_DATA, ids=count())
+def test_use_cached(term1: KInner, term2: KInner) -> None:
+    # When
+
+    cached_values: _Cache[KInner] = _Cache()
+
+    id1 = cached_values.cache(term1)
```
tothtamas28, line 37:
```suggestion
    # Given
    cached_values: _Cache[KInner] = _Cache()

    # When
    id1 = cached_values.cache(term1)
```

src/tests/unit/kcfg/test_store.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+@pytest.mark.parametrize('term1,term2', NOT_EQUAL_TEST_DATA, ids=count())
+def test_not_use_cached(term1: KInner, term2: KInner) -> None:
+    # When
+
+    cached_values: _Cache[KInner] = _Cache()
+
+    id1 = cached_values.cache(term1)
+    id2 = cached_values.cache(term2)
```
tothtamas28, line 60:
```suggestion
    # Given
    cached_values: _Cache[KInner] = _Cache()

    # When
    id1 = cached_values.cache(term1)
    id2 = cached_values.cache(term2)
```

tothtamas28 (commented):
I added a few more nitpicks, but apart from those looks good to me.

As far as I'm concerned, the only thing left before merging is testing this in Kontrol. Once you consider the PR ready, let me know, and I'll run the Kontrol integration tests with this version.

src/pyk/kcfg/store.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def __getitem__(self, key: int) -> KCFG.Node:
+        return self._nodes[key]
+
+    def __setitem__(self, key: int, node: KCFG.Node) -> None:
+        old_cterm = node.cterm
+        new_config = self._optimize(old_cterm.config)
+        new_constraints = tuple(self._optimize(c) for c in old_cterm.constraints)
+        new_node = KCFG.Node(node.id, CTerm(new_config, new_constraints))
+        self._nodes[key] = new_node
```
ehildenb, line 46: I don't fully understand how this is working....

When I select a given node with something like `kcfg.node(node_id)`, it appears to me here that what I'll get back is the _optimized_ version. Where does it actually reconstruct the original `CTerm`, so that I can pass it to the rest of the pyk library for manipulation?

virgil-serbanuta: Hmm... I'm not sure I understand the question, I hope this answers it:

"Optimized" has two meanings here.

First, objects derived form "_OptInner" are "optimized". An _OptInner object, as used here, is equivalent to a KInner in the context of a OptimizedNodeStore object. However, _OptInner is supposed to be private to this file, so it is not directly related to CTerms (it is just a helper to make the optimization efficient).

Second, new_node contains a CTerm that is optimized (in the sense that it shares memory with other CTerms), but which is equal (i.e. the `==` operator will evaluate to True) to the CTerm inside the node passed to this function. This means that the original CTerm was reconstructed two lines above, in the `CTerm(new_config, new_constraints)` call.

This optimization works like this (I'm ignoring a few details):

Before optimization, a KInner term is decomposed in a bottom-up way into _OptInner objects. These objects are stored in a _Cache, which ensures that there is a single copy of each _OptInner object. Additionally, for each _OptInner object, the OptimizedNodeStore holds a KInner object created from that _OptInner. The subterms of this KInner also correspond to _OptInner objects in the cache.

As an example, let's say that we have a term that looks like this: `f(t1, g(t2, t1'))`, where `t1 == t1'`. This is decomposed bottom-up, there will be a single _OptInner object (o1) corresponding to `t1` and `t1'`, and the OptimizedNodeStore will hold a KInner object `T1` corresponding to `o1` such that `T1 == t1 == t1'`. Then , when reconstructing the original term, it will look like this: `f(T1, g(T2, T1))`. Note that this uses the same object (`T1`) for both occurrences (i.e. it shares memory), while the original term used two identical objects (`t1` and `t1'`).

In other words, if the same KInner (K1) occurs in different parts of a KInner object (K2) created from an _OptInner, then there is a single object for K1 which is shared for all occurrences in K2. This optimization is less important, but we get it for free.

Next, when another KInner term (e.g., from a different CTerm) is optimized, it is also decomposed and reconstructed as above. This means that this KInner shares the objects for the common subterms with the previous KInner terms that were optimized. This is the optimization that matters.

The end result is that an optimized KCFG is the same as an unoptimized one (i.e., '==' returns True), but the optimized version shares a single object for all occurrences of a (sub)term, while the unoptimized one has a copy for each occurrence.

ehildenb: Ah I see. So it's really about using Python's object references in a smart way, by basically making sure that for each immutable `KInner` we ever see, we only store one copy of it, and any new ones we see that are the same are adjusted to point at the same copy as the first one we saw, instead of being their own copy?

ehildenb:
This looks good to me, @tothtamas28 if you can do final sign-off then I think we're good. Awesome idea @virgil-serbanuta !

tothtamas28 (approved):
Tested in `kontrol`, LGTM.

PR #875, opened 2024-02-13
Make advance proof functionality generic

ehildenb:
~Blocked on: https://github.com/runtimeverification/pyk/pull/878~

In an attempt to make the prover interfaces simpler, we are making it so that `advance_proof` is a generic method among all `*Prover` classes. This will allow us to have a generic way to call subproofs and advance them, then call superproofs and advance them, which will allow simplifying some downstream logic a bit.

- Makes it so that `mlEqualsTrue` and `mlEqualsFalse` can take sort arguments for their outer sort.
- Moves all the `APRProof` specific arguments of `APRProof.advance_proof` to the constructor instead.
- Adds an `abstractmethod Proof.step_proof(...)` which should take one step of executing a proof to completion.
- Adds an `abstractmethod Proof.can_progress` which lets you know if it's possible to make progress on a given proof.
- Implements generic `Proof.advance_proof(...)` which handles calling `step_proof` in a loop until either `max_iterations` is reached, the proof can no longer progress, or if `fail_fast` trigger hits. It also handles writing proof data each iteration.
- Adjusts all the tests to the new interface.


ehildenb:
KEVM is updated here to handle this new interface: https://github.com/runtimeverification/evm-semantics/pull/2294

nwatson22 (approved):

PR #876, opened 2024-02-13
Add timeouts to CI pipeline

ehildenb:
I made a PR here https://github.com/runtimeverification/pyk/pull/875 which caused CI to jump from 8 minutes to an infinite loop. We shouldn't let CI machines be taken up because of mistakes in PRs.

This also adds the pytest-timeout module so we can do `make cov-integration COV_ARGS='--timeout 300'` to timeout individual tests.


pyproject.toml
```diff
@@ -36,6 +36,7 @@ pep8-naming = "*"
 pytest = "*"
 pytest-cov = "*"
 pytest-mock = "*"
+pytest-timeout = "*"
```
tothtamas28, line 39: How well does this plugin work together with `pytest-xdist`? How is a timeout reported?

ehildenb: It seems to work fine with it, I've been running many tests in parallel with timeouts locally on KEVM and Kontrol for figuring out which tests take longer than a certain threshold for disabling them for a while. `pytest-timeout` has already been added to those repos.

Here is an example output:

image

And further in the output:

image


And further in the output:

image


So it seems to be pretty seamless!

ehildenb: This was with running 20 tests at a time.

tothtamas28 (approved):

PR #878, opened 2024-02-14
Overhaul ImpliesProof class hierarchy

ehildenb:
While working on https://github.com/runtimeverification/pyk/pull/875, I found many places in `ImpliesProof` class hierarchy that we can make things simpler. In general:

- `mlEqualsTrue` and `mlEqualsFalse` are given sorts they can use for the outer sort of the equality.
- The various renaming of `antecedent` and `consequent` in subclasses are removed.
- The `ImpliesProof` gets its own `dict` and `from_dict` methods, and the subclasses and reuse it.
- `advance_proof` no longer needs to be overloaded in the subclasses.
- Various really simple setter methods are removed in favor of direct setting of paramaters in the proofs.
- The now-useless classes `EqualityProver` and `RefutationProver` are removed.

Essentially, after this, `EqualityProof` and `RefutationProof` are just "smart constructors" for `ImpliesProof`, with different names for the antecedent/consequent and specific ways of constructing the antecedent/consequent. The key difference here is that `EqualityProver` and `RefutationProver` were tracking their own proof state before (by storing and manipulating `_antecedent_kast` and `_consequent_kast` themselves directly), and now they rely only on the `*Proof` objects to store state.


ehildenb:
Changes to KEVM to make it build successfully with this change are in this branch: https://github.com/runtimeverification/evm-semantics/pull/2294

nwatson22 (approved):

PR #881, opened 2024-02-14
Added proof execution time to APRProof

lisandrasilva:
This PR records how long a proof takes to execute in the `APRProof` class so it can be accessed in Kontrol to display in the output message and in the `xml` report.


src/pyk/proof/reachability.py
```diff
@@ -433,11 +445,14 @@ def write_proof_data(self) -> None:
         ensure_dir_path(self.proof_dir)
         ensure_dir_path(self.proof_subdir)
         proof_json = self.proof_subdir / 'proof.json'
-        dct: dict[str, list[int] | list[str] | bool | str | int | dict[int, str] | dict[int, list[dict[str, Any]]]] = {}
+        dct: dict[
+            str, list[int] | list[str] | bool | float | str | int | dict[int, str] | dict[int, list[dict[str, Any]]]
+        ] = {}
```
tothtamas28, line 450:
```suggestion
        dct: dict[str, Any] = {}
```

src/pyk/proof/reachability.py
```diff
@@ -53,6 +54,7 @@ class APRProof(Proof, KCFGExploration):
     logs: dict[int, tuple[LogEntry, ...]]
     circularity: bool
     failure_info: APRFailureInfo | None
+    execution_time: float
```
tothtamas28, line 57: Let's make this private. Also, consider a shorter name:

```suggestion
    _exec_time: float
```

src/pyk/proof/reachability.py
```diff
@@ -160,6 +164,12 @@ def prune(self, node_id: NodeIdLike, keep_nodes: Iterable[NodeIdLike] = ()) -> l
             self._bounded.discard(nid)
         return pruned_nodes
 
+    def update_execution_time(self, execution_time: float) -> None:
```
tothtamas28, line 167:
```suggestion
    def add_exec_time(self, execution_time: float) -> None:
```

src/pyk/proof/reachability.py
```diff
@@ -160,6 +164,12 @@ def prune(self, node_id: NodeIdLike, keep_nodes: Iterable[NodeIdLike] = ()) -> l
...
+        self.execution_time += execution_time
+
+    def reset_execution_time(self) -> None:
```
tothtamas28, line 170:
```suggestion
    def reset_exec_time(self) -> None:
```

src/pyk/proof/reachability.py
tothtamas28: You'll then need a way to read the value, for that you can add a read-only property:

```python
def exec_time(self) -> float:
    return self._exec_time
```

tothtamas28:
Eventually, such metadata should be provided by a generic runner of provers (but I understand that right now that's not an option).

src/pyk/proof/reachability.py
lisandrasilva: Makes sense. Thank you!

src/pyk/proof/reachability.py
lisandrasilva: Actually, I changed this function, it is now `set_exec_time(self, exec_time: float)` and updates the `self._exec_time` to the value of the received argument.

tothtamas28 (approved):

PR #884, opened 2024-02-14
Adds new tests for node refutation

lucasmt:
Adds tests in `test_refute_node.py` for the fixes from #877.

* Initialization of the prover is factored out to its own function `build_prover`.
* `test_apr_proof_read_node_refutations` tests that the node refutations can be successfully and correctly read from the `.json` file (before the above PR this would raise an error).
* `test_apr_proof_refute_node_no_successors` tests that a node can only be refuted if it has no successors.


src/tests/integration/proof/test_refute_node.py
```diff
@@ -195,3 +190,60 @@ def test_apr_proof_refute_node(
...
+        claim_id = 'split-int-succeed'
+
+        prover = self.build_prover(kprove, proof_dir, kcfg_explore, spec_file, spec_module, claim_id)
+
+        # When
+        prover.advance_proof(max_iterations=1)
+        frontier_nodes = prover.proof.pending
+        assert prover.proof.status == ProofStatus.PENDING
+
+        assert len(frontier_nodes)
```
ehildenb, line 212:
```suggestion
        assert len(frontier_nodes) == 2
```

Perhaps more specific check?

lucasmt: I copied this from `test_apr_proof_unrefute_node` above, so maybe I should make the change there as well?

src/tests/integration/proof/test_refute_node.py
```diff
@@ -195,3 +190,60 @@ def test_apr_proof_refute_node(
...
+        prover.advance_proof(max_iterations=1)
+        frontier_nodes = prover.proof.pending
+        assert prover.proof.status == ProofStatus.PENDING
+
+        assert len(frontier_nodes) == 2
+        frontier_node = frontier_nodes[0]
+        prover.proof.refute_node(frontier_node)
+
+        proof_from_file = APRProof.read_proof_data(proof_dir, prover.proof.id)
+        refutation_id = prover.proof.get_refutation_id(frontier_node.id)
```
ehildenb, line 217: I think we can test the correct JSON deserialization by just using the `dict` and `from_dict` methods:

```
proof_reread = APRProof.from_dict(prover.proof.dict)
```

ehildenb (approved):

src/tests/integration/proof/test_refute_node.py
lucasmt: We could, but there's a couple of caveats:

1. `from_dict` and `read_proof_data` have separate implementations, so if `from_dict` passes the test it doesn't mean that `read_proof_data` does (and vice versa). So maybe we should either test both or refactor the code so that one calls the other.
2. One of the bugs I fixed happens specifically when we read the JSON from file. The `node_refutations` dict has the key as an int, but when this is saved to a file the key turns into a string. This was causing a problem when reading it back from the file, because the code expected an int instead of a string. So this bug would not be exercised if we just deserialized from `prover.proof.dict` instead of from file.

PR #886, opened 2024-02-15
Self-contained proof steps for EqualityProver and APRProver

nwatson22:
Refactor to `advance_proof` intended to split `step_proof` stages `get_steps` (figure out which steps are available), `exec` (standalone chunk of work which has all the information necessary to perform it contained in `ProofStep`, and `commit`, which modifies `proof` with the results of the proof steps. Intended to be an incremental step toward parallelizing individual proofs. Pulled largely from https://github.com/runtimeverification/pyk/pull/727
Adds:
- `ProofStep` and subclasses `APRProofStep` and `ImpliesProofStep`
- `StepResult` and subclasses `APRProofResult` and `ImpliesProofResult`, for returning proof update information from `ProofStep.exec`
- subclasses of `APRProofResult` `APRProofExtendResult` for if the current node should be extended and `APRProofSubsumeResult` for if it should be subsumed into the target
- Change `Proof.advance_proof` to, instead of calling `step` in a loop, calls `get_steps`, and then calls `exec` and then `commit` on each of these steps.

Tested against kontrol test suite as well.


ehildenb:
Let's make the following changes:

- `get_steps() -> list[ProofStep]` and `commit(ProofStepResult) -> None`, are now methods of the `*Proof` classes, not of the `*Prover` classes.
- The `step_proof(ProofStep) -> list[ProofStepResult]` method is part of the `*Prover` class.
- The `ProofStep` and `ProofStepResult` do _not_ hold a `KCFGExplore` (that stays in `*Prover` classes), and _only_ holds the minimum data for `step_proof(...)` to be able to operate.

ehildenb:
With this design, we may be able to make it so that `*Prover` does not even hold a `*Proof` in its hands.

palinatolmach:
@nwatson22 is anything blocking this PR? Is it still relevant?

nwatson22:
Closing in favor of 
- https://github.com/runtimeverification/k/pull/4250
- https://github.com/runtimeverification/k/pull/4246
- https://github.com/runtimeverification/k/pull/4256

PR #887, opened 2024-02-16
Add support for non-deterministic branches in `refute_node`

lucasmt:
Blocked on: #884

Previously, `refute_node` was implemented in a way that only supports refuting nodes after a `Split`, not an `NDBranch`. This PR removes that restriction.

* Replaces the `pre_constraints` and `last_constraint` members of the `RefutationProof` class with a single `refuted_constraints` list, so that refutations can be easily built for `NDBranch` as well (which lack a branching condition). The implication is then written as `refuted_constraints -> bottom`, instead of `pre_constraints -> last_constraint = false`.
* Removes the checks in `construct_node_refutation` preventing refuting nodes from a non-deterministic branch.
* Adds a test that refuting a node from a non-deterministic branch succeeds.


lucasmt:
Closing this, as the discussion above suggest that it's preferable to `refute-node` from a `Split` rather than an `NDBranch`, and that a better solution is to give the user a way to turn an `NDBranch` into a `Split` via a `split-node` functionality.

PR #888, opened 2024-02-16
Allow universal binding in ImpliesProof

ehildenb:
The Haskell backend does not allow passing implications in that have unbound variables in teh consequent (variables that are not present in the antecedent). Until now, we've always bound these existentially.

However, on a `RefutationProof` or an `EqualityProof`, it makes more sense to bind them universally. For example, in a `RefutationProof`, you are asserting that a given branch condition goes to `false`, given the path condition up to that point. If there is _no_ path condition, then the variables in the branch condition will be unbound, and so you'll get a proof that says `Exists FV . pred(FV) == false` where `pred` is the final branch condition, and `FV` are the free variables in that branch condition. This is just checking that the branch condition _could_ be `false`, not that it's _always_ `false`, which is the users intention.

This PR:

- Renames the various files associated with Implication proofs from `equality` to `implies`.
- Adds a field to `KCFGExplore.cterm_implies(... bind_universally: bool ...)` which allows the user to specify whether they want existential or universal binding of the free consequent variables.
- Updates to `ImpliesProof` hierarchy:
  - Renames files/modules from `*equality*.py` to `*implies*.py`.
  - Adds `ImpliesProof.{from_dict,dict}`, and re-uses them in subclasses `EqualityProof` and `RefutationProof`.
  - Removes the fields of `RefutationProof` turning them into properties instead.
  - Has `RefutationProof` take as input the boolean constraints for the antecedent, not the ML preds, and handles turning them into mlPreds directly. The consequent was already handled this way.
  - Adds field `bind_universally` to `ImpliesProof`, and sets it to `true` for both subclasses `RefutationProof` and `EqualityProof`.
- Adjustments to tests:
  - Adds direct tests of `ImpliesProof`, which can exercise both existential and universal binding of the consequent variables.
  - Adjusts the tests of node refutation to send boolean predicates for the refutation, not ml preds.
  - Disables a test of `EqualityProof` that was only passing because of the default existential binding.


lucasmt:
I have a draft PR (#887) that would merge `pre_constraints` and `last_constraint` in `RefutationProof` into just one `refuted_constraints` list, and change the implication from `pre_constraints -> last_constraint = false` to `refuted_constraints -> bottom`. This would mean no more unbound variables in the consequent. But I assume the changes in this PR would still be relevant for other cases.

nwatson22 (approved):

PR #905, opened 2024-02-21
Remove KPrint class from KCFGExplore

ehildenb:
Blocked on: https://github.com/runtimeverification/evm-semantics/pull/2302

While looking at how to implement a `pyk prove ...` CLI interface that uses the RPC prover (not the legacy prover), for the BOB conference presentation, I wanted to import the `KCFGExplore` into the `KProve` class to provide the same top-level interface that can be accessed on the CLI. Unfortunately, this creates a circular import because `KCFGExplore` already holds a `KPrint` in hand.

The uses of `KPrint` in `KCFGExplore` break into 2 categories:

- Uses that are `kast_to_kore` and `kore_to_kast`. For these uses, we can just call the underlying `konvert.kast_to_kore` and `konvert.kore_to_kast` instead, and have the application crash rather than falling back to `kast` binary.
- Uses taht are to do pretty-printing, which we can imitate by just building a `PrettyPrinter` directly (a class defined in the `kast` submodule).

So this PR removes `KPrint` from `KCFGExplore`, in favor of more direct manipulation of the underlying stuffs. There are a lot of places where this change needs to be propagated, so I'll open a PR into KEVM and Kontrol with the needed updates.

This PR also renames `KProve.prove` to `KProve.prove_legacy`, to indicate that it's using the legacy `kprove ...` executable.


ehildenb:
We should wait on this until the various downstream issues blocking pyk updates to KEVM are resolved: https://github.com/runtimeverification/evm-semantics/pull/2302

PR #909, opened 2024-02-22
Move kcfg commit logic to KCFG class

nwatson22:
- `KCFGExplore.extend_kcfg()` replaced by `KCFG.extend()`
- Logging of `ExtendResult`s moved earlier to when they are created (before, they were logged when applied, but this is no longer possible with application logic (`extend()`) living in `KCFG`.
- `KCFGExplore.extend()` removed and its logic moved up into `advance_pending_node()`


src/pyk/kcfg/kcfg.py
```diff
@@ -1031,3 +1067,59 @@ def read_cfg_data(cfg_dir: Path, id: str) -> KCFG:
             cfg.create_ndbranch(source_id, target_ids)
 
         return cfg
+
+
+class ExtendResult(ABC):
```
ehildenb, line 1072: Can we rename this to `KCFGExtendResult`, or `KCFGStep`? Something that indicates it's for the KCFG class.

ehildenb (approved):

src/pyk/kcfg/kcfg.py
nwatson22: Done.

PR #912, opened 2024-02-23
Add cterm module for cterm.py

ehildenb:
We've had `cterm.py` at the top-level of the pyk library. This moves it into a submodule in preparation for refactors to `KCFGExplore` to factor out basic cterm manipulation commands from more complicated KCFG manipulation commands.

- `cterm.py` is moved to `cterm/cterm.py`. Imports are adjusted.
- `cterm/__init__.py` is added, and setup so that almost no imports elsewhere need to change.


nwatson22 (approved):

ehildenb:
Don't merge untir https://github.com/runtimeverification/evm-semantics/pull/2302 goes through.

PR #914, opened 2024-02-23
Add test failing simplify endpoint for ceil condition generation

ehildenb:
In https://github.com/runtimeverification/evm-semantics/pull/2302, we are failing a test because it's not generating the appropriate ceil conditions. This calls out a bit more specifically what is failing, by asserting how many ceil conditions should be generated there.

### DIFF
--- a/src/tests/integration/proof/test_mini_kevm.py
+++ b/src/tests/integration/proof/test_mini_kevm.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


nwatson22 (commented):
Is the test `ensures-constraint-simplification` supposed to fail now? Or is 2 actually the incorrect number of ceil conditions?

ehildenb:
The test passes on master, but fails on the other PR. This PR is to make the test more specific so it's clear what's failing.

PR #916, opened 2024-02-24
Improve default argument handling

### LINKED ISSUE #885: Improve handling of command line options
We should fix the codebase across pyk, kontrol, and evm-semantics to reduce the number of times default values for command options need to be specified, as per https://github.com/runtimeverification/kontrol/pull/298#pullrequestreview-1838295387. Ideally we'd also want to be able to move the instantiation of the classes analogous to [ProveOptions](https://github.com/runtimeverification/kontrol/blob/master/src/kontrol/options.py) up as far as possible so as to minimize passing around of individual options, so instead of looking like
```
def exec_prove(
    foundry_root: Path,
    max_depth: int = 1000,
    max_iterations: int | None = None,
    reinit: bool = False,
    tests: Iterable[tuple[str, int | None]] = (),
    include_summaries: Iterable[tuple[str, int | None]] = (),
    workers: int = 1,
    break_every_step: bool = False,
    break_on_jumpi: bool = False,
    break_on_calls: bool = True,
    break_on_storage: bool = False,
    break_on_basic_blocks: bool = False,
    break_on_cheatcodes: bool = False,
    bmc_depth: int | None = None,
    bug_report: BugReport | None = None,
    kore_rpc_command: str | Iterable[str] | None = None,
    use_booster: bool = True,
    smt_timeout: int | None = None,
    smt_retry_limit: int | None = None,
    smt_tactic: str | None = None,
    failure_info: bool = True,
    counterexample_info: bool = False,
    trace_rewrites: bool = False,
    auto_abstract_gas: bool = False,
    run_constructor: bool = False,
    fail_fast: bool = False,
    port: int | None = None,
    maude_port: int | None = None,
    use_gas: bool = False,
    summary_path: Path | None = None,
    **kwargs: Any,
) -> None:
```
exec prove could look more like
```
def exec_prove(
    options: ProveOptions
) -> None:
```

argparse has a feature where you can pass an object as a `Namespace` to `parse_args()` which it will populate with the args and their values as fields, so it would be nice if we could just use that out of the box, but it doesn't seem powerful enough for what we want to do (multiple different classes for different subcommands). The problem is that we would like the Options classes to be specific to a specific command, like `ProveOptions` for `prove`, `RunOptions` for `run`, etc. but to know which of these commands has been invoked you first need to call `parse_args`. It would be nice if there was a way to split the argument parsing into two stages where during the first it gets the subcommand and during the second it gets the options for that subcommand.

Possibly we could have routing logic like
```
    args = cli_parser.parse_args()
    if args.command.lower() == 'prove':
       options = ProveOptions(args)
    if args.command.lower() == 'run':
       options = RunOptions(args)
    ...
    executor_name = 'exec_' + args.command.lower().replace('-', '_')
    if executor_name not in globals():
        raise AssertionError(f'Unimplemented command: {args.command}')

    execute = globals()[executor_name]
    execute(options)
```
and then in the constructor for each Options class that takes a dict, that would be where we set the default values if the entry is missing from the dict.

@palinatolmach also pointed out that we want to be able to, in some cases, override the default values of inherited options for semantics-specific Options subclasses, so this should also be built with that in mind. 

Adding better default argument handling in #916 (in progress).

nwatson22:
Closes https://github.com/runtimeverification/pyk/issues/885.

Overhauls the way command line argument parsing and parameter passing is done.
When building a pyk-based command-line tool, for each subcommand of the tool, extend `class Command` . This subclass should contain all information about what arguments are accepted by that command (through providing its own arguments or inheriting from other `Options` classes), the default values of those arguments, the name of the command, and the help string for the command, as static fields.  The values of those options for a specific invocation of that command are stored as non-static fields of a `Command`. The `Command` subclass contains the code that runs when that command is called in `exec()`. In addition, default values of arguments inherited from other `Options` classes can be overridden in the subclass.

The `CLI` class manages the tool's CLI options. It is constructed by passing in the name of every command subclass to be included in the tool. It can then build the whole `ArgumentParser` for the tool and process the arguments to instantiate a new `*Command` of the correct type and with the correct arguments.

Advantages:
- All information about a subcommand is consolidated into one place
- Default values specified in only one place
- Commands only have to be listed once, when instantiating `CLI`
- Setting up an argument parser with all subcommands is done automatically.
- Routing of requested command to its associated execution function is done automatically.


src/pyk/cli/args.py
```diff
@@ -1,18 +1,290 @@
...
+            return KoreToJsonOptions(args)
+        case 'json-to-kore':
+            return JsonToKoreOptions(args)
+        case _:
+            raise ValueError('Unrecognized command.')
+
+
+class Options(object):
+    def __init__(self, args: dict[str, Any]) -> None:
+        # Get defaults from all subclasses that define them, preferring the most specific class
```
nwatson22, line 50:
```suggestion
        # Get defaults from this and all superclasses that define them, preferring the most specific class
```

src/pyk/cli/args.py
```diff
@@ -1,18 +1,290 @@
...
+
+        # Overwrite defaults with args from command line
+        _args = defaults | args
+
+        for attr, val in _args.items():
+            self.__setattr__(attr, val)
+
+    @classmethod
+    def all_args(cls: Type[Options]) -> ArgumentParser:
+        # Collect args from all superclasses
```
nwatson22, line 66:
```suggestion
        # Collect args from this and all superclasses
```

src/pyk/cli/args.py
```diff
@@ -1,155 +1,732 @@
...
+
+
+class Command(LoggingOptions):
+    @staticmethod
+    @abstractmethod
+    def name() -> str:
+        ...
+
+    @staticmethod
+    @staticmethod
```
ehildenb, line 134:
```suggestion
    @staticmethod
```

src/pyk/cli/args.py
```diff
@@ -1,155 +1,732 @@
...
+        return pyk_args
+
+
+class Options:
+    def __init__(self, args: dict[str, Any]) -> None:
+        # Get defaults from this and all superclasses that define them, preferring the most specific class
+        defaults: dict[str, Any] = {}
+        mro = type(self).mro()
+        mro.reverse()
+        for cl in mro:
```
ehildenb, line 88:
```suggestion
        for cl in type(self).mro().reverse():
```

Can probably make this more compact, below as well.

src/pyk/cli/args.py
```diff
@@ -1,155 +1,732 @@
...
+        parser.add_argument('--debug', action='store_true', help='Debug output.')
+
+
+class Command(LoggingOptions):
+    @staticmethod
+    @abstractmethod
+    def name() -> str:
+        ...
+
+    @staticmethod
```
ehildenb, line 133: I feel like `Command` and `Options` class hierarchies should be in separate modules. Could do `cli/command.py` and `cli/options.py`.

ehildenb: Could also have `Command` and `CLI` class hierarchies both in `cli/cli.py` perhaps...

src/pyk/cli/args.py
nwatson22: Fixed.

src/pyk/cli/args.py
nwatson22: The reason it's written like that is because `reverse()` only updates in place, but I've changed it to use `reversed` so it can be done in one line.

src/pyk/cli/args.py
nwatson22: Moved `Command` and `CLI`, and all subclasses of `Command` into their own module `cli.py`.

src/pyk/cli/cli.py
```diff
@@ -0,0 +1,432 @@
...
+    def generate_command(self, args: dict[str, Any]) -> Command:
+        command = args['command'].lower()
+        for cmd_type in self.commands:
+            if cmd_type.name() == command:
+                return cmd_type(args)
+        raise ValueError(f'Unrecognized command: {command}')
+
+    # Generate the parsers for all commands
+    def add_parsers(self, base: _SubParsersAction) -> _SubParsersAction:
+        for cmd_type in self.commands:
```
ehildenb, line 66: `CLI` and `Command` seem generic, and everything else is specific to pyks CLI here. So I guess it may make sense to have them in separate modules too.

Also, I'm not sure I understand why `LoggingOptions` is the superclass of `Command`. It seems that if we always want those logging options included, then we can do that directly in `CLI` or `Command` class, but don't need an explicit `LoggingOptions` class to enable that.

ehildenb: I guess I would move all the pyk-specific things to a new module `cli/pyk.py` or something, it's the CLI for pyk.

src/pyk/__main__.py
```diff
@@ -1,345 +1,55 @@
...
-    )
-    coverage_args.add_argument('coverage_file', type=FileType('r'), help='Coverage file to build log for.')
-    coverage_args.add_argument('-o', '--output', type=FileType('w'), default='-')
-
-    pyk_args_command.add_parser('kore-to-json', help='Convert textual KORE to JSON', parents=[k_cli_args.logging_args])
+    cli_parser = cli.create_argument_parser()
+    args = cli_parser.parse_args()
 
-    pyk_args_command.add_parser('json-to-kore', help='Convert JSON to textual KORE', parents=[k_cli_args.logging_args])
+    command = cli.generate_command({key: val for (key, val) in vars(args).items() if val is not None})
```
ehildenb, line 49: What is this dictionary comprehension?  Could that happen inside `generate_command(...)` instead, because we'll have to do it for every time we use it? It looks like we're doing the same here: https://github.com/runtimeverification/evm-semantics/pull/2319/files#diff-8ef99db40b2010c0b478f54603fda2b277c332347b8945df0e83bb686a6eafc1R101

src/pyk/__main__.py
```diff
@@ -1,345 +1,55 @@
...
+    command.exec()
 
-    return pyk_args
+    logging.basicConfig(level=loglevel(command), format=LOG_FORMAT)
```
ehildenb, line 52: Do logging options work correctly like this? Do we need to set the logging options before `command.exec()`?

nwatson22: No, good catch, it needs to be before. Fixed.

src/pyk/__main__.py
nwatson22: I've added some new methods to `Command` that can do this step for the user, so now all they have to do in the simplest use case is call the `CLI` constructor and then call `cli.get_and_exec_command()`.

src/pyk/cli/cli.py
nwatson22: `LoggingOptions` is now no longer intrinsically part of `Command`s, and I've moved the pyk-specific `Command`s into `cli/pyk.py`.

ehildenb:
It looks pretty good overall, maybe @tothtamas28 would like to have another look. If we have a working version for each downstream semantics, I'd say let's get it merged! KEVM/Kontrol, KWasm, KMIR.

src/pyk/cli/args.py
```diff
@@ -1,155 +1,311 @@
...
+        for attr, val in _args.items():
+            self.__setattr__(attr, val)
+
+    @classmethod
+    def all_args(cls: type[Options]) -> ArgumentParser:
+        # Collect args from this and all superclasses
+        parser = ArgumentParser(add_help=False)
+        mro = cls.mro()
+        mro.reverse()
+        for cl in mro:
```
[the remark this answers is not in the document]
ehildenb: I like the idea of this refactor, but let's get that merged in a follow-up PR. @nwatson22 can you pull this into an issue?

src/pyk/cli/cli.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # Return an instance of the correct Options subclass by matching its name with the requested command
+    def generate_command(self, args: dict[str, Any]) -> Command:
+        command = args['command'].lower()
+        for cmd_type in self.commands:
+            if cmd_type.name() == command:
+                if issubclass(cmd_type, Options):
+                    return cmd_type(args)
+                else:
+                    return cmd_type()
```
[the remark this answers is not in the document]
nwatson22: Options is an object that holds a set of options as fields, which could be a command or just grouped options that can be inherited by another Options or Command. Command that's not an Options would be a command without any options.

src/pyk/cli/args.py
```diff
@@ -1,155 +1,311 @@
...
-        return args
-
-    @cached_property
-    def bug_report_args(self) -> ArgumentParser:
-        args = ArgumentParser(add_help=False)
-        args.add_argument(
-            '--bug-report',
-            type=bug_report_arg,
-            help='Generate bug report with given name',
+class Options:
```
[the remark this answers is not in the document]
nwatson22: Moved.

src/pyk/cli/args.py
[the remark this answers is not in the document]
nwatson22: New issue is here: https://github.com/runtimeverification/pyk/issues/954

ehildenb:
Going to merge this, and get it pushed through to all downstream repos, then can work on next batch of refactorings. Mostly because I have other CLI changes that I want to go through, and want ot try out the new interface for making those changes.

After we get this pushed through, we can discuss the next refactoring/design that we'd like to pursue here: https://github.com/runtimeverification/pyk/issues/954

ehildenb (approved):

PR #917, opened 2024-02-26
Minimize constraints when constructing refutation proofs

lucasmt:
This PR introduces the following changes:
* Refactors `remove_useless_constraints` to take a list of constraints rather than a `CTerm`, and reimplements the original version as `remove_useless_constraints_from_cterm`.
* Use the new `remove_useless_constraints` to restrict the constraints used to build a refutation proof to only those that have a dependency on the most recent branch.


src/pyk/cterm/cterm.py
```diff
@@ -338,26 +338,38 @@ def apply(self, cterm: CTerm) -> CTerm:
...
+        prev_len_used_vars = len(used_vars)
+        for c in constraints:
             if c not in new_constraints:
                 new_vars = free_vars(c)
                 if any(v in used_vars for v in new_vars):
                     new_constraints.append(c)
                     used_vars.extend(new_vars)
         used_vars = list(set(used_vars))
-    return CTerm(cterm.config, new_constraints)
+    return new_constraints
```
ehildenb, line 372: I guess this method can live in `kast/manip.py` instead of in `cterm.py`.

ehildenb (approved):

ehildenb:
It would also be good to add a test of this. In `test_refute_node.py`, you could add a test where the original proof has contraints over both an `X` and `Y`, and then it has a branch on variable `X` only, and you do a node refutation on the branch and see that the refutation only contains constraints over variable `X` instead of over them both.

lucasmt:
Updated to move `remove_useless_constraints` to `manip.py` and add a test in `test_refute_node.py`, as per @ehildenb's suggestions above.

src/pyk/cterm/cterm.py
```diff
@@ -338,25 +339,15 @@ def apply(self, cterm: CTerm) -> CTerm:
...
-        for c in cterm.constraints:
-            if c not in new_constraints:
-                new_vars = free_vars(c)
-                if any(v in used_vars for v in new_vars):
-                    new_constraints.append(c)
-                    used_vars.extend(new_vars)
-        used_vars = list(set(used_vars))
+    initial_vars = free_vars(cterm.config) + list(keep_vars)
+    new_constraints = remove_useless_constraints(cterm.constraints, initial_vars)
     return CTerm(cterm.config, new_constraints)
```
ehildenb, line 351: Make this a static method of `CTerm`, as `CTerm.remove_useless_constraints(CTerm, Iterable[str]) -> CTerm`.

ehildenb (approved):
Looks great! Nice test!

src/pyk/cterm/cterm.py
lucasmt: If we're moving it inside `CTerm`, wouldn't it be better to make it a non-static method? Just make the `CTerm` argument `self`? `new_cterm = cterm.remove_useless_constraints(...)`

Issue #686, opened 2023-10-18
`KCFG.replace_node` is broken


tothtamas28:
Reproduction script (tested on: 2d745013c6a92a49ec745345768bc95a74836235):

```python3
from pathlib import Path
from pprint import pprint

from pyk.cterm import CTerm
from pyk.kcfg import KCFG
from pyk.kcfg.show import KCFGShow
from pyk.ktool.kompile import kompile
from pyk.ktool.kprint import KPrint
from pyk.prelude.ml import mlTop, mlBottom


main_file = Path('test.k')
main_file.write_text('module TEST endmodule')
definition_dir = kompile(main_file, syntax_module='TEST')
kprint = KPrint(definition_dir)

cfg = KCFG()
n1 = cfg.create_node(CTerm(mlTop()))
n2 = cfg.create_node(CTerm(mlTop()))
cfg.create_edge(n1.id, n2.id, depth=1)
cfg.replace_node(n2.id, CTerm(mlBottom()))

pprint(vars(cfg))

show = KCFGShow(kprint)
for line in show.pretty(cfg):
    print(line)
```

Output:

```
{'_aliases': {},
 '_covers': {},
 '_created_nodes': {1, 2},
 '_deleted_nodes': set(),
 '_edges': {1: {2: Edge(source=Node(id=1,
                                    cterm=CTerm(config=KApply(label=KLabel(name='#Top',
                                                                           params=(KSort(name='GeneratedTopCell'),)),
                                                              args=()),
                                                constraints=())),
                        target=Node(id=2,
                                    cterm=CTerm(config=KApply(label=KLabel(name='#Top',
                                                                           params=(KSort(name='GeneratedTopCell'),)),
                                                              args=()),
                                                constraints=())),
                        depth=1)}},
 '_lock': <unlocked _thread.RLock object owner=0 count=0 at 0x7fdf012c2100>,
 '_ndbranches': {},
 '_node_id': 3,
 '_nodes': {1: Node(id=1,
                    cterm=CTerm(config=KApply(label=KLabel(name='#Top',
                                                           params=(KSort(name='GeneratedTopCell'),)),
                                              args=()),
                                constraints=())),
            2: Node(id=2,
                    cterm=CTerm(config=KApply(label=KLabel(name='#Bottom',
                                                           params=(KSort(name='GeneratedTopCell'),)),
                                              args=()),
                                constraints=()))},
 '_splits': {},
 '_stuck': set(),
 '_vacuous': set(),
[...]

PR #920, opened 2024-02-26
Fix KCFG.replace_node functionality

ehildenb:
Fixes: https://github.com/runtimeverification/pyk/issues/686

Currently, `KCFG.replace_node` is broken because `KCFG.Successor` stores directly a `KCFG.Node` for the sources and targets. So when we replace a node in the store currently, we are only replacing the node itself, but not iterating through the successors and replacing occurances of the same node there.

This PR uses a pretty brute-force way to do that (to be able to make progress on storing Node attributes on nodes directly instead of as separate lists), and later we can make a better way of storing this information: https://github.com/runtimeverification/pyk/issues/919.

### DIFF
--- a/src/tests/unit/test_kcfg.py
+++ b/src/tests/unit/test_kcfg.py
@@ -394,6 +394,29 @@ def test_vacuous() -> None:
     assert cfg.vacuous, node(3)
 
 
+def test_replace_node() -> None:
+    # Given
+    d = {
+        'nodes': node_dicts(4),
+        'edges': edge_dicts((1, 2), (2, 3)),
+    }
+
+    cfg = KCFG.from_dict(d)
+    cfg.replace_node(2, term(5))
+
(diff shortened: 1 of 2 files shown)


src/tests/unit/test_kcfg.py
```diff
@@ -394,6 +394,29 @@ def test_vacuous() -> None:
...
+    assert node is not None
+    assert node.cterm == term(5)
+
+    first_edge = cfg.edge(1, 2)
+    assert first_edge is not None
+    assert first_edge.target.cterm == term(5)
+
+    second_edge = cfg.edge(2, 3)
+    assert second_edge is not None
+    assert second_edge.source.cterm == term(5)
```
ehildenb, line 417: Before this PR, it reported back `term(2)` as teh source and target node of the edge.

tothtamas28 (approved):

PR #922, opened 2024-02-27
cache prior loops discovered in bmc

ovatman:
This pr adds a `prior_loops_cache` that holds a dictionary of {loop : prior same loops} that has been discovered so far. 
During `bmc_depth` bound checking the new loop node's cache is populated with the previously discovered same nodes.
The cache is also written/read as proof data. 
Closes: https://github.com/runtimeverification/kontrol/issues/283


src/pyk/kcfg/kcfg.py
```diff
@@ -265,6 +265,7 @@ def replace_target(self, node: KCFG.Node) -> KCFG.NDBranch:
     _aliases: dict[str, int]
     _lock: RLock
     cfg_dir: Path | None
+    _shortest_paths_cache: dict[tuple[NodeIdLike, NodeIdLike], tuple[Successor, ...]]
```
ehildenb, line 268:
```suggestion
    _shortest_paths_cache: dict[tuple[int, int], tuple[Successor, ...]]
```

ehildenb: When user calls external functions, the external function must resolve the `NodeIdLike` to an `int`.

src/pyk/kcfg/kcfg.py
```diff
@@ -887,10 +889,16 @@ def prune(self, node_id: NodeIdLike, keep_nodes: Iterable[NodeIdLike] = ()) -> l
...
+            return self._shortest_paths_cache[(source_node_id, target_node_id)]
         paths = self.paths_between(source_node_id, target_node_id)
         if len(paths) == 0:
             return None
-        return sorted(paths, key=(lambda path: KCFG.path_length(path)))[0]
+
+        self._shortest_paths_cache[(source_node_id, target_node_id)] = sorted(
+            paths, key=(lambda path: KCFG.path_length(path))
+        )[0]
+        return self._shortest_paths_cache[(source_node_id, target_node_id)]
```
ehildenb, line 902: Caching things like this will become invalid if the user modifies KCFGs that touch these nodes. For example, removes a node that is in this shortest path between.

ehildenb: Can maybe look at all the `def remove_` and `def discard_`...

src/pyk/proof/reachability.py
```diff
@@ -80,6 +80,7 @@ class APRProof(Proof, KCFGExploration):
     circularity: bool
     _exec_time: float
     error_info: Exception | None
+    prior_loops_cache: dict[NodeIdLike, list[NodeIdLike]]
```
ehildenb, line 83:
```suggestion
    prior_loops_cache: dict[int, list[int]]
```

Please make the necessary adjustments so that we're not storing node aliases or node strings here, just integers. That means, when updating the `prior_loops_cache`, we need to use `KCFG.resolve(node_id)`, to get out the actual integer referring to the node.

ehildenb:
The logic of `APRProof.prune` also needs to be updated to take into account `prior_loops_cache`. Any nodes that are pruned should also be removed the the cache, as keys or as values.

src/pyk/proof/reachability.py
ehildenb: Yeah, if we already have the `id` from `node.id`, we don't need `KCFG.resolve(...)`.

src/pyk/proof/reachability.py
```diff
@@ -717,18 +735,21 @@ def step_proof(self) -> Iterable[StepResult]:
...
-                if not (
-                    self.proof.kcfg.zero_depth_between(_pl, curr_node.id)
-                    or any(self.proof.kcfg.zero_depth_between(_pl, pl) for pl in prior_loops)
-                ):
-                    prior_loops.append(_pl)
+
+            prior_loops = []
+            for succ in reversed(self.proof.shortest_path_to(curr_node.id)):
+                if self.kcfg_explore.kcfg_semantics.same_loop(succ.source.cterm, curr_node.cterm):
+                    if succ.source.id in self.proof.prior_loops_cache.keys():
```
ehildenb, line 742:
```suggestion
                    if succ.source.id in self.proof.prior_loops_cache:
```

`_in_` is implemented directl yfor dictionaries.

src/pyk/proof/reachability.py
```diff
@@ -480,6 +489,9 @@ def read_proof_data(proof_dir: Path, id: str) -> APRProof:
         node_refutations = {
             kcfg._resolve(int(node_id)): proof_id for node_id, proof_id in proof_dict['node_refutations'].items()
         }
+        prior_loops_cache = {
+            loop_node_id: list(same_loops) for (loop_node_id, same_loops) in proof_dict['loops_cache'].items()
+        }
```
ehildenb, line 494: This will cause a breakage in reading existing proofs from disk because they won't have the `loops_cache` key. Also, we can just use the dictionary directly, instead of doing any processing on it I think:

```suggestion
        prior_loops_cache = proof_dict.get('loops_cache', {})
```

ehildenb (approved):

ehildenb:
@ovatman I approved, but please don't merge yet. Run these tests:

- Run a simple proof with a loop and `--bmc-depth` current Kontrol master, and save the proof.
- Modify your kontrol to use this version of pyk, so it's using the new format, and make sure it can read in the old format that was stored.
- Check the newly written proof data on disk and make sure it has the correct new field `loops_cache`, with the correct contents (this likely requires runinng the proof out a few more steps to populate the cache).

PR #924, opened 2024-02-27
Explicit result type for CTermSymbolic.implies

ehildenb:
We currently have `CTermExecute` to wrap up the result value from `CTermSymbolic.execute`, this adds a similar result type for `CTermSymbolic.implies`.

- `CTermImplies` is added, which contains either a succeeded implication (`CSubst | None`), or contains failure information about the implication.
- The routine `CTermSymbolic.implies` is updated to return an instance of this object instead of just the `CSubst | None`. The logic for populating the failure information that is expected by `KCFGExplore.implication_failure_reason` is inlined into this method.
- The routine `KCFGExplore.implication_failure_reason` is updated to receive this object, then do the post-processing needed to pretty-print the failure information.
- The tests of `KCFGExplore.implication_failure_reason` are updated. For some reason, they were actually incorrect, something that should have been reporting as properly subsumed was reporting as not subsumed before. Those tests started failing after these changes, and I updated the tests appropriately.


src/tests/integration/proof/test_imp.py
```diff
@@ -151,7 +153,7 @@ def same_loop(self, c1: CTerm, c2: CTerm) -> bool:
             mlAnd(
                 [
                     mlEqualsTrue(KApply('_<Int_', [KVariable('A', 'Int'), KToken('1', 'Int')])),
-                    mlEqualsTrue(KApply('_<Int_', [KVariable('B', 'Int'), KToken('1', 'Int')])),
+                    mlEqualsTrue(KApply('_<Int_', [KVariable('B', 'Int'), KToken('2', 'Int')])),
                 ]
             ),
```
ehildenb, line 158: This test actually should have been allowing subsumption before, but was reporting that it should fail. The consequent _was_ a more general state than the antecedent. This change fixes it, now the consequent is less general.

src/tests/integration/proof/test_imp.py
```diff
@@ -179,7 +182,7 @@ def same_loop(self, c1: CTerm, c2: CTerm) -> bool:
...
         ),
```
ehildenb, line 188: Same here, the consequent is now actually more general than the antecedent.

nwatson22 (approved):

PR #925, opened 2024-02-27
Fix circular import

nwatson22:
#915 introduced a circular import crash which is triggered when executing `pyk rpc-kast` `rpc.py > kprove.py > cterm > symbolic.py > rpc.py`

Moves `KoreExecLogFormat` to `kore/rpc.py`


src/pyk/cterm/__init__.py
```diff
@@ -1,2 +1,3 @@
 from .cterm import CSubst, CTerm, build_claim, build_rule
-from .symbolic import CTermSymbolic
+
+#  from .symbolic import CTermSymbolic
```
nwatson22, line 3:
```suggestion
```

ehildenb (approved):

PR #928, opened 2024-02-27
Remove KPrint object from KCFGExplore

ehildenb:
Currently, the `KCFGExplore` object holds both a `KPrint` and `CTermSymbolic`. But it really only needs `KPrint` for the `pretty_print` functionality, which only really relies on the `PrettyPrinter` class in the `kast/pretty` module. So this PR reduces that dependency a bit.

- Places where `KCFGExplore` needed a `KDefinition` now reuse the one held by `CTermSymbolic`.
- `KCFGExplore._pretty_printer` is added for building a `PrettyPrinter` directly. `KCFGExplore.pretty_print` is added to expose the pretty printing functionality of that `PrettyPrinter`.
- Tests and other files are updated.


src/pyk/testing/_kompiler.py
```diff
@@ -253,11 +253,11 @@ def semantics(self, definition: KDefinition) -> KCFGSemantics:
     def kcfg_explore(
         self,
         cterm_symbolic: CTermSymbolic,
-        kprint: KProve,
+        kprint: KPrint,
```
tothtamas28, line 256: This no longer needs to be a `KPrintTest`.

src/pyk/kcfg/explore.py
```diff
@@ -40,25 +41,29 @@
...
         id: str | None = None,
     ):
-        self.kprint = kprint
         self.cterm_symbolic = cterm_symbolic
         self.kcfg_semantics = kcfg_semantics if kcfg_semantics is not None else DefaultSemantics()
         self.id = id if id is not None else 'NO ID'
 
+    @cached_property
+    def _pretty_printer(self) -> PrettyPrinter:
+        return PrettyPrinter(self.cterm_symbolic._definition)
```
tothtamas28, line 62: Does this handle the case when we need to customize the symbol table?

ehildenb: No. But the pretty printer in the `KCFGExplore` is just being used for printing out some log and error messages. Specifically:

- When building `implication_failure_reason`, it's used to print out what is almost 100% going to be boolean predicates (so likely default printer will work fine).
- When getting branches during KCFG exploration, it's used to log the branch conditions (boolean conditions).

I think long term, all printing functionality should be removed from this class and done at a higher level.

src/pyk/testing/_kompiler.py
ehildenb: Done.

tothtamas28 (approved):

tothtamas28:
Please do not merge until runtimeverification/evm-semantics#2302 goes through.

PR #932, opened 2024-02-28
Refactor KCFG loading capabalities

ehildenb:
We want to move towards having a KCFG store (https://github.com/runtimeverification/pyk/issues/919). This PR is a step in that direction:

- All of the `KCFG` elements (`Node` and `Successor`) implement their own `from_dict` functionality. The `Successor` need to be passed in teh `_nodes` mapping in order to construct themselves properly.
- The `KCFG.read_cfg_data` is refactored to make use of existing `KCFG.from_dict`, instead of re-implementing.
- A new method `KCFG.add_successor` is added for directly adding new successors, instead of for each individual edge type. The various `KCFG.create_{Successor}` methods are refactored to use this.
- The `KCFG.from_dict` directly creates the needed `Succesor` directly, and then uses `KCFG.add_successor` instead of the individual creation methods which are used as helpers for passing in less data.

This is to prepare us for storing just a single `_successors: dict[int, KCFG.Successor]` mapping later, without having to make major changes to the code for loading/unloading KCFGs.

### DIFF
--- a/src/pyk/kcfg/kcfg.py
+++ b/src/pyk/kcfg/kcfg.py
@@ -468,26 +468,19 @@ class KCFG(Container[Union['KCFG.Node', 'KCFG.Successor']]):
         cfg = KCFG(optimize_memory=optimize_memory)
 
         for node_dict in dct.get('nodes') or []:
-            node_id = node_dict['id']
-            cterm = CTerm.from_dict(node_dict['cterm'])
-            node = KCFG.Node(node_id, cterm)
+            node = KCFG.Node.from_dict(node_dict)
             cfg.add_node(node)
 
         max_id = max([node.id for node in cfg.nodes], default=0)
         cfg._node_id = dct.get('next', max_id + 1)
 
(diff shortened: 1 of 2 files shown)


src/pyk/kcfg/kcfg.py
```diff
@@ -77,6 +81,11 @@ def replace_target(self, node: KCFG.Node) -> KCFG.Successor:
         def to_dict(self) -> dict[str, Any]:
             ...
 
+        @staticmethod
+        @abstractmethod
+        def from_dict(dct: dict[str, Any], nodes: MutableMapping[int, KCFG.Node]) -> KCFG.Successor:
```
tothtamas28, line 86:
```suggestion
        def from_dict(dct: dict[str, Any], nodes: Mapping[int, KCFG.Node]) -> KCFG.Successor:
```

tothtamas28 (approved):

src/pyk/kcfg/kcfg.py

PR #933, opened 2024-02-29
Add `to_claim` method to `RefutationProof`

lucasmt:
This PR adds a `to_claim` method to `RefutationProof` which returns a claim representing the refutation. This claim is of the form
```
claim ['claim-id']: LAST_CONSTRAINT => false
  requires PRE_SPLIT_CONSTRAINTS [label(claim-id)]
```
where `LAST_CONSTRAINT` is the branch condition immediately preceding the refuted node and `PRE_SPLIT_CONSTRAINTS` is the set of path constraints that depend (directly or indirectly) on the variables of `LAST_CONSTRAINT`.

It also makes the following refactors to support this implementation:
* Refactor `apply_existential_substitutions` to take configuration and constraints separately, rather than as a conjoined term, and return them separately as well. This allows avoiding `split_config_and_constraints`, which requires the config to be wrapped in a cell.
* Refactor `build_claim` and `build_rule` to take configuration and constraints separately, rather than packed in a `CTerm`, and creates separate `cterm_build_claim` and `cterm_build_rule` wrappers for the `CTerm` version.


src/pyk/proof/implies.py
```diff
@@ -368,6 +368,17 @@ def pretty(self, kprint: KPrint) -> Iterable[str]:
...
+        init_config = KApply('<k>', self.last_constraint)
+        init_constraints = self.pre_constraints
+        init_cterm = CTerm(init_config, init_constraints)
+
+        final_config = KApply('<k>', FALSE)
+        final_constraints: list[KInner] = []
+        final_cterm = CTerm(final_config, final_constraints)
+
+        return build_claim(claim_id, init_cterm, final_cterm)
+
```
ehildenb, line 381: I would prefer if we built a functional claim instead of a semantic claim. Basically don't include the wrapper `KApply('<k>' ...)`. When KEVM reads in a functional claim, it should be able to build instead an `EquationalProof`, which has a different simpler routine for discharging.

lucasmt: The problem is that `build_claim` calls `split_config_and_constraints` and `apply_existential_substitutions`, which assume there is a configuration wrapped in a cell. I figured that since when we write claims for new lemmas we wrap them in `<k>` cells anyway this shouldn't make a big difference.

lucasmt: But if it's better to omit the `<k>` cell I can refactor `apply_existential_substitutions` to take and return the config and constraints already split.

ehildenb: Well, it may also be good to refactor `build_rule` and `build_claim`. Most of what they do is variable renaming to make it a valid printable claim, so we could make a specific routine for that which looks like:

```
def build_rule(lhs: KInner, rhs: KInner, requires: Iterable[KInner] = (), ensures: Iterable[KInner] = (), keep_vars: Iterable[str] = ()) -> tuple[KRule, Subst]:
    ...
```

And it basically does everything that `build_rule` currently does (build the minimized rewrite between lhs and rhs, do variable name adjustments, convert ML predicates to boolean, etc...), and returns the basic data needed for building a `KRule` or a `KClaim`. Then you can rename `build_rule` in `cterm.py` to `cterm_build_rule`, and have it reuse this inner routine, which you can also use for the above.

PR #903, opened 2024-02-21
Separate execution of proof steps from modification of proof state

nwatson22:
- `step_proof` no longer modifies proof, only produces list of `ProofResult`
- `advance_proof` now retrieves `ProofResult`s from `step_proof` and applies them using `Proof.commit()`.
- `ImpliesProof` and `APRProof` reworked to follow this scheme
- `APRProofResult`, `APRProofExtendResult`, `APRProofSubsumeResult`, `APRProofBoundedResult` for different outcomes of stepping an `APRProof`

PR #936, opened 2024-02-29
Subsumption checking fix

nwatson22:
Bug introduced by #903


src/pyk/proof/reachability.py
```diff
@@ -725,9 +725,11 @@ def step_proof(self) -> Iterable[StepResult]:
                 return [APRProofBoundedResult(curr_node.id)]
 
         is_terminal = self.kcfg_explore.kcfg_semantics.is_terminal(curr_node.cterm)
-        target_is_terminal = (self.proof.target in self.proof._terminal,)
 
-        if target_is_terminal and self.always_check_subsumption:
+        target_is_terminal = self.proof.target in self.proof._terminal
+        check_subsume = (target_is_terminal and is_terminal) or (not target_is_terminal)
+
+        if check_subsume:
```
PetarMax, line 732: Note for future: this means we can do away with `always_check_subsumption`, as well as the simplification cache.

PetarMax (approved):

src/pyk/proof/reachability.py
ehildenb: `always_check_subsumption` is used elsewhere too though. At least it was, maybe that functionality was lost too....

PetarMax: I've checked the `pyk` repo and it doesn't seem to be used anymore. I could be wrong.

PR #939, opened 2024-03-01
Maintaining information on nodes being terminal

PetarMax:
This PR ensures that information about nodes being terminal, which was lost in the recent refactoring, is preserved.


src/pyk/proof/reachability.py
```diff
@@ -723,18 +722,25 @@ def step_proof(self) -> Iterable[StepResult]:
...
         target_is_terminal = self.proof.target in self.proof._terminal
-        check_subsume = (target_is_terminal and is_terminal) or (not target_is_terminal)
+
+        terminal_result = [APRProofTerminalResult(node_id=curr_node.id)] if is_terminal else []
+
+        # Subsumption should be checked if and only if the target node
+        # and the current node are either both terminal or both not terminal
+        check_subsume = is_terminal == target_is_terminal
 
         if check_subsume:
```
ehildenb, line 735:
```suggestion
        if is_terminal == target_is_terminal:
```

This intermediate value is only used once, and the expression is not large.

ehildenb (approved):

ehildenb:
It would be good to add a test to this before merge, since we're fixing a regression that actually happened, we can expect it to happen again on future refactorings.

You can do so in the directory `src/tests/integration/proof`, by adding a new file `test_simple.py`. You can copy the `test_goto.py` file, and then adjust it to point at `simple-proofs.k` and `simple-proofs-spec.k` (instead of the `goto*` files). You can find the spec files at `src/tests/integration/k-files/{simple-proofs.k,simple-proofs-spec.k}`.

Add a `KCFGSemantics` object (adjust tho `GotoSemantics` object), to dectect the state `<k> e </k>` and `<k> e ~> . </k>` as `is_terminal.

Then make some specifications in `simple-proofs-spec.k` as:

```
claim <k> 
```

nwatson22 (approved):

nwatson22:
Added regression test and confirmed it fails correctly on master.

src/tests/integration/k-files/simple-proofs.k
```diff
@@ -36,13 +36,13 @@ module SIMPLE-PROOFS
...
 
     syntax KItem ::= Step
-    syntax Step ::= "a" | "b" | "c" | "d" | "e"
+    syntax Step ::= "a" | "b" | "c" | "d" | "e" | "f"
 
     rule [a.b]: <k> a => b ... </k>
     rule [b.c]: <k> b => c ... </k>
-    rule [b.d]: <k> b => d ... </k>
-    rule [c.e]: <k> c => e ... </k>
+    rule [c.e]: <k> c => d ... </k>
```
ehildenb, line 43: Why was this change made? Likely it's setup this way for another test in the test-suite, we just need to add the one more state to the end `f`.

nwatson22: I just did it to simplify the test since I looked and thought I saw simple-proofs.k wasn't being used anywhere but on further inspection I see that it is. I've reverted this change, which required also a slight modification to the test (now checks all predecessor nodes to the target), and retested against master.

ehildenb:
Let's make sure to resolve all conversations before merging @PetarMax !

PetarMax:
Isn't there an option that doesn't allow merging before all conversations are resolved? I've seen it in other PRs and if that's our philosophy, then it should be enabled by default.

PR #951, opened 2024-03-06
Fix failure info not being set in advance_proof on done proof

nwatson22:
https://github.com/runtimeverification/kontrol/issues/413


palinatolmach (approved):

lisandrasilva:
I ran it locally, and it fixes the issue with the XML report!

PR #957, opened 2024-03-08
Generate the `format` attribute in `_module_to_kore`

tothtamas28:
Blocked on:
* ~runtimeverification/k#4086~
* ~runtimeverification/k#4089~


src/pyk/kast/outer.py
```diff
@@ -293,6 +294,62 @@ def argument_sorts(self) -> list[KSort]:
         """Return the sorts of the non-terminal positions of the productions."""
         return [knt.sort for knt in self.non_terminals]
 
+    @cached_property
+    def is_prefix(self) -> bool:
+        """
+        The production is of the form `t* "(" (n ("," n)*)? ")"`.
+
+        Here, `t` is a terminal other than `"("`, `","` or `")"`, and `n` a non-terminal.
+        Example: `Int ::= "mul" "(" Int "," Int ")"`
```
Baltoli, line 303: Is this documented somewhere else, or is the comment based on your own experiments / analysis of the compiler's current behaviour?

Baltoli (approved):

src/pyk/kast/outer.py
tothtamas28: The latter, based on this function in the Frontend to be precise: https://github.com/runtimeverification/k/blob/ba7c3247377ff3beb63e3ab8907863b0c483002b/kore/src/main/scala/org/kframework/definition/outer.scala#L806

PR #959, opened 2024-03-09
Crash when using dependencies as rules

nwatson22:
Sometimes when executing a proof using dependencies using the booster backend, a crash occurs:
```
 ValueError: Expected identifier, got: Rule#Var'QuesUnds'GENERATEDCOUNTER'Unds'CELL'Unds'6de8d71b
 ```
 The booster appends `Rule#` to the variables in some internalized rules which causes variable names like
 ```
 Rule#Var'QuesUnds'GENERATEDCOUNTER'Unds'CELL'Unds'6de8d71b
 ```
to show up in the RPC result. This causes the crash because the `#` doesn't match 
```
    _PATTERN_STR: ClassVar = "[a-zA-Z][0-9a-zA-Z'-]*"
```
from `syntax.py`

Proposed fix:
replace with
```
    _PATTERN_STR: ClassVar = "[a-zA-Z][0-9a-zA-Z#'-]*"
```

### CODE REVIEW

File: src/pyk/kore/syntax.py [hunk 1]
```diff
@@ -29,7 +29,7 @@
 class Id:
     value: str
 
-    _PATTERN_STR: ClassVar = "[a-zA-Z][0-9a-zA-Z'-]*"
+    _PATTERN_STR: ClassVar = "[a-zA-Z][0-9a-zA-Z#'-]*"
```
tothtamas28, line 32: This is not how KORE identifiers are defined: https://github.com/runtimeverification/haskell-backend/blob/4d5d1fe232c41400e09a2db0d8cb4989a5dd5832/docs/kore-syntax.md?plain=1#L94

If the Booster returns a variable with `#` in its name, then it's a bug in the Booster, and issue should be opened there.

nwatson22: Thanks for clarifying. I've opened the issue [here](https://github.com/runtimeverification/hs-backend-booster/issues/547).

nwatson22:
Closing as not a pyk issue

PR #961, opened 2024-03-09
Fix NDBranch typing.

RaoulSchaffranek:
The kcfg module contains two classes named `NDBranch`. When using this module from Simbolik - mypy confused the types. This PR disambiguates some type annotations to aid mypy.


tothtamas28 (approved):

tothtamas28:
The root of the problem should also be addressed, please open a PR or an issue for renaming the colliding `KCFGExtendResult` subclasses.

PR #962, opened 2024-03-11
Set `failure_info` if `fail_fast` is on

palinatolmach:
Unblocks https://github.com/runtimeverification/kontrol/pull/429.

This PR sets `failure_info` for a failed proof even if the `fail_fast` option is on and the proof terminates early.

It also adds a `fail_fast` parameter and a corresponding test claim to `test_failure_info` (but not [`test_failure_info_recomputed_on_proof_reinit`](https://github.com/runtimeverification/pyk/blob/a442ceaebeef0872de26749d1513fc37d9675206/src/tests/integration/proof/test_imp.py#L1070) for the reason mentioned in the [comment](https://github.com/runtimeverification/pyk/pull/962#issuecomment-1988023070)).


palinatolmach:
There's a strange thing that I noticed while trying to add a test for `failure_info` to [`test_failure_info_recomputed_on_proof_reinit`](https://github.com/runtimeverification/pyk/blob/a442ceaebeef0872de26749d1513fc37d9675206/src/tests/integration/proof/test_imp.py#L1070), where the proof is written to and then read from disk. 

The claim, which terminates early, has 1 terminal and 1 pending node when it's written to the disk, but after it's reloaded and is used to initialize `APRProver`, it has 2 terminal and 0 pending node — in other words, it looks like the pending node gets executed during `APRProver` initialization:
```py
        # reload proof from disk
        proof = APRProof.read_proof_data(proof_dir, proof_id)
        prover = APRProver(proof, kcfg_explore=kcfg_explore)
```
Is it correct behavior? I'm just skipping this `failure_info` test in `test_failure_info_recomputed_on_proof_reinit` for now, so it's only run in `test_failure_info`.

tothtamas28 (approved):

tothtamas28:
> The claim, which terminates early, has 1 terminal and 1 pending node when it's written to the disk, but after it's reloaded and is used to initialize `APRProver`, it has 2 terminal and 0 pending node (...)
> Is it correct behavior?

This sounds like a bug, can you open an issue?

anvacaru (approved):

palinatolmach:
> This sounds like a bug, can you open an issue?

Done: https://github.com/runtimeverification/pyk/issues/964

PR #963, opened 2024-03-11
KCFG viewer tweaks

goodlyrottenapple:
Hide custom view when none is supplied and show constraints in the tree view as booleans not ML. 

Note: These are being added to make the `kimp` demo tool more user friendly. See the original:

image

vs tweaks to pyk + better pretty printing in the LHS tree viewer:

image

@tothtamas28 the one thing I don't see a way to remove is the toggle custom key binding, as that seems to be a class constant. Would you know of a way to add it conditionally?


tothtamas28:
> @tothtamas28 the one thing I don't see a way to remove is the toggle custom key binding, as that seems to be a class constant. Would you know of a way to add it conditionally?

This can be made a regular object-level attribute / method if necessary.

tothtamas28 (commented):
Looks good to me. The only thing I'm a bit concerned about is how it affects downstream projects (`evm-semantics`, `kontrol`). Can you check preemptively?

goodlyrottenapple:
> Looks good to me. The only thing I'm a bit concerned about is how it affects downstream projects (`evm-semantics`, `kontrol`). Can you check preemptively?

I used this branch of pyk to test and display some proofs in kevm and didn't observe any issues...

tothtamas28 (approved):

tothtamas28 (approved):

src/pyk/kcfg/show.py
```diff
@@ -159,7 +159,7 @@ def _print_split_edge(split: KCFG.Split, target_id: int) -> list[str]:
             csubst = split.splits[target_id]
             ret_split_lines: list[str] = []
             substs = csubst.subst.items()
-            constraints = csubst.constraints
+            constraints = [ml_pred_to_bool(c, unsafe=True) for c in csubst.constraints]
```
ehildenb, line 162: This change will also upset a lot of output in downstream projects. We should make it optional at the very least.

PR #969, opened 2024-03-11
Refactor `Proof.status`

lucasmt:
- Replaces the original `Proof.status` abstract method with a new abstract method `own_status` that returns only the status of the current proof, not accounting for any subproofs.
- Adds a general implementation of `Proof.status` that considers both `own_status` and `subproofs_status`.
- Updates `APRProof` and `ImpliesProof` to implement `own_status` instead of `status`.


src/pyk/proof/proof.py
```diff
@@ -176,9 +176,21 @@ def subproofs_status(self) -> ProofStatus:
...
+        if self.admitted:
+            return ProofStatus.PASSED
+
+        if self.own_status == ProofStatus.FAILED or self.subproofs_status == ProofStatus.FAILED:
+            return ProofStatus.FAILED
+        elif self.own_status == ProofStatus.PENDING or self.subproofs_status == ProofStatus.PENDING:
+            return ProofStatus.PENDING
+        else:
+            return ProofStatus.PASSED
+
```
ehildenb, line 193:
```suggestion
    @property
    def status(self) -> ProofStatus:
        if self.admitted:
            return ProofStatus.PASSED
        if self.own_status == ProofStatus.FAILED or self.subproofs_status == ProofStatus.FAILED:
            return ProofStatus.FAILED
        elif self.own_status == ProofStatus.PENDING or self.subproofs_status == ProofStatus.PENDING:
            return ProofStatus.PENDING
        else:
            return ProofStatus.PASSED

```

Also probably don't need `elif` logic here, since they're all just returning. But otherwise looks good.

ehildenb (approved):

src/pyk/proof/proof.py
ehildenb: I'm not sure why the diff is all crazy, I just deleted a blank line.

PR #972, opened 2024-03-12
pretty print kore pattern in `SmtSolverError`

ovatman:
Closes runtimeverification/kontrol#396
Handels `SmtSolverError`s thrown by the `KoreClient` within `CTermSymbolic`, pretty prints the Kore definition in Kast.


anvacaru:
Looks good to me! But I would prefer @tothtamas28 to review it as well before merging.

src/pyk/kore/rpc.py
```diff
@@ -423,10 +424,16 @@ def __init__(self, error: str, context: Iterable[str]):
...
+    definition_dir: Path | None
 
-    def __init__(self, pattern: Pattern):
+    def __init__(self, pattern: Pattern, definition_dir: Path | None = None):
         self.pattern = pattern
-        super().__init__(f'Smt solver error: {self.pattern.text}')
+        self.definition_dir = definition_dir
+        if self.definition_dir is not None:
+            printer = KPrint(self.definition_dir)
+            super().__init__(f'Smt solver error: {printer.kore_to_pretty(self.pattern)}')
```
palinatolmach, line 434: I'd recommend making "SMT" uppercase since it's an abbreviation, here and below:
```suggestion
            super().__init__(f'SMT solver error: {printer.kore_to_pretty(self.pattern)}')
```

src/pyk/kore/rpc.py
```diff
@@ -866,12 +873,14 @@ class KoreClient(ContextManager['KoreClient']):
     _KORE_JSON_VERSION: Final = 1
 
     _client: JsonRpcClientFacade
+    definition_dir: Path | None
```
tothtamas28, line 876: This feature should leave the `KoreClient` and the `kore.prc` module unchanged. Instead, you can implement the error translation in `CTermSymbolic`.

tothtamas28 (changes requested):

src/pyk/cterm/symbolic.py
```diff
@@ -88,17 +90,22 @@ def execute(
...
+                module_name=module_name,
+                log_successful_rewrites=True,
+                log_failed_rewrites=self._trace_rewrites,
+                log_successful_simplifications=self._trace_rewrites,
+                log_failed_simplifications=self._trace_rewrites,
+            )
+        except SmtSolverError as err:
+            kast = self.kore_to_kast(err.pattern)
+            pretty_pattern = PrettyPrinter(self._definition).print(kast)
+            raise RuntimeError(f'\nSMT solver error:\n{pretty_pattern}') from err
```
tothtamas28, line 108: You can extract a method:

```python3
def _smt_solver_error(self, err: SmtSolverError) -> RuntimeError:
    kast = self.kore_to_kast(err.pattern)
    pretty_pattern = PrettyPrinter(self._definition).print(kast)
    return RuntimeError(f'\nSMT solver error:\n{pretty_pattern}')
```

Then in all four places:

```suggestion
        except SmtSolverError as err:
            raise self._smt_solver_error(err) from err
```

tothtamas28: Also, consider removing the leading newline:

```python3
RuntimeError(f'SMT solver error:\n{pretty_pattern}')
```

PR #976, opened 2024-03-13
Options classes

nwatson22:
Adds `Options` class hierarchy which stores all the parameters associated with command line options for pyk, and is extendable by downstream projects. This is a part of the changes pulled out from #967.
- Adds `Options` which automatically populates its fields from a prepared `dict` from the parsed command line args and sets default values when the values are missing
- Adds `LoggingOptions`, etc., which are groups of options that can be inherited by commands
- Adds `ProveOptions`, etc. which are directly associated with particular pyk commands but otherwise behave the same
- `exec_*` functions now take the corresponding `*Options` type instead of generic `Namespace`
- Defaults are no longer set at the argparse level since they are now set in `Options` constructor.


src/pyk/ktool/kprove.py
```diff
@@ -379,14 +378,14 @@ def _prove_claim_rpc(claim: KClaim) -> Proof:
             )
 
         all_claims = self.get_claims(
-            spec_file,
-            spec_module_name=spec_module_name,
+            Path(options.spec_file),
```
ehildenb, line 381: Isn't this already a `Path`?

ehildenb (approved):

src/pyk/ktool/kprove.py
nwatson22: I see it was assigned to `Path` in `ProveOptions` but the argparse type was still `str`. Fixed.

PR #980, opened 2024-03-13
Import the regression-new testsuite from K

ehildenb:
This PR imports the entire `regression-new` test-suite from K.

- The directory `k-distribution/tests/regression-new` from the K repository is copied to `regression-new`: https://github.com/runtimeverification/pyk/pull/958/commits/e33cc1fbc2f6b2fe571a9fb35bc2fd8067645a7a
- The tests that use regular `ktest.mak` are adjusted to point at `regression-new/include/ktest.mak` instead of the old path.
- The default backend to run these tests with is corrected to `llvm` in `regression-new/include/ktest.mak`.
- All of the newly added tests are added to a skipped list `regression-new/skipped`, so that they are not run on CI. The existing tests `regression-new/kprove-haskell` from https://github.com/runtimeverification/pyk/pull/958 and `regression-new/withConfig` from https://github.com/runtimeverification/pyk/pull/973 are left enabled.


regression-new/include/ktest.mak
```diff
@@ -22,7 +22,7 @@ KOMPILED_DIR=$(DEFDIR)/$(notdir $(DEF))-kompiled
 RUN_TESTS?=$(wildcard $(TESTDIR)/*.$(EXT))
 PROOF_TESTS?=$(wildcard $(TESTDIR)/*-spec.k) $(wildcard $(TESTDIR)/*-spec.md)
 # default KOMPILE_BACKEND
-KOMPILE_BACKEND?=haskell
+KOMPILE_BACKEND?=llvm
```
ehildenb, line 25: Here we correct the default backend being used.

regression-new/Makefile
```diff
@@ -1,2 +1,3 @@
-SUBDIRS=$(filter-out Makefile include, $(wildcard *))
+SKIPPED=$(shell cat skipped)
+SUBDIRS=$(filter-out Makefile include $(SKIPPED), $(wildcard *))
 include ./include/ktest-group.mak
```
ehildenb, line 3: Here we filter out most of the tests except those that are known passing.

PR #986, opened 2024-03-15
Address more issues in regression-new test-suite

ehildenb:
Part of: https://github.com/runtimeverification/pyk/issues/981

Here we address a bunch of the `kompile ...` options not being present in the `pyk kompile ...` CLI.

- `KompileOptions` is exposed directly to `KompileCommandOptions` class.
- The `Kompile` tool now expects structured input for options `llvm_kompile_type: LLVMKompileType` and `llvm_output: Path`, not strings.
- Uses of `--concrete-rules` flag are eliminated. In most cases, it's switched out with `[concrete]` attribute on rules, but in one case it means the test goes away because it tests a bad usage of the `--concrete-rules` flag directly.
- Options `--emit-json`, `--enable-search`, `--coverage`, `-ccopt`, `--llvm-kompile-type`, `--llvm-kompile-output`, `--bison-lists`, `--gen-bison-parser`, `--gen-glr-bison-parser`, and `--llvm-proof-hint-instrumentation` and threaded through from the CLI to their implementations.
- Logic is added to `kompile ...` to check if options which should only be passed for LLVM backend have been set.
- Increases the default timeout of `KoreClientTest` to avoid flakiness that happens on CI somewhat frequently (https://github.com/runtimeverification/pyk/actions/runs/8304847490/job/22731011234).
- A handful more tests are removed from `regression-new/skipped` and enabled.


src/pyk/ktool/kompile.py
```diff
@@ -254,27 +254,27 @@ class LLVMKompileType(Enum):
...
+    llvm_proof_hint_instrumentation: bool
 
     def __init__(
         self,
         base_args: KompileArgs,
         *,
-        llvm_kompile_type: str | LLVMKompileType | None = None,
-        llvm_kompile_output: str | None = None,
+        llvm_kompile_type: LLVMKompileType | None = None,
+        llvm_kompile_output: Path | None = None,
```
tothtamas28, line 270:
```suggestion
        llvm_kompile_type: str | LLVMKompileType | None = None,
        llvm_kompile_output: str | Path | None = None,
```

I like having these strings, it makes scripting / interpreter use very convenient because then calling `kompile` does not require the ceremony of importing everything.

src/pyk/cli/args.py
```diff
@@ -229,6 +258,48 @@ def kompile_args(self) -> ArgumentParser:
         args.add_argument('-O1', dest='o1', default=False, action='store_true', help='Optimization level 1.')
         args.add_argument('-O2', dest='o2', default=False, action='store_true', help='Optimization level 2.')
         args.add_argument('-O3', dest='o3', default=False, action='store_true', help='Optimization level 3.')
+        args.add_argument(
+            '--enable-search',
+            dest='enable_search',
+            default=False,
```
tothtamas28, line 264: IIRC these should be left unset in the new design:

```suggestion
```

tothtamas28 (approved):

Baltoli (approved):
Looks good to me from the perspective of the test outputs; will defer to Tamas for the Python review

src/pyk/cli/args.py
ehildenb: Hmmmm, I'm not sure. Most of the other arguments are coming with defaults currently (after reverting the original design), I tihnk we're not to where they don't come with defaults here yet.

I think the plan is for the defaults to be drawn from the `*Options` classes directly eventually, but not sure. @nwatson22 should I not be setting this flag right now? What about all the other `bool` arguments?

src/pyk/ktool/kompile.py
tothtamas28: I agree, for production code, we should definitely write `llvm_kompile_output=LLVMKompileType.C`, but it is still very convenient from time to time to just drop in an interpreter and write

```
> kompile(llvm_kompile_type='c')
```

and not go through all the ceremony.

For `llvm_kompile_output`, `Path` does not have any advantage in type checking against `str | Path`.

ehildenb: Done.

src/pyk/cli/args.py
ehildenb: Merging as-is, since more changes to Options parsing will catch this if it's a problem. I suspect it's not, because all the other options ar esetting defaults currently.

src/pyk/ktool/kompile.py
```diff
@@ -254,24 +254,26 @@ class LLVMKompileType(Enum):
...
     enable_llvm_debug: bool
+    llvm_proof_hint_instrumentation: bool
 
     def __init__(
         self,
         base_args: KompileArgs,
         *,
         llvm_kompile_type: str | LLVMKompileType | None = None,
-        llvm_kompile_output: str | None = None,
+        llvm_kompile_output: str | Path | None = None,
```
tothtamas28, line 270: The type checker does not flag this, but the argument should be converted to `Path | None` before setting the attribute.

ehildenb: Done.

src/pyk/cli/args.py
nwatson22: Right now I have it so the pyk command exclusive options, i.e. the ones corresponding to the classes in `pyk/cli/pyk.py` should be getting their default values entirely from `Options` but those in `pyk/cli/args.py` must specify the defaults in both places because KEVM, kontrol, and kmir depend on those arg groups.

PR #991, opened 2024-03-18
Add class `POSet`

tothtamas28:
Adds a class `POSet` that takes a binary relation, computes it's transitive closure and stores the image of the resulting relation. The sort ordering is re-implemented to use `POSet` for both KAST and KORE.

Note that technically, that the name `Proset` would be more fitting for the current implementation as neither antisymmetry, nor asymmetry is enforced. (Reflexivity isn't either, but that's easy to work around.)


src/pyk/kast/outer.py
```diff
@@ -1203,19 +1203,19 @@ def argument_sorts(self, label: KLabel) -> list[KSort]:
...
-        return list(set(_subsorts))
+    @cached_property
+    def subsort_table(self) -> FrozenDict[KSort, frozenset[KSort]]:
+        """Return a mapping from sorts to all their proper subsorts."""
+        poset = POSet(
+            (prod.sort, prod.items[0].sort)
+            for prod in self.productions
+            if len(prod.items) == 1 and isinstance(prod.items[0], KNonTerminal)
+        )
+        return poset.image
```
ehildenb, line 1214: I would recommend factoring out a method `KProduction.is_subsort` and using that here for filtering the productions, so that it's defined with a specific name to it. Otherwise looks good.

You could also consider add a method `KFlatModule.subsorts -> list[KProduction]` and `KDefinition.subsorts -> list[KProduction]`, which then use that filter.

That can also be pulled out into a separate issue where we move lots of the "sentence classifiers" onto the sentences themselves.

ehildenb (approved):

src/pyk/kast/outer.py
tothtamas28: I added a property `KProduction.as_subsort` that returns the pair `(supersort, subsort)` if the production is a subsort production, and `None` otherwise.

I didn't add `KDefinition.subsorts` as there's already a method with that name, and because the productions themselves do no really seem to be processed directly, only their transitive closure.

tothtamas28: Opened an issue: https://github.com/runtimeverification/pyk/issues/993

Issue #965, opened 2024-03-11
Store node attributes directly on the node


ehildenb:
Separated from https://github.com/runtimeverification/pyk/issues/919

Currently in the KCFG, we store the list of attributes as separate from the nodes themselves. Instead, we can make a `NodeAtt` class and store the list of attributes directly as a field on the node. This makes it easy to have a uniform attribute system that works even as we add more attributes, and to allow users to register hooks for checking different node attributes on loading new nodes.

For example, this issue (https://github.com/runtimeverification/pyk/issues/964), we have that on loading a proof we are checking each node for being terminal. But we could also just rely on the fact that every node, when entered into the graph, is checked for all the attributes the user may care about, and so we know that nodes which are `terminal` are already marked as such. No need to have explicit place in algorithm to mark them as such, and no need to check them on load time.

ehildenb:
Another issue that came up, @PetarMax is working on minimizing KCFGs for summarization, and found that node attributes are sticking around post-summarization when they probably shouldn't (in this one) https://github.com/runtimeverification/pyk/pull/995

Attaching attributes directly to nodes would solve this issue too.

Any updates @nwatson22 ?

nwatson22:
I've been thinking about this some, and I wasn't sure how we can eliminate the need to check on load time, since `KCFG.from_dict` loads components of the graph one at a time, but the attributes themselves may depend on other parts of the graph. Would it be a good idea to only check when the node is first added to the graph except `from_dict`, and then have all the attributes be saved/loaded from disk, the way `vacuous` and `stuck` already are?

PR #1002, opened 2024-03-19
Node attribute storage in nodes

nwatson22:
Fixes: https://github.com/runtimeverification/pyk/issues/965

Stores node attributes `terminal`,`stuck`, and `vacuous` in nodes rather than in KCFG or KCFGExploration.
- adds `attr` field to KCFG.Node
- adds enums `NodeAttr` (base), `KCFGNodeAttr`, and `KCFGExplorationNodeAttr`
- adds `add_attr`, `remove_attr`, and `discard_attr` to kcfg for updating the attrs of a given node, which replaces the node with a new one, since nodes are immutable.
- `replace_node` now takes a node. `let_node` is now for updating specific fields of nodes.
- Checks `terminal`, `stuck`, and `vacuous` by reference to the attributes on the node
- Removes separate `terminal`, `stuck` and `vacuous` lists.
- `from_dict` and `to_dict` on `Node` and `KCFG` updated to store relevant attributes in the node dicts. `KCFGStore` translates to and from the existing on-disk format.
- Tests in `test_proof.py` previously using `read_proof` and `write_proof` are now using `read_proof_data` and `write_proof_data`, since the older way of reading and writing does not use the `KCFGStore` and relies on `KCFG.to_dict` directly, without the translation.


src/pyk/kcfg/kcfg.py
```diff
@@ -84,19 +95,45 @@ def read_cfg_data(self) -> dict[str, Any]:
...
+
+        attrs = set(node.attrs)
+        if check_root(node_id):
+            attrs.add(KCFGNodeAttr.ROOT)
+        else:
+            attrs.discard(KCFGNodeAttr.ROOT)
+        if check_leaf(node_id):
+            attrs.add(KCFGNodeAttr.LEAF)
+        else:
+            attrs.discard(KCFGNodeAttr.LEAF)
```
ehildenb, line 115: `ROOT` and `LEAF` are not node attributes, same with `SPLIT`. These are properties of teh KCFG structure, not of the node itself.

src/pyk/kcfg/kcfg.py
```diff
@@ -87,11 +95,27 @@ def read_node_data(self, node_id: int) -> dict[str, Any]:
...
+    def compute_attrs(self, node_id: int) -> None:
+        node = self.node(node_id)
+
+        attrs = set(node.attrs)
+
+        node = KCFG.Node(node_id, node.cterm, attrs=attrs)
+
+        self._nodes[node_id] = node
+        self.update_refs(node_id)
+
```
ehildenb, line 107: What is this doing at this point? It seems to basically be a no-op now....

src/pyk/kcfg/kcfg.py
```diff
@@ -629,18 +651,14 @@ def remove_node(self, node_id: NodeIdLike) -> None:
...
-    def replace_node(self, node_id: NodeIdLike, cterm: CTerm) -> None:
-        node_id = self._resolve(node_id)
-        node = KCFG.Node(node_id, cterm)
-        self._nodes[node_id] = node
-        self._created_nodes.add(node.id)
+        for node in self.nodes:
+            self.compute_attrs(node.id)
 
+    def update_refs(self, node_id: int) -> None:
+        node = self.node(node_id)
```
ehildenb, line 661: Is this only intended to be used internally (hence why it takes an `int` and not a `NodeIdLike`)? If so, let's prefix it with `_update_refs`.

ehildenb: Though, I don't quite see how this isn't a no-op as well actually.... Since it seems that we're just replacing the node with itself again by looking up itself.

ehildenb: Oh I see now that `replace_refs` makes sure that each edge/successor is the newly looked up node.... Nice, I like it.

src/pyk/kcfg/kcfg.py
```diff
@@ -663,6 +681,32 @@ def replace_node(self, node_id: NodeIdLike, cterm: CTerm) -> None:
...
 
+    def remove_attr(self, node_id: NodeIdLike, attr: NodeAttr) -> None:
+        node = self.node(node_id)
+        new_attrs = set(node.attrs)
+        new_attrs.discard(attr)
+        new_node = KCFG.Node(node.id, node.cterm, new_attrs)
+        self._nodes[node.id] = new_node
+        self._created_nodes.add(node.id)
+        self.update_refs(node.id)
+        self.compute_attrs(node.id)
```
ehildenb, line 692: I guess we should implement `KCFG.Node.{remove,discard,add}_attr(attr: NodeAttr) -> KCFG.Node`, and use that functionality here. Basically return a copy of the previous node and then call `replace_node` routine.

src/pyk/kcfg/kcfg.py
```diff
@@ -789,30 +842,28 @@ def edge_likes(self, *, source_id: NodeIdLike | None = None, target_id: NodeIdLi
...
+        self.add_attr(node_id, KCFGNodeAttr.VACUOUS)
 
     def remove_vacuous(self, node_id: NodeIdLike) -> None:
-        node_id = self._resolve(node_id)
-        if node_id not in self._vacuous:
+        node = self.node(node_id)
+        if KCFGNodeAttr.VACUOUS not in node.attrs:
             raise ValueError(f'Node is not vacuous: {node_id}')
-        self._vacuous.remove(node_id)
+        self.remove_attr(node_id, KCFGNodeAttr.VACUOUS)
```
ehildenb, line 851: These should just call the `KCFG.Node.remove_attr(KCFGNodeAttr.VACUOUS)` (which will do the error handling there), and then `replace_ref`.

src/pyk/kcfg/kcfg.py
```diff
@@ -830,7 +881,7 @@ def create_split(self, source_id: NodeIdLike, splits: Iterable[tuple[NodeIdLike,
         source_id = self._resolve(source_id)
         split = KCFG.Split(self.node(source_id), tuple((self.node(nid), csubst) for nid, csubst in list(splits)))
         self.add_successor(split)
-        return split
+        return self._splits[source_id]
```
ehildenb, line 884: Why do we need this change?

ehildenb: I guess that `add_successor` was calling `compute_attrs` before, so we needed to do `_update_refs`. But if `compte_attrs` is a no-op now, we can just skip that there, and don't need this extra logic?

src/pyk/kcfg/kcfg.py
```diff
@@ -663,6 +681,32 @@ def replace_node(self, node_id: NodeIdLike, cterm: CTerm) -> None:
...
+        self.update_refs(node.id)
+        self.compute_attrs(node.id)
+
+    def update_node_cterm(self, node_id: NodeIdLike, cterm: CTerm) -> None:
+        node = self.node(node_id)
+        new_node = KCFG.Node(node.id, cterm, attrs=node.attrs)
+        self._nodes[node.id] = new_node
+        self._created_nodes.add(node.id)
+        self.update_refs(node.id)
+        self.compute_attrs(node.id)
```
ehildenb, line 708: Could we instead call this:

```
def let_node(self, node_id: NodeIdLike, cterm: CTerm | None = None, attrs: Iterable[KCFGNodeAttr] | None = None) -> None:
    ....
```

For more consistent naming with other parts of the codebase. Then you can use it as `KCFG.let_node(node_id, cterm=new_cterm)` if you need to update the cterm.

ehildenb: For example all the `KInner` classes implement `let(...)` methods.

ehildenb: If you were also to implement `KCFG.Node.let(cterm: CTerm | None = None, attrs: Iterable[KCFGNodeAttr] | None = None) -> KCFG.Node`, this could basically call that routine and then call `replace_node`.

src/pyk/kcfg/kcfg.py
ehildenb: I guess the only thing it's really doing is `update_refs`?

nwatson22: Removed.

src/pyk/kcfg/kcfg.py
nwatson22: I've renamed `update_refs` to `_update_refs`.

src/pyk/kcfg/kcfg.py
nwatson22: There are now `Node.{remove,discard,add}_attr` and these are used by corresponding `KCFG.{remove,discard_add}_attr`.

src/pyk/kcfg/kcfg.py
nwatson22: The error handling should be in one place now.

src/pyk/kcfg/kcfg.py
nwatson22: Yes, that's the reason. I've reverted this line.

src/pyk/kcfg/kcfg.py
nwatson22: Added `Node.let` and `KCFG.let_node`.

src/pyk/kcfg/kcfg.py
```diff
@@ -40,10 +41,17 @@
...
 _LOGGER: Final = logging.getLogger(__name__)
 
 
-NodeIdLike = int | str
+class NodeAttr(str, Enum): ...
+
+
+class KCFGNodeAttr(NodeAttr, Enum):
+    VACUOUS = 'vacuous'
+    STUCK = 'stuck'
```
ehildenb, line 54: What does inheriting from `str` give us?

Also, why do we need to doubly inherit from `NodeAttr, Enum`, isn't inheriting from `NodeAttr` giving us inheritance from `Enum` already?

lucasmt:
Do we also need to add the `refuted` attribute? I added it to the list displayed by `show.py` in PR #877, but it's possible that there were other places where I should have added it.

src/pyk/kcfg/kcfg.py
```diff
@@ -92,13 +120,35 @@ class KCFG(Container[Union['KCFG.Node', 'KCFG.Successor']]):
...
+        attrs: frozenset[NodeAttr]
+
+        def __init__(self, id: int, cterm: CTerm, attrs: Iterable[NodeAttr] = ()) -> None:
+            object.__setattr__(self, 'id', id)
+            object.__setattr__(self, 'cterm', cterm)
+            object.__setattr__(self, 'attrs', frozenset(attrs))
 
         def to_dict(self) -> dict[str, Any]:
-            return {'id': self.id, 'cterm': self.cterm.to_dict()}
+            return {'id': self.id, 'cterm': self.cterm.to_dict(), 'attrs': self.attrs}
```
ehildenb, line 131:
```suggestion
            return {'id': self.id, 'cterm': self.cterm.to_dict(), 'attrs': [a.value for a in self.attrs]}
```

I guess this is why we inherit `str`, but it's probably not necessary to do that, can just use `.value()` from `Enum`.

src/pyk/kcfg/kcfg.py
```diff
@@ -850,7 +903,7 @@ def create_ndbranch(
         source_id = self._resolve(source_id)
         ndbranch = KCFG.NDBranch(self.node(source_id), tuple(self.node(nid) for nid in list(ndbranches)), tuple(rules))
         self.add_successor(ndbranch)
-        return ndbranch
+        return self._ndbranches[source_id]
```
ehildenb, line 906: Do we still need this change? I believe not anymore...

src/pyk/proof/reachability.py
```diff
@@ -71,7 +71,6 @@ class APRProof(Proof, KCFGExploration):
 
     node_refutations: dict[int, RefutationProof]  # TODO _node_refutatations
     init: int
-    target: int
     bmc_depth: int | None
```
ehildenb, line 74: Why is `target` removed here? I believe that it's still being used...

nwatson22: Fixed.

src/pyk/kcfg/kcfg.py
nwatson22: Fixed.

src/pyk/kcfg/kcfg.py
nwatson22: I was trying to have it be a list of `str`, but I went with just having the attrs in the dict be a list of `NodeAttr`, so I guess it's no longer necessary to have it inherit `str`. But if we wanted it to be a list of `str` I don't know how we get from a string back to a `NodeAttr`, while allowing for arbitrary `NodeAttr` extensions.

src/pyk/kcfg/kcfg.py
nwatson22: I thought I read you needed to inherit directly from `Enum` to define enum values in that class, but upon testing it seems you don't. For the inheriting from `str`, I was basically trying to have us be able to read a string (the enum value) from disk and then be converted into the appropriate `NodeAttr`, but I think now I have it so that's not necessary.

nwatson22:
> Do we also need to add the `refuted` attribute? I added it to the list displayed by `show.py` in PR #877, but it's possible that there were other places where I should have added it.

@lucasmt I guess we aren't defining node attributes by the ones which are printed in the KCFGs. We're trying to just have internal node attributes here where the attribute could be determined just by looking at the node, i.e. not the rest of the CFG or Proof.

src/pyk/kcfg/exploration.py
```diff
@@ -70,13 +79,10 @@ def remove_terminal(self, node_id: int) -> None:
...
         node_id = self.kcfg._resolve(node_id)
         self.kcfg.remove_node(node_id)
-        self.remove_terminal(node_id)
 
     # Pruning a KCFG subtree starting from a given node
     def prune(self, node_id: NodeIdLike, keep_nodes: Iterable[NodeIdLike] = ()) -> list[int]:
         pruned_nodes = self.kcfg.prune(node_id, keep_nodes=keep_nodes)
-        for node_id in pruned_nodes:
-            self.remove_terminal(node_id)
         return pruned_nodes
```
ehildenb, line 86: I guess we don't need intermediate `pruned_nodes` anymore, can just `return self.kcfg.prune(...)`

src/pyk/kcfg/kcfg.py
```diff
@@ -603,18 +644,23 @@ def get_node(self, node_id: NodeIdLike) -> Node | None:
...
         self._nodes[node.id] = node
         self._created_nodes.add(node.id)
 
-    def create_node(self, cterm: CTerm) -> Node:
-        node = KCFG.Node(self._node_id, cterm)
-        self._node_id += 1
-        self._nodes[node.id] = node
-        self._created_nodes.add(node.id)
-        return node
+    def create_node(self, cterm: CTerm, node_id: int | None = None) -> Node:
```
ehildenb, line 653:
```suggestion
    def create_node(self, cterm: CTerm, node_id: int | None = None) -> KCFG.Node:
```

ehildenb: Why are we modifying this to allow specifying the `node_id` it gets? I can't see where that is used, and it seems that exposing that bit of internal state about the `KCFG` datastructure could only be a bad idea.

ehildenb (approved):

src/pyk/kcfg/exploration.py
nwatson22: Fixed.

src/pyk/kcfg/kcfg.py
nwatson22: Fixed.

src/pyk/kcfg/kcfg.py
```diff
@@ -590,31 +631,36 @@ def _resolve(self, id_like: NodeIdLike) -> int:
...
         self._nodes[node.id] = node
         self._created_nodes.add(node.id)
 
-    def create_node(self, cterm: CTerm) -> Node:
-        node = KCFG.Node(self._node_id, cterm)
-        self._node_id += 1
-        self._nodes[node.id] = node
-        self._created_nodes.add(node.id)
-        return node
+    def create_node(self, cterm: CTerm, node_id: int | None = None) -> KCFG.Node:
```
ehildenb, line 653: Still wondering why we add `node_id` optional attribute?

nwatson22: I was trying to minimize the number of spots in the code the nodes were being added, but it's not necessary now as it no longer performs attribute updates on node insertion.

PR #1008, opened 2024-03-20
add toml support for pyk options

ovatman:
this pr adds toml config support for pyk.

- First it introduces an Options class to handle config file (`--config-file`) and config profile (`--confişg-profile`) parameters. This argument set is added to all the pyk related commands as base arguments. 
- Later, toml file is processed to return a dictionary of option-value pairs. Afterwards, this dictionary is updated from the arguments coming from the command line.
- During toml handling, we need mapping between the option strings in the tomlş file and actual option destinations. To perform this mapping `from_option_string()` method is added whenever there's a mapping that involves a transformation other than dash-to-underscore transformation. These mappings are unified in `get_option_string_destination()` functions to form a lookup table related to the command being issued.
- Finally, for better testability some methods (most importantly `create_argument_parser()`) in the main module is moved to the `cli.pyk` module


Baltoli:
https://github.com/runtimeverification/k/pull/4254

PR #1012, opened 2024-03-21
Update kompile CLI and regression-new error messages

ehildenb:
This PR is part of https://github.com/runtimeverification/pyk/issues/981#issuecomment-2012635223.

Specifically:

- The `pyk kompile ...` command now accepts `--no-exc-wrap` option, and that option is added back to the `regression-new` test-suite for calls to `kompile`.
- The `run_process` in `pyk` now is a bit more careful about handling failed calls to subprocess that the user wants checked. Specifically, instead of causing a stack trace to be printed, just forwards the subprocesses errors, prints out the command that failed and teh exit code, and exits gracefully with the same exit code.
- A bunch of tests in `regression-new/` are updated with this new expected output.
- Many tests in `regression-new/` are now enabled as they have the correct output now.


src/pyk/utils.py
```diff
@@ -452,8 +452,10 @@ def run_process(
     delta_time = time.time() - start_time
     logger.info(f'Completed in {delta_time:.3f}s with status {res.returncode}: {command}')
 
-    if check:
-        res.check_returncode()
+    if check and res.returncode != 0:
+        sys.stderr.write(f'[ERROR] Running process failed with returncode {res.returncode}:\n    {shlex.join(args)}\n')
+        sys.stderr.flush()
+        sys.exit(res.returncode)
```
ehildenb, line 458: @tothtamas28 should I be writing to `stderr` directly or using `logger.error(...)`? I think the later may be more standard I guess?

ehildenb: Hmmmm, using the logger causes timestamps to be printed. Going to stick with `stderr` for now.

tothtamas28: I think the `pyk kompile` tool should just print to `stderr` / `stdout` as these messages are part of normal tool usage.

But `run_process(..., check=True)` should not result in these side effects, as `CalledProcessError` already has all the data:

```python3
proc_res = subprocess.run(['kompile', '--foo'], stderr=subprocess.PIPE)
try:
    proc_res.check_returncode()
except CalledProcessError as err:
    print(err.cmd)
    print(err.returncode)
    print(err.stderr)
```

So the `kompile` function can just let the exception propagate or do some error translation. The CLI can than catch the exception and perform the side effects.

ehildenb: Wouldn't we want a standard behaviour of `run_process(...)` across all our invocations of subtools though? I figured that would be a nice feature to have.

ehildenb: Especially since the callsite passed `check`, I think we should just go for standard behavior. If the callsite wants to do something specific with teh result, they can pass `check=False` and handle the returned `CompletedProcess` themselves directly, which is what we do in several places.

tothtamas28: I agree, this uniformity is desirable for the CLI. But this function has other uses as well (downstreams even ones we're not aware of), so in my opinion it'd be better to perform the side effects on the call site.

We can have a function that handles the error:

```python3
def exit_with_err(err: CalledProcessError) -> None:
    sys.stderr.write(f'[ERROR] Running process failed with returncode {err.returncode}:\n    {shlex.join(err.cmd)}\n')
    sys.stderr.flush()
    sys.exit(err.returncode)
```

Then:

```python3
try:
    kompile(...)
except CalledProcessError as err:
    exit_with_err(err)
```

src/pyk/utils.py
```diff
@@ -452,8 +452,10 @@ def run_process(
     delta_time = time.time() - start_time
     logger.info(f'Completed in {delta_time:.3f}s with status {res.returncode}: {command}')
 
-    if check:
-        res.check_returncode()
+    if check and res.returncode != 0:
```
tothtamas28, line 455: I do not agree with this change. The desired output can be achieved without a breaking change, consider implementing the uniform tool output on the command line layer.

tothtamas28 (approved):
LGTM except for the breaking change in `run_process`.

src/pyk/ktool/kompile.py
```diff
@@ -172,12 +172,7 @@ def __call__(
         try:
             proc_res = run_process(args, logger=_LOGGER, cwd=cwd, check=check)
         except CalledProcessError as err:
-            raise RuntimeError(
-                f'Command kompile exited with code {err.returncode} for: {self.base_args.main_file}',
-                err.stdout,
-                err.stderr,
-                err.returncode,
-            ) from err
+            exit_with_process_error(err)
```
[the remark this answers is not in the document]
ehildenb: I'd rather have `exit_with_process_error(...)` in `__main__` so that it's directly in the CLI application.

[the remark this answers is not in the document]
tothtamas28: > I'd rather have `exit_with_process_error(...)` in `__main__`

I agree that would be the ideal way to handle this.

At this point, it might be a good idea to extract `KompileError` that carries all the information. But just for making this PR pass with minimal breakage, you can include `err.cmd` in the `RuntimeError` along with the rest:

```python3
raise RuntimeError(
    f'Command kompile exited with code {err.returncode} for: {self.base_args.main_file}',
    err.stdout,
    err.stderr,
    err.returncode,
    err.cmd,
)
```

and access them in `__main__` from `err.args`.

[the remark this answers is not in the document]
tothtamas28: And in a separate PR, we can extract `KompileError(KToolError)` where

```python3
class KToolError(Exception, ABC):
    def exit(self) -> None:
        ...
```

so the CLI can just catch `KToolError` and call `exit` to print the error in the desired format.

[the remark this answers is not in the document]
ehildenb: Ok, I've threaded through the needed info to `__main__` for now, and made an issue here: https://github.com/runtimeverification/pyk/issues/1024

PR #1013, opened 2024-03-21
Computing substitutions when creating `Split` nodes

PetarMax:
This PR corrects the creation of `Split` nodes by computing the substitutions from the source to the target node, just as is done in the case of `Cover` nodes.


src/pyk/proof/reachability.py
```diff
@@ -415,7 +415,7 @@ def path_constraints(self, final_node_id: NodeIdLike) -> KInner:
             if type(edge) is KCFG.Split:
                 assert len(edge.targets) == 1
                 csubst = edge.splits[edge.targets[0].id]
-                curr_constraint = mlAnd([csubst.subst.ml_pred, csubst.constraint, curr_constraint])
+                curr_constraint = mlAnd([csubst.subst.minimize().ml_pred, csubst.constraint, curr_constraint])
```
PetarMax, line 418: I've changed this because I was getting prints in Kontrol with all of the identities, which, although important for the substitution, do not contribute to the path constraint.

src/tests/unit/test_kcfg.py
```diff
@@ -630,6 +630,17 @@ def test_minimize() -> None:
     assert cfg.contains_edge(KCFG.Edge(node_25, node(23, True), 155, ('r1', 'r2', 'r3', 'r4', 'r6', 'r7', 'r8')))
 
 
+def test_split_csubsts() -> None:
+    cfg = KCFG()
+    cfg.create_node(term(11))
+    cfg.split_on_constraints(1, [x_equals(10), x_equals(20)])
```
PetarMax, line 636: The test is for `split_on_constraints`, so it used it directly.

src/pyk/kast/inner.py
```diff
@@ -707,7 +707,7 @@ def to_dict(self) -> dict[str, Any]:
 
     def minimize(self) -> Subst:
         """Return a new substitution with any identity items removed."""
-        return Subst({k: v for k, v in self.items() if v != KVariable(k)})
+        return Subst({k: v for k, v in self.items() if not (type(v) is KVariable and v.name == k)})
```
PetarMax, line 710: This allows us to remove the identity substitution also if a sort is provided. I believe that this is fine because there shouldn't be a sort mismatch if a binding is already in the substitution, and the matcher for `KVariable`s doesn't look at the sort.

src/pyk/proof/reachability.py
```diff
@@ -581,9 +581,9 @@ def construct_node_refutation(self, node: KCFG.Node) -> RefutationProof | None:
         assert type(closest_branch) is KCFG.Split
         refuted_branch_root = closest_branch.targets[0]
         csubst = closest_branch.splits[refuted_branch_root.id]
-        if len(csubst.subst) > 0:
+        if len(csubst.subst.minimize()) > 0:
```
PetarMax, line 584: This needs also to be adjusted since substitutions in `Split`s are now non-empty, but are expected to be identities instead.

ehildenb: Maybe we could extract a function `Subst.is_identity`, and then can do `csubst.subst.is_identity`?

That way the logic for determining whether a substitution is an identity is opaque and happens in that class instead of here at the callsite.

src/pyk/kast/inner.py
ehildenb: The predicate `not (type(v) is KVariable and v.name == k)` is somewhat difficult to read, I guess what we'd want is `type(v) is KVariable implies v.name != k`, we could also say `type(v) is not KVariable or v.name != k`, which maybe reads slightly better.

ehildenb: But how it is is also fine too, just commenting on the readability.

src/pyk/proof/reachability.py
PetarMax: Yes, that's a good idea.

src/pyk/kast/inner.py
PetarMax: Yes, I'm ok with that change.

PR #1014, opened 2024-03-21
Add JSON-RPC server to pyk

nwatson22:
Addresses:
- https://github.com/runtimeverification/pyk/issues/1019

Adds:
- Class `JsonRpcServer`, an abstract class for exposing a JSON RPC interface over HTTP from pyk.
- The server is initialized with a host and port. Methods are registered with `JsonRpcServer.register_method`, and the server is started with `JsonRpcServer.serve`.
- A test that uses  JSON-RPC methods to interact with a k configuration and request concrete execution steps on it.


src/pyk/rpc/rpc.py
```diff
@@ -0,0 +1,201 @@
...
+        pattern = self.krun.kast_to_kore(self.config, sort=KSort('GeneratedTopCell'))
+        output_kore = self.krun.run_pattern(pattern)
+        self.config = self.krun.kore_to_kast(output_kore)
+        k_cell = get_cell(self.config, 'K_CELL')
+        if type(k_cell) is KSequence:
+            assert len(k_cell.items) == 1
+            k_cell = k_cell.items[0]
+
+        assert type(k_cell) is KToken
+        return int(k_cell.token)
```
ehildenb, line 201: These shouldn't be part of teh `main` functionality of pyk, just part of the test you make. They are specific to the semantics.

ehildenb:
We don't need to expose a default RPC server in `main`, just teh example one in the test. So the semantics specific stuff can move out of the `main` and into the test file directly.

src/pyk/rpc/rpc.py
```diff
@@ -0,0 +1,139 @@
...
+        self.send_json_success(result, request['id'])
+
+
+class ExampleJsonRpcServer(JsonRpcServer):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self.register_method('print_hello', self.exec_print_hello)
+
+    def exec_print_hello(self, obj: None) -> str:
+        return 'hello.'
```
ehildenb, line 139: What's this for? Can it be deleted? It seems the test we have serves as an example.

src/pyk/rpc/rpc.py
```diff
@@ -0,0 +1,139 @@
...
+from functools import partial
+from http.server import BaseHTTPRequestHandler, HTTPServer
+from typing import TYPE_CHECKING, Any, Final
+
+from typing_extensions import Protocol
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+    from ..cli.args import ServeRpcOptions
```
ehildenb, line 14: It doesn't seem right to import from the surface level `cli` module in an internal module here, because likely that module will also import this one eventually.

So maybe better to move `ServeRPCOptions` to this file, and then let the user import it as needed.

ehildenb (approved):
LGTM, maybe @tothtamas28 has some input too.

src/pyk/rpc/rpc.py
nwatson22: Yes, the test should be good enough. Removed.

src/pyk/rpc/rpc.py
nwatson22: Good point. `ServeRpcOptions` does still inherit from `Options`, though, but at least I can just import `pyk.cli.cli` which only has `Options`

ehildenb: Does it need to inherit from `Options` for its functionality though? Perhaps we could have the `ServeRPCOptions` be the class in `rpc.rpc`, and then have `ServeRPCCLIOptions` inherit from that, and add the `default()...` method? Ideally, I think we wouldn't have internal modules importing from CLI modules.

PR #1018, opened 2024-03-22
Allow loading single kcfg nodes

virgil-serbanuta:
This is useful when, e.g., binary searching for the first node where a certain value occurred.


tothtamas28 (approved):

src/pyk/kcfg/kcfg.py
```diff
@@ -81,6 +81,9 @@ def read_cfg_data(self) -> dict[str, Any]:
         dct['nodes'] = nodes
         return dct
 
+    def read_node_data(self, node_id: int) -> dict[str, Any]:
+        return json.loads(self.kcfg_node_path(node_id).read_text())
+
```
ehildenb, line 86: Can you make it so that `KCFG.read_cfg_data` reuses this function?

virgil-serbanuta: Done

ehildenb (approved):

PR #1022, opened 2024-03-25
Make KAST preprocessing pipeline explicit in `_module_to_kore`

tothtamas28:
* Define class `KompilerPass`
* Implement KAST preprocessing steps in terms of `KompilerPass`
* Refactor `simplified_module` to use a `Sequence[KompilerPass]`


ehildenb:
LGTM. I see that it seems the only non-`SingleModulePass` are `AddAnywhereAtt` and `AddPrioritiesAtt`? The `anywhere` attribute one could likely be solved by requiring the user to add it to the symbol that is an anywhere symbol, similar to how we do `macro` or `alias`, perhaps? But maybe that's not always possible with the uses of that attribute, which we'd have to audit.

ehildenb (approved):

tothtamas28:
> The `anywhere` attribute one could likely be solved

Opened an issue:
* runtimeverification/k#4132

Baltoli (approved):
Very nice, I like this design a lot. @gtrepta worth taking a look at this when we do the Scala -> Java port for the frontend's compiler pipeline.

PR #1025, opened 2024-03-25
SMT solver error timeout message added

ovatman:
This PR reflects changes in the runtimeverification/haskell-backend#3747 to pyk. When reporting `SMTSolverError` the cause is reported as an SMT timeout. In order to be able to report the current timeout parameter, we pass the exception within a new exception class `CTermSMTError`


palinatolmach (approved):
Approved! Please add a suggestion to increase the timeout using `--smt-timeout`.

palinatolmach (approved):
lgtm, @anvacaru could you please take a look too?

anvacaru (approved):
Looks good to me as well, perhaps I would change the `CTermSMTError` in `CTermSMTException`, but I do not have a strong opinion.

PR #1030, opened 2024-03-26
Fix `AddDomainValueAtts`

tothtamas28:
`AddDomainValueAtts` filtered out sentences other than sort declarations, which the test failed to catch due to a typo.


ehildenb (approved):
You could consider adding helpers to `KFlatModule` for projecting out certain sentence types directly, rather than filtering them at use-sites. Then tehy would be type-checked nicely too.

Or perhaps something like:

```
KFlatModule.get_sentence_type(T: type) -> list[T]:
```

Where you tell it the types of sentences you want projected?

tothtamas28:
> You could consider adding helpers to `KFlatModule` for projecting out certain sentence types directly

Done: #1036

PR #1009, opened 2024-03-20
Further KCFG-to-KCFG transformations

PetarMax:
This PR expands on #995  by adding a transformation that joins consecutive `Split` nodes into a single `Split`.

Additionally, the PR:
- corrects a bug in `lift_splits` that led to infinite looping in the case a split could not be lifted due to freshly created variables;
- corrects the substitution computation in `lift_split`;
- non-destructively expands the testing helpers in `test_kcfg.py` to allow for more expressivity;
- fixes a typo in the documentation of `CSubst.from_dict`.

PR #1040, opened 2024-03-28
Remove manip.get_cell, add CTerm.free_vars

ehildenb:
Inspired while reviewing: https://github.com/runtimeverification/pyk/pull/1009/

This makes two changes which hopefully tighten the `CTerm` abstraction slightly by reducing the places where `CTerm.kast` is used:

- Places that were doing `get_cell(CTerm.kast, CELL_NAME)` now say `CTerm.cell(CELL_NAME)`. There are no more uses of `get_cell` here or in downstream repos, so it's removed.
- A new `@cached_property CTerm.free_vars` is added, which is used to replace the pattern `free_vars(CTerm.kast)`.

This means that the remaining uses of `CTerm.kast` are doing something weird with the resulting kast, or printing it, or converting it to Kore.

### DIFF
--- a/src/pyk/cterm/cterm.py
+++ b/src/pyk/cterm/cterm.py
@@ -148,6 +148,11 @@ class CTerm:
         """Return the unstructured bare `KInner` representation of a `CTerm` (see `CTerm.from_kast`)."""
         return mlAnd(self, GENERATED_TOP_CELL)
 
+    @cached_property
+    def free_vars(self) -> set[str]:
+        """Return the set of free variable names contained in this `CTerm`"""
+        return set(free_vars(self.kast))
+
     @property
     def hash(self) -> str:
         """Unique hash representing the contents of this `CTerm`."""
@@ -235,7 +240,7 @@ class CTerm:
                 new_cterm = new_cterm.add_constraint(mlEqualsTrue(orBool([disjunct_lhs, disjunct_rhs])))
 
         new_constraints = []
-        fvs = free_vars(new_cterm.kast)
+        fvs = list(new_cterm.free_vars)
         len_fvs = 0
         while len_fvs < len(fvs):
             len_fvs = len(fvs)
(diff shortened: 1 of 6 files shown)


src/pyk/cterm/cterm.py
```diff
@@ -148,6 +148,11 @@ def kast(self) -> KInner:
         """Return the unstructured bare `KInner` representation of a `CTerm` (see `CTerm.from_kast`)."""
         return mlAnd(self, GENERATED_TOP_CELL)
 
+    @cached_property
+    def free_vars(self) -> set[str]:
+        """Return the set of free variable names contained in this `CTerm`"""
+        return set(free_vars(self.kast))
```
tothtamas28, line 154:
```suggestion
    def free_vars(self) -> frozenset[str]:
        """Return the set of free variable names contained in this `CTerm`"""
        return frozenset(free_vars(self.kast))
```

tothtamas28: Note however that making this a set will make iteration order nondeterministic, so sorting might be necessary on the call site.

src/pyk/cterm/symbolic.py
```diff
@@ -177,8 +177,8 @@ def implies(
     ) -> CTermImplies:
         _LOGGER.debug(f'Checking implication: {antecedent} #Implies {consequent}')
         _consequent = consequent.kast
-        fv_antecedent = free_vars(antecedent.kast)
-        unbound_consequent = [v for v in free_vars(_consequent) if v not in fv_antecedent]
+        fv_antecedent = list(antecedent.free_vars)
```
tothtamas28, line 180:
```suggestion
        fv_antecedent = antecedent.free_vars
```

tothtamas28 (approved):

src/pyk/cterm/cterm.py
ehildenb: Done. And yeah, I figure people shouldn't rely on the order the variables are reported back in anyway.

src/pyk/cterm/symbolic.py
ehildenb: Done.

PR #1044, opened 2024-03-28
Regression new updated expected output

ehildenb:
I'm opening this issue to discuss the various changes that running `make update-results` causes in `regression-new` test-suite. I figure we can keep this open as we triage all these things.


regression-new/checkClaimError/rule-spec.k.out
```diff
@@ -1,6 +1,33 @@
-[Error] Compiler: Only claims and simplification rules are allowed in proof modules.
+[Error] Compiler: Only claims and simplification rules are allowed in proof
+modules.
 	Source(rule-spec.k)
 	Location(6,10,6,43)
 	6 |	    rule <k> doIt(foo) => doIt(0) ... </k>
 	  .	         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 [Error] Compiler: Had 1 structural errors.
+Traceback (most recent call last):
```
ehildenb, line 8: I guess for these errors we can:

- Add option `--no-wrap-exc` to `pyk prove ...`, and pass it through to `kprove --dry-run ...`
- Implement custom error handling for `pyk prove ...` similar to what we do for `kompile`

The only challenge is that here we are using a temporary file as the JSON dump, and that will show up in the error message. So I guess we also need to:

- Allow specifying `--tmp-dir ...` for `pyk prove ...`, and use that directory for dumping the temporary files.
- Use a more deterministic filename for the temporary JSON file we dump.

OR

- Dump the JSON in a file immediately next to the spec file as a convention.

OR

- Strip that line of output in the testing harness.

regression-new/concrete-function-cache/a2-spec.k.out
```diff
@@ -1,24 +1,27 @@
...
+  Node id: 5
+  Failure reason:
+    Matching failed.
+    The following cells failed matching individually (antecedent #Implies consequent):
+    K_CELL: end ( bar ( X:Int ) ) #Implies end ( baz ( X:Int ) )
+  Path condition:
+    #Ceil ( bar ( X:Int ) )
+  Failed to generate a model.
+
+Join the Runtime Verification Discord server for support: https://discord.gg/CurfmXNtbN
```
ehildenb, line 27: This failure information message contains largely the same information as the original message, basically the part that failed and path condition to it. Are we happy with this new way of presenting information?

regression-new/io-llvm/4.test.out
```diff
@@ -1,3 +1,10 @@
...
+<generatedTop>
+  <k>
+    "open"
+    ~> "(\"foo\", \"r\")\n"
+    ~> ""
+  </k>
+  <generatedCounter>
+    0
+  </generatedCounter>
+</generatedTop>
```
ehildenb, line 10: pyk prints out `KSequence` on newlines, as an aesthetic choice. Do we want to allow this and make it default, or retain the original version? I guess printing `KSequence` on newlines would look bad if there is a `KSequence` inside of some other construct in the configuration.

regression-new/issue-1098/1.test.out
```diff
@@ -1,3 +1,8 @@
...
-  x = ( 0 ~> .K ) ; .RecordDescr ~> .K
-</k>
+<generatedTop>
+  <k>
+    ... key: x = value: 0 ~> .K ; .RecordDescr ~> .K
+  </k>
+  <generatedCounter>
+    0
+  </generatedCounter>
+</generatedTop>
```
ehildenb, line 8: This seems to be a bug in how pyk unparser works.

regression-new/issue-1528/1.test.out
```diff
@@ -1,3 +1,9 @@
...
-</k>
+<generatedTop>
+  <k>
+    bar ( 0 , 0 )
+    ~> bar ( 0 , 0 )
+  </k>
+  <generatedCounter>
+    1
+  </generatedCounter>
+</generatedTop>
```
ehildenb, line 9: We should be including the tail of the sequence every time in pyks unparser.

PR #1046, opened 2024-03-28
Fix `as_subsort`

tothtamas28:
If `symbol` is present on a production, then it is not a subsort production, even if it has a single, non-terminal production item.


ehildenb:
Can we test this? I guess it must have shown up in some definition somewhere as a bug.

ehildenb (approved):

PR #1052, opened 2024-03-30
Location information in outer lexer

gtrepta:
runtimeverification/k#4180


src/pyk/kast/outer_lexer.py
```diff
@@ -169,8 +174,45 @@ class State(Enum):
 _BUBBLY_STATES: Final = {State.BUBBLE, State.CONTEXT}
 
 
+class LocationIterator(Iterator[str]):
```
tothtamas28, line 177: I think this can be simplified to:

```python3
class LocationIterator(Iterator[str]):
    _line: int
    _col: int

    ...

    def loc(self) -> tuple[int, int]:
        return self._line, self._col
```

and let the caller handle all other stateful parts.

src/pyk/kast/outer_lexer.py
```diff
@@ -60,6 +61,10 @@ class TokenType(Enum):
 class Token(NamedTuple):
     text: str
     type: TokenType
+    start_line: int = 0
+    start_column: int = 0
+    end_line: int = 0
+    end_column: int = 0
```
tothtamas28, line 67: Here I'd just track the start location of the `Token`:
```suggestion
    loc: tuple[int, int]
```

The reason for this:
- The lexer is already pretty complicated.
- This is straightforward to implement (and test), it just requires saving `it.loc` on the first character of the token.
- For most tokens of a given parsed text, neither start, nor end location is interesting, so calculating and storing only one of them saves resources.
- For all tokens other than `BUBBLE`, the end location is `(token.line, token.col + len(token.text) - 1)`, so calculating it in the parser is straightforward.
- `BUBBLE` will be inner-parsed anyways, so the end location there is probably not indispensable.

gtrepta:
Marking this for review without doing tokens for attributes. Attributes don't have any location information in the kast format, so supporting it would require more work than just getting locations on these tokens.

src/pyk/kast/outer_lexer.py
```diff
@@ -57,9 +59,50 @@ class TokenType(Enum):
...
 
+class Location:
+    _line: int = 0
+    _col: int = 0
+
+    def __init__(self, line: int = 1, col: int = 0) -> None:
+        self._line = line
+        self._col = col
+
+    def __add__(self, s: str) -> Location:
```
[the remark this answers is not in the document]
gtrepta: Adding two `Location`s unfortunately doesn't really make much sense to me.

What I need is some kind of convenience function for incrementing a `Location` based on some sequence of characters that have been read in. The logic here for iterating through the characters and checking for newlines needs to exist somewhere. Putting it in the addition operator makes it feel easier to invoke in the places I need it. But I could put it in a member method named something like `add_str` if you prefer.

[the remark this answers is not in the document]
tothtamas28: I think it would make sense in at least a few places  For example, I think

```
loc + bubble[: match.start('lbrack')]
```

could be replaced by adding the location tuple extracted from the `Match` object. This way, you avoid processing the chunk of text twice.

If this works, you can consider all the other uses of `+` with `str`, maybe those can be eliminate as well.

src/pyk/kast/outer_lexer.py
```diff
@@ -617,22 +705,25 @@ def _raw_bubble(la: str, it: Iterator[str], keywords: Collection[str]) -> tuple[
...
-        return [], bubble
+        return [], bubble, loc
 
     return (
         [
-            _LBRACK_TOKEN,
-            Token(match['label'], TokenType.RULE_LABEL),
-            _RBRACK_TOKEN,
-            _COLON_TOKEN,
+            _LBRACK_TOKEN.with_loc(loc + bubble[: match.start('lbrack')]),
```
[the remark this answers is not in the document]
gtrepta: There might be a way to reduce the amount of re-traversing, but I'm not sure there's a way to avoid it entirely as the text can contain any number of newlines between all of the tokens, and they need to be accounted for.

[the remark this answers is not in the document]
gtrepta: I've made a commit with the best I think I can do. I'm not sure what location info in the `Match` you're referring to, afaik you can only get string offsets from there and they don't do any handling of line/column information.

src/tests/unit/test-data/outer-lexer/rule-label-attr.test
```diff
@@ -53,25 +61,33 @@ BUBBLE
     </k>
 
 LBRACK
+1,0
```
tothtamas28, line 64: This doesn't look right.

src/tests/unit/kast/test_outer_lexer.py
```diff
@@ -5,6 +5,9 @@
 import pytest
 
 from pyk.kast.outer_lexer import (
+    _INIT_LOC,
```
tothtamas28: Consider dropping the underscore.

src/tests/unit/kast/test_outer_lexer.py
```diff
@@ -578,6 +662,9 @@ def _tokens(lines: Iterator[str]) -> list[Token]:
             token_lines: list[str] = []
             token_type = TokenType[line]
 
+            line = next(lines)
+            loc = Loc(*map(int, line.split(',')))
```
tothtamas28:
```suggestion
            row, col = line.split(',')
            loc = Loc(row, col)
```

src/pyk/kast/outer_lexer.py
```diff
@@ -57,56 +58,74 @@ class TokenType(Enum):
     BUBBLE = auto()
 
 
+class Loc(NamedTuple):
+    line: int
```
tothtamas28, line 62:
```suggestion
    row: int
```

(This is a nitpick, so feel free to disregard the suggestion.)

tothtamas28 (commented):
I appreciate the work you put in to update the tests, thank you!

src/tests/unit/test-data/outer-lexer/rule-label-attr.test
gtrepta: This is one of the attribute tokens, which aren't getting location info because attributes don't have source locations in KAST, so it would be extra work in this PR for nothing. Instead, they're getting the default `INIT_LOC`.

tothtamas28: It is not for nothing, it enables more fine-grained error reporting for attributes. Please open an issue so that we can address this later.

tothtamas28: In that issue, please also include that we should add the following test.

After lexing, for each token, if we consider the substring of the input text starting at the token location, it will start with the token text.

tothtamas28 (approved):
Looks great!

tothtamas28:
Converting this to draft to avoid accidentally merging it before the repo merge.

src/tests/unit/test-data/outer-lexer/rule-label-attr.test
gtrepta: ~~#1078~~ https://github.com/runtimeverification/k/issues/4187

Baltoli:
https://github.com/runtimeverification/k/pull/4251

PR #1056, opened 2024-04-02
Exporting `get_requires` utils function

Robertorosmaninho:
This PR introduces a new file, `utils.py`, which should contain auxiliary functions that need to be converted from and to llvm but don't set any field on an object. 

In this context we're using a function from llvm backend's [pattern_matching.cpp](https://github.com/runtimeverification/llvm-backend/blob/e2f583a05b9bf362deabd58f8acc1ce534aaf4c4/lib/ast/pattern_matching.cpp#L174) to avoid reimplement it in Python. This function is exported as Pyhton binding in this [PR](https://github.com/runtimeverification/llvm-backend/pull/1019), and therefore, the current PR can only be merged after we merge the LLVM Backend one.


tothtamas28 (commented):
Please add a unit test for this feature. That will also block merging this PR until the binding becomes available.

tothtamas28:
> In this context we're using a function from llvm backend's [pattern_matching.cpp](https://github.com/runtimeverification/llvm-backend/blob/e2f583a05b9bf362deabd58f8acc1ce534aaf4c4/lib/ast/pattern_matching.cpp#L174) to avoid reimplement it in Python

Hasn't there been a plan to eventually implement it in `pyk` though?

Robertorosmaninho:
> > In this context we're using a function from llvm backend's [pattern_matching.cpp](https://github.com/runtimeverification/llvm-backend/blob/e2f583a05b9bf362deabd58f8acc1ce534aaf4c4/lib/ast/pattern_matching.cpp#L174) to avoid reimplement it in Python
> 
> 
> 
> Hasn't there been a plan to eventually implement it in `pyk` though?

That was revisited, and we decided to move in another direction to avoid having the exact implementation in two different places when we could just export them.

Robertorosmaninho:
> Please add a unit test for this feature. That will also block merging this PR until the binding becomes available.

I've just added a unit test, and I'm waiting for this [PR](https://github.com/runtimeverification/k/pull/4149) to be approved

src/pyk/kcfg/kcfg.py
```diff
@@ -1080,8 +1080,8 @@ def lift_splits(self) -> bool:
         """
 
         def _lift_split(finder: Callable, lifter: Callable) -> bool:
+            result = False
```
tothtamas28, line 1083: There are some changes in the diff from meged commits, please rebase on `master`.

Robertorosmaninho: Solved! Sorry!

tothtamas28 (approved):

PR #1063, opened 2024-04-04
Use total function for variable lookup in `IMP`

tothtamas28:
Related: https://github.com/runtimeverification/hs-backend-booster/issues/569#issuecomment-2036847940

Updates `IMP` to handle `Map` lookups the same way it is done in the Booster test suite, which  significantly improves execution time for the Booster.

| Backend |   Old   |   New   |
|---------|---------|---------|
| Legacy  |  10.92s |  11.31s |
| Booster | 130.47s |  13.48s |


jberthold (approved):
LGTM

PR #1065, opened 2024-04-05
Ordered constraint accumulation

PetarMax:
Currently, `CSubst` constraints are always sorted when a new `CSubst` is created. However, when a KCFG is minimized and splits are lifted into other splits, this sorting results in a misordering of constraints and therefore a slightly poorer user experience when trying to debug the control flow. For example, even though the branchings constraints were chronologically `C1, C2, C3`, when the KCFG is minimized they could appear as `C2, C3, C1`.

This PR enables constraint order preservation and uses it when lifting split nodes.


src/tests/unit/test_kcfg.py
```diff
@@ -752,6 +752,102 @@ def test_minimize_02() -> None:
     assert contains_edge(cfg, 8, 6, 40, ('r1', 'r3'))
 
 
+def test_split_constraint_accumulation() -> None:
```
tothtamas28, line 755: Nice!

src/pyk/cterm/cterm.py
```diff
@@ -93,11 +93,12 @@ def _check_config(config: KInner) -> None:
...
 
     @staticmethod
-    def _normalize_constraints(constraints: Iterable[KInner]) -> tuple[KInner, ...]:
+    def _normalize_constraints(constraints: Iterable[KInner], maintain_order: bool = False) -> tuple[KInner, ...]:
         constraints = (constraint for _constraint in constraints for constraint in flatten_label('#And', _constraint))
         constraints = unique(constraints)
         constraints = (constraint for constraint in constraints if not CTerm._is_spurious_constraint(constraint))
-        constraints = sorted(constraints, key=CTerm._constraint_sort_key)
+        if not maintain_order:
+            constraints = sorted(constraints, key=CTerm._constraint_sort_key)
```
tothtamas28, line 101: IIRC, sorting of constraints was important for the old KSummarizer where we compared nodes for equality a lot. I think it should be fine to just drop sorting without introducing an additional parameter, after testing it on `kontrol`.

@ehildenb, what do you think?

ehildenb: Well, I think there are still some places where we could be comparing equality of `CTerm` without considering different constraint ordering. So I wolud drop the parameter, not sort them by default, and then test with Kontrol. If there are no problems, good to merge. Otherwise, maybe we need to implement explicit equality operator for `CTerm` directly.

So in short, I agree with @tothtamas28 to not provide the option to have the constraints out of order.

PetarMax: Great, thank you, I will co-ordinate with @Baltoli and make the changes once the PR has been moved to the new setting of `pyk`.

Baltoli:
https://github.com/runtimeverification/k/pull/4249

PR #1069, opened 2024-04-05
Add separate get proof steps and execute proof step phase

nwatson22:
Separates `get_steps` from `step_proof`. Simple implementation of proof steps for now, i.e. just a reference to the proof plus the node ID to step on, or in the case of `ImpliesProof`, just the proof, since there is always exactly one step for those.


src/pyk/proof/implies.py
```diff
@@ -55,6 +60,11 @@ def __init__(
         self.simplified_consequent = simplified_consequent
         self.csubst = csubst
 
+    def get_steps(self) -> Iterable[ProofStep]:
```
tothtamas28:
```suggestion
    def get_steps(self) -> list[ProofStep]:
```

src/pyk/proof/implies.py
```diff
@@ -404,29 +414,31 @@ def __init__(self, proof: ImpliesProof, kcfg_explore: KCFGExplore):
         super().__init__(kcfg_explore)
         self.proof = proof
 
-    def step_proof(self) -> Iterable[StepResult]:
-        proof_type = type(self.proof).__name__
-        _LOGGER.info(f'Attempting {proof_type} {self.proof.id}')
+    def step_proof(self, step: ProofStep) -> Iterable[StepResult]:
+        assert isinstance(step, ImpliesProofStep)
```
tothtamas28: This does not have to be addressed in this PR. But in order to have stronger typing guarantees, consider using generics:

```python3
class Prover[Generic[PS, SR]]:

    @abstractmethod
    def step_proof(self, step: PS) -> Iterable[SR]: ...
```

Classes

```python3
class ProofStep: ...

class StepResult: ...

...
```

would then be the upper bounds on the respective type variables (or as long as they are empty, could even be omitted).

src/pyk/proof/reachability.py
```diff
@@ -132,6 +138,9 @@ def __init__(
                 assert type(subproof) is RefutationProof
                 self.node_refutations[node_id] = subproof
 
+    def get_steps(self) -> Iterable[APRProofStep]:
```
tothtamas28:
```suggestion
    def get_steps(self) -> list[APRProofStep]:
```

src/pyk/proof/reachability.py
```diff
@@ -715,37 +724,37 @@ def _check_subsume(self, node: KCFG.Node) -> CSubst | None:
             _LOGGER.info(f'Subsumed into target node {self.proof.id}: {shorten_hashes((node.id, self.proof.target))}')
         return csubst
 
-    def step_proof(self) -> Iterable[StepResult]:
-        if not self.proof.pending:
-            return []
-        curr_node = self.proof.pending[0]
+    def step_proof(self, step: ProofStep) -> Iterable[StepResult]:
```
tothtamas28:
```suggestion
    def step_proof(self, step: ProofStep) -> list[StepResult]:
```

tothtamas28 (approved):

tothtamas28:
Converted this to draft to avoid accidentally merging it before the pyk -> K migration is done.

src/pyk/proof/implies.py
nwatson22: Fixed.

src/pyk/proof/implies.py
nwatson22: It's a small change, so I've gone ahead and included the generics in the current PR.

src/pyk/proof/reachability.py
nwatson22: Fixed.

src/pyk/proof/reachability.py
nwatson22: Fixed.

Baltoli:
https://github.com/runtimeverification/k/pull/4246

PR #1076, opened 2024-04-11
add spec module in definition_args

### LINKED ISSUE #2367: Unable to run claims with `kevm prove`
I have a `lemmas.k` file with a `LEMMAS` module and a `LEMMAS-SPEC` module with a claim `claim-1`. I'm trying to run this claim with the commands
```
kevm kompile-spec lemmas.k --target haskell --main-module LEMMAS --syntax-module LEMMAS -o kompiled
kevm prove lemmas.k --claim LEMMAS-SPEC.claim-1 --definition kompiled
```
but I get the error `ValueError: Claim labels not found: ['LEMMAS.LEMMAS-SPEC.claim-1']`. In other words, it automatically prefixes the claim label with the main module `LEMMAS`, which is incorrect because the claim is in the `LEMMAS-SPEC` module. It's not even possible for the claim to be in the main module because in this case `kevm kompile-spec` fails with the error message `Claims are not allowed in the definition.`

`kevm prove` used to have a `--spec-module` option, but it no longer seems to accept it, so I don't know how else to specify the spec module.

anvacaru:
Closes: https://github.com/runtimeverification/evm-semantics/issues/2367 
- adding `spec_module` to `definition_args`.


Baltoli:
Superseded by https://github.com/runtimeverification/k/pull/4172

